PR #41, opened 2023-08-16
fix: add ports for docker compose

adam2k:
This is WIP, but adding ports here at least allows for ingress for RPC commands.


willcl-ark:
Thanks Adam. Closing as stale.

PR #44, opened 2023-08-17
remove unused, modularise

willcl-ark:
re-run `pip install -e .` after pull

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,5 +20,5 @@ dependencies = [
 dynamic = ["version"]
 
 [project.scripts]
-warnet = "warnet.connect_nodes:main"
+warnet = "warnet.main:main"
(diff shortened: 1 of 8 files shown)


josibake:
:shipit:

pinheadmz:
`graph` is undefined:


```python


def generate_compose(graph_file: str):
    """
    Read a graph file and build a docker compose.

    :param graph_file: The path to the graph file
    """
    try:
        logging.info(
            f"  Graph file contains {len(graph.nodes())} nodes and {len(graph.edges())} connections"
        )
        logging.info("  Generated docker-compose.yml file")
    except Exception as e:
        logging.error(f"An error occurred while running new node: {e}")
```

pinheadmz:
actually i think that whole function is dead, so

PR #47, opened 2023-08-18
don't bake in conf files

### LINKED ISSUE #8: Don't bake bitcoin.conf into the docker image
This means we have to rebuild the images every time we want a conf change

willcl-ark:
closes # 8

merge this or 46 first?


pinheadmz:
might be better to have a template with "common" conf in one file and then generate the other needed files on the fly, like in a tmp dir or something. Ideally i think wed be able to add a list of options in the graph file directly. wha chu think bout that?

willcl-ark:
OK will do that

PR #49, opened 2023-08-18
Add traffic shaping per-container

### LINKED ISSUE #17: simulate network latency
This can happen on the node itself or by doing something with docker containers

willcl-ark:
Closes #17

### DIFF
--- a/Dockerfile_build
+++ b/Dockerfile_build
@@ -5,6 +5,7 @@ RUN apt-get update && apt-get install -y \
     ccache \
     python3 \
     vim \
+    iproute2 \
     build-essential \
     wget \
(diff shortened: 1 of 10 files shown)


willcl-ark:
I think this is roughly ready now

Traffic shaping can be provided on a per-node basis using the graph property `tc_netem`. The commands are sanitized using a crazy regex from ChatGPT, which seems to work on current examples.

If no graph specified conditions are found it currently will use a random condition profile. I think this is quite nice, but we might want to have this require a startup option flag (or other), and not be the default? e.g.

`warnet --randomize-network-conditions` to turn that function on?

PR #50, opened 2023-08-18
Tor: install and run tor daemon

pinheadmz:
bitcoin listens and connects to Tor, even connects to other warnet nodes over Tor:

Screen Shot 2023-08-18 at 3 36 18 PM


willcl-ark:
Does it not start automagically?

pinheadmz:
> Does it not start automagically?

It should but you may need to wipe docker cache

willcl-ark:
That seems acceptable?

pinheadmz:
This PR got a bit out of scope with a few extra bug fixes but I now have the sensitive relay PR running perfectly in Warnet, with real Tor connections. 

I think we should consider 11 nodes as the minimum network size, kinda how like you need at least 100 blocks in a chain before anything really works. I removed a bunch of unused stuff from the graphml file. Also, that default graphml file is kinda like our single integration test: it just does all the features!

<a href="https://asciinema.org/a/fXo5uDR8wJ3SvWE7XeBvdtjYb" target="_blank"></a>

willcl-ark:
Can we proxy these Tor connections and spoof tor latency using #49, so that it doesn't screw up the real Tor network?

pinheadmz:
Update: this branch now deploys a local Tor network! Inspired by https://github.com/antitree/private-tor-network

Here's how it works:
- Every tank has Tor installed
    - Each tank is configured as a Tor relay
    - Bitcoin Core in each tank connects to Tor and creates a hidden service
- `tank_000000` specifically is ALSO configured as a Tor Directory Authority
    - This tank has a hard coded IP (`100.20.15.18`)
    - This tank has extra pre-generated keys dumped into `/root/.tor/keys`
    - These (public) keys and the IP address is set in the `torrc` files of every tank as the `DirAuthority`

To see it all in action, checkout this branch then:
Start:  `warnet start from-file src/templates/example.graphml`
Watch in one terminal: `watch docker exec -it tank_000010 bitcoin-cli -netinfo 4`
Run: `warnet run sens_relay`

You should see the `srelay` onion connection pop up for a moment when the test node sends its transaction. 

TODO:
- We could move the Tor directory service to its own container
- Hard coding the directory IP breaks our flexibility about subnets and should probably be generated on the fly

willcl-ark:
So as I was (trying) to do with bind, it should be possible to just update torrc after the DA is assigned to one node and restart tor on each tank?

Edit: in fact, perhaps we should have a specific step between generating the docker compose (and all starting tank ip addresses are known), where various roles can be assigned, and config file (templates) can be updated. This would currently be the zone file and torrc, but could be others in the future...

pinheadmz:
We may not even need to restart anything:

```python
    wn = Warnet.from_graph_file(graph_file, network)  # <-- All IP addresses are generated in this step
    wn.write_bitcoin_confs()
    wn.write_docker_compose()

    # write zone file and torrc files

    wn.write_prometheus_config()
    wn.docker_compose_up()
    wn.apply_network_conditions()
    wn.connect_edges()

```

willcl-ark:
> We may not even need to restart anything:
> 
> 
> 
> ```python
> 
>     wn = Warnet.from_graph_file(graph_file, network)  # <-- All IP addresses are generated in this step
> 
>     wn.write_bitcoin_confs()
> 
>     wn.write_docker_compose()
> 
> 
> 
>     # write zone file and torrc files
> 
> 
> 
>     wn.write_prometheus_config()
> 
>     wn.docker_compose_up()
> 
>     wn.apply_network_conditions()
> 
>     wn.connect_edges()
> 
> 
> 
> ```

Yes this is what I was envisioning

pinheadmz:
or actually, write the files before docker-compose so the volumes can include the files from the tmp dir

PR #51, opened 2023-08-21
Use random ip addresses within given subnet

willcl-ark:


Extheoisah (approved):

pinheadmz:
tested ACK, good work. will help ensure peers are in different buckets as well

pinheadmz:
merging now to rebase tor branch... i think since this will affect bucketing it will also reduce the addman spamming in the logs, because we will be more liekly to connect to hosts we pull form the addman list.

PR #55, opened 2023-08-21
fixes #54

### LINKED ISSUE #54: bitcoin.conf ain't quite right...
For starters, we are using the same conf file in each tank:

https://github.com/bitcoin-dev-project/warnet/blob/030b3b128a43323307c0f8fde39df48c46824ffc/src/warnet/docker_compose.py#L129-L130

When I tried to fix this myself it kinda worked but `regtest=1` was missing from all the configure files and everything tried to sync main!

willcl-ark:
* use node_id from graph instead of iterator
* use distinct conf

@pinheadmz want to try this solution?

### DIFF
--- a/config/bitcoin.conf
+++ b/config/bitcoin.conf
@@ -1,8 +1,8 @@
-# [network]
+# Warnet environment
 dnsseed=0
-
-# [debug]
 regtest=1
+
+# Debugging
 debug=1
 logips=1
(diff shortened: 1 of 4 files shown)


pinheadmz:
> @pinheadmz want to try this solution?

not bad, I added another commit that fixed a few other bugs like having multiple values for same key and rebased on main

PR #56, opened 2023-08-21
feat: add warnet api service

Extheoisah:
Todo(in a follow up pr):
    - add auth
    - add middleware
    - define network properties

### DIFF
--- a/api-compose.yml
+++ b/api-compose.yml
@@ -10,8 +10,6 @@ services:
       POSTGRES_USER: warnet
     networks:
       - warnet-api
-    volumes:
-      - ./db:/var/lib/postgresql/data
 
   api:
     build:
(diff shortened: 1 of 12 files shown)


api-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3"
+services:
+  db:
+    image: postgres:14.0
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_PASSWORD: password
```
okjodom, line 8: there's a number of config and app secrets committed in code, and we should consider secure provisioning

api/main.py
```diff
@@ -0,0 +1,27 @@
...
+    version="0.0.1",
+)
+
+@app.get("/")
+def read_root():
+    return {"Hello Viking": "Welcome to Warnet!"}
+
+@app.post("/warnet")
+def read_root():
+    return {"Hello": "warnet"}
```
okjodom: ❔ : it's unclear to me how we'd execute warnet automated docker functions from this api which is also hosted within a docker environment

okjodom (commented):
tACK. Works as proposed.
Needs clarification o how a docker wrapped api can run `warnet`

api-compose.yml
Extheoisah: Agree. This is for local dev. We're hardcoding some secrets in others parts of the code too. I'll create an issue for it

PR #60, opened 2023-08-22
update readme

willcl-ark:
remove unused attributes from graph


pinheadmz:
I think we can remove even more from the graph file, see how skinny I made it in the Tor PR: https://github.com/bitcoin-dev-project/warnet/blob/4d56920d545e8c907286ea2ce063466dd95700ca/graphs/basic3.graphml

willcl-ark:
Yes, will do in a followup

Issue #62, opened 2023-08-23
Create channel opening flow 


josibake:
we need:

* a weighted edge - direction is the channel open and weight is the capacity (UTXO)
* need a new "connect nodes" function. This will need to:

Now we need a scenario for opening channels:

* need a miner node to mine X liquidity for the network
* send a UTXO per channel to the funding lighting node (the funding node is defined by the direction of the edge)
* create a readiness check func - node has nonzero balance + synced to chain
* create a lightning_rpc() function (similar to bitcoin_rpc)
  * this function dispatches a bash command for impl specific channel open
  * create all the channels
* mine 6 blocks

now run a warnet heartbeat scenario? so that the network has some activity

josibake:
@carlaKC can you provide the bash templates for each lightning implementation for opening channels? we might also need templates for "polling" , if need that.

Idealy, each template takes the same arguments: im imagining a python function that takes 3 args and then that gets translated to each impl specific

carlaKC:
Instead of switching inside of warnet, each lightning container will have some bash scripts which call the specific commands that they need to perform an operation. This way, warnet can just call a generic function signature and each container will handle it's own CLI etc. 

The functions we'll define are: 
- `node_ready() -> pubkey`: returns the public key of the node when it is synced to the chain and is ready to open a channel, returns an error otherwise.
- `open_channel(pubkey, amt, ip:port) -> txid` : open a channel to the node specified at the IP given, returning the funding transaction ID.
- `channel_ready(txid)`: returns an error until the channel with the funding txid provided is known to the node.

PR #66, opened 2023-08-24
scripts: add lnd scripts for open channel

carlaKC:
Scripts for #62: 
* Return value used for error handling
* Output echoed where we need txids etc

This assumes that the CLI will be installed in the container (most images have it) + the LND node is running with default ports. 

*requires JQ installed*


carlaKC:
Closing for now, will re-open when I've got some capacity to work on this!

PR #73, opened 2023-08-24
warnetd: create a warnet rpc server

okjodom:
## Proposal

- adds a `warnetd` rpc service which executes warnet scripts on host machine
- updates `warnet` cli to use the underlying rpc

## Usage

- run `pip install -e .` to install `warnet` and `warnetd`
- run `warnetd` to start the rpc server
- run `warnet` to issue any cli commands 

## Further considerations
- [ ] All warnetd api calls are blocking. Need to make it threaded server, possibly with sockets
- [ ] If there is need for streaming data to/from warnetd, we should consider refactoring to a more suitable rpc framework like grpc


josibake:
ACK, will try to test later today!

willcl-ark:
Nice! Gave it a quick run through, it seemed to be mostly working :)

I took a stab at a branch based on this one which could daemonise `warnetd`, and add ability to kill the daemon via RPC (along with logging to a file): https://github.com/bitcoin-dev-project/warnet/compare/main...willcl-ark:warnet:warnetd_killable

It took a little more work that I thought, but could be a decent enough route if we so choose.

Typer really starting to frustrate me now though, not sure what's the best option there... Also our method names could do with a bit of unification and general tidying, but all that is unrelated to this pull.

willcl-ark:
Added even more commits on top now, so I hope others can chime in and say they're happy to merge this one 🤞🏼

PR #74, opened 2023-08-24
refactor, clean up and classify

pinheadmz:
Massive changes:

Codebase organization:
- Combine build & release dockerfiles for better maintenance and caching during builds
- Move bitcoin.conf, Dockerfile and the example graph into new directory `templates/`
- Create new classes `Warnet` and `Tank`
- Move all existing warnet configuration code from rpc_utils, docker_utils, etc into those two classes and clear out files
- Move all remaining code into `src/warnet/utils.py`

Program flow:
- Use global `logging` for literally everything (including `docker-compose up` output!), no more `print()`
- Use a tmpdir to save all ephemeral files like `bitcoin.conf.******` and any `.yml` file we create
- Edges from the graph will (once again) connect nodes using `addnode` instead of `addpeeraddress`
    - We'll have a DNS seeder soon so nodes can truly discover their own peers
    - That will probably make graph-defined edges mostly unnecessary except for specific experiments
- The `run` command is now used to run scenarios (also `warnet list` will list known scenario scripts)
- Use `start` to start warnet, e.g. `warnet start from-file ./src/templates/example.gaphml`

TODO:
- [ ] I haven't looked at the API/frontend stuff yet but it would be cool if we could clean up `Dockerfile_api` and `api-compose.yml` I think they should move out of root directory if they can
- [ ] Clean up the test_framework stuff for easier future maintnence
- [ ] Scenarios aren't quite happy with the new CLI yet


templates/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# Base requirements
+RUN apt-get update && apt-get install -y \
+        ccache \
+        python3 \
+        vim \
+        iproute2 \
+        build-essential \
+        wget \
+        iproute2; \
```
willcl-ark: doesn't matter but this iproute2 is duped

src/warnet/utils.py
```diff
@@ -27,3 +31,156 @@ def wrapper(*args, **kwargs):
...
+    """
+    Sanitize the tc-netem command to ensure it's valid and safe to execute, as we run it as root on a container.
+
+    Args:
+    - command (str): The tc-netem command to sanitize.
+
+    Returns:
+    - bool: True if the command is valid and safe, False otherwise.
+    """
+    if not command.startswith("tc qdisc add dev eth0 root netem"):
```
willcl-ark, line 99: I have been meaning to add something that can detect if we are using `eth0` in the container (which does seem to be the default) before we apply these commands.... maybe next week tho

src/scenarios/miner_std.py
```diff
@@ -26,7 +27,7 @@ def run_test(self):
             miner = ensure_miner(self.nodes[current_miner])
```
willcl-ark, line 27: would be nice to have a controller container who's job it was to run scenarios like this, so that the docker host  could disconnect and leave things running

src/warnet/client.py
```diff
@@ -58,18 +58,18 @@ def get_messages(src_index: int, dst_index: int):
     messages.sort(key=lambda x: x["time"])
     return messages
 
-def stop_network(network: str = "warnet"):
```
willcl-ark, line 69: I think this re-limits us to one warnet per docker host, which is something other said we didn't want to limit ourselves to...?

src/warnet/main.py
```diff
@@ -65,13 +72,31 @@ def messages(node_a: int, node_b: int):
     except Exception as e:
         typer.echo(f"Amidst the fog of war, we failed to relay messages between strongholds {node_a} and {node_b}: {e}")
 
+@warnet_app.command()
```
willcl-ark, line 75: yay, these were the next two commands I was going to add 🥇

willcl-ark (approved):
Seems to work pretty nicely, and I am a big fan of the organisation.

Didn't check where it comes from, but this debug logging seems a bit overkill:

```
2023-08-25 09:24:02 [INFO] Using `addndode` to connect tanks 2 to 0
2023-08-25 09:24:02 [DEBUG] Trying paths: ['/Users/will/.docker/config.json', '/Users/will/.dockercfg']
2023-08-25 09:24:02 [DEBUG] Found file at path: /Users/will/.docker/config.json
2023-08-25 09:24:02 [DEBUG] Found 'auths' section
2023-08-25 09:24:02 [DEBUG] Auth data for https://index.docker.io/v1/ is absent. Client might be using a credentials store instead.
2023-08-25 09:24:02 [DEBUG] Found 'credsStore' section
2023-08-25 09:24:02 [DEBUG] http://localhost:None "GET /version HTTP/1.1" 200 None
2023-08-25 09:24:02 [DEBUG] http://localhost:None "GET /v1.43/containers/tank_000002/json HTTP/1.1" 200 None
2023-08-25 09:24:03 [DEBUG] http://localhost:None "POST /v1.43/containers/ad57ad3dd84d61627233b673e78591560e040a354b1cf0c90a6d44c9195a6823/exec HTTP/1.1" 201 None
2023-08-25 09:24:03 [DEBUG] http://localhost:None "POST /v1.43/exec/c952320bf83c435a83454138d442d25f89d2c7e754123a19974ae8fedb058882/start HTTP/1.1" 101 0
2023-08-25 09:24:03 [DEBUG] http://localhost:None "GET /v1.43/exec/c952320bf83c435a83454138d442d25f89d2c7e754123a19974ae8fedb058882/json HTTP/1.1" 200 None
```

src/warnet/client.py
pinheadmz: ok ill revert this and make the necessary changes elsewhere

willcl-ark: Ok. I did rebase the dns seed on this and @okjodom may be rebased now too, so try and be nice 😋

pinheadmz:
> Didn't check where it comes from, but this debug logging seems a bit overkill:

I set the default log level to DEBUG, which is spammed by the docker python library we use I think. I'll change it back to INFO. There might be a few thiings that get spammy when we scale to >1000 nodes

willcl-ark:
You can change the level of one logging object too

willcl-ark:
Quick eg from GPT

```python
import logging

# Set up basic configuration to DEBUG for all loggers in the application
logging.basicConfig(level=logging.DEBUG, format='%(name)s - %(levelname)s - %(message)s')

# Now, we'll turn down the logging level for the Docker logger to INFO
docker_logger = logging.getLogger('docker')
docker_logger.setLevel(logging.INFO)

# Testing the loggers
logging.debug("This is a debug message from the main application.")  # This will be displayed
docker_logger.debug("This is a debug message from the Docker logger.")  # This will not be displayed
docker_logger.info("This is an info message from the Docker logger.")
```

templates/Dockerfile
pinheadmz: fixed

pinheadmz:
@willcl-ark pushed a commit to address nit and add network to all cli commands and new classes

PR #80, opened 2023-08-28
WIP: Add bind9 dns seeder

willcl-ark:
TODO:

* [ ] read SOA records if warnet running, and increment them
* [ ] reroute traffic destined for hardcoded DNS seeds to the `dns-seed` container

### DIFF
--- a/src/templates/Dockerfile_bind9
+++ b/src/templates/Dockerfile_bind9
@@ -1,0 +1,5 @@
+FROM ubuntu/bind9:9.16-20.04_beta
+COPY dns-seed.zone /etc/bind/dns-seed.zone
+COPY named.conf.local /etc/bind/named.conf.local
+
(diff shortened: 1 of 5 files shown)


willcl-ark:
@josibake @pinheadmz:

>  reroute traffic destined for hardcoded DNS seeds to the dns-seed container

This will obviously only work on Signet (or TestNet or MainNet), which is fine, I think. It does make me wonder though, do we actually want this to run on mainnet mode, so that we exercise as many mainnet codepaths as possible. I don't think there are _that_ many `if (ChainType::REGTEST) ...` cases in the codebase, but I'm also not totally sure about that either...

Another possibility could be to re-engineer this so that the nodes ran on mainnet (and capture dnsseed requests in this bind9 container, along with all other outbound requests?

PR #87, opened 2023-08-31
WIP: add fork-observer

willcl-ark:
Currently using my own docker image and fork which we'd want to switch to @0xB10C s image if this change https://github.com/0xB10C/fork-observer/pull/15 moves ahead...

Also these commits seems mangled, will fix soon

### DIFF
--- a/src/templates/fork_observer_config.toml
+++ b/src/templates/fork_observer_config.toml
@@ -1,0 +1,30 @@
+# fork-observer base configuration file
+
+# Database path of the key value store. Will be created if non-existing.
+database_path = "db"
+
+# path to the location of the static www files
+www_path = "./www"
+
+# Interval for checking for new blocks
+query_interval = 1
+
+# Webserver listen address
+address = "0.0.0.0:2323"
+
+# Custom footer for the site.
+footer_html = """
+    <div class="my-2">
+      <div>
+        <span class="text-muted">Warnet fork-observer</span>
+      </div>
+    </div>
+  """
+
+[[networks]]
+id = 0xDEADBE
+name = "Warnet"
+description = "A custom regtest network"
+min_fork_height = 0
+max_interesting_heights = 1
+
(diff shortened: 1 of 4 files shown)


src/templates/fork_observer_config.toml
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+    </div>
+  """
+
+[[networks]]
+id = 0xDEADBE
+name = "Warnet"
+description = "A custom regtest network"
+min_fork_height = 0
+max_interesting_heights = 1
```
0xB10C, line 29: Blaming myself for missing documentation, but I think you definitively want a higher number here.

See e.g. this comment https://github.com/0xB10C/fork-observer/blob/bde6c4ddd5d8deddd46642ea504a349104914842/src/headertree.rs#L55-L58

0xB10C:
I'm curious, what do you see when opening the fork-observer frontend?

willcl-ark:
It looks like this. I needed to start from height 0, but agree that perhaps interesting should be higher...

image

src/templates/fork_observer_config.toml
```diff
@@ -0,0 +1,30 @@
+# fork-observer base configuration file
+
+# Database path of the key value store. Will be created if non-existing.
+database_path = "db"
+
+# path to the location of the static www files
+www_path = "./www"
+
+# Interval for checking for new blocks
+query_interval = 1
```
0xB10C, line 10: I have this a 10 seconds for my nodes. Every second might be overkill. Maybe do 5s intervals?

0xB10C:
(manually) pushed a recent image to https://hub.docker.com/r/b10c/fork-observer

This will happen automatically once https://github.com/0xB10C/fork-observer/pull/17 is merged.

 Thanks @willcl-ark for picking this up!

0xB10C:
Images are now on https://hub.docker.com/r/b10c/fork-observer

PR #90, opened 2023-09-03
WIP move to ruimarinho images + other changes

### LINKED ISSUE #79: spin-up a fork-observer instance in the warnet docker-compose
The [fork-observer](https://github.com/0xb10c/fork-observer) tool visualizes the header tree of multiple nodes and which header/blocks nodes consider to be active/stale. This allows tracking forks and reorgs. For warnet, it could also highlight _stuck_  nodes. A live version of fork-observer is available on https://fork.observer.

image

While the tool is still  work-in-progress, I think it makes sense to have the option to spin up an instance along side the Bitcoin nodes and the grafana dashboard. The the configuration file for fork-observer can be generated from the XML file with the node information.

### LINKED ISSUE #86: Don't mount files from /tmp/ directories
Currently we mount volumes (files) that are created and live in /tmp/ which are obviously ephemeral and could be erased at any time, don't do this:

https://github.com/bitcoin-dev-project/warnet/blob/9888538716f262b1441fb0751a83dfe380f67846/src/warnet/warnet.py#L119-L128

(I am in the process of adding a second like this, fork-observer)

willcl-ark:
* [x] closes #87
* [x] closes #79
* [x] ruimarinho images
* [x] dedicated alpine Tor Directory Authority container (currently running, but misconfigured)
* [x] fork-observer alpine container, now being pushed to https://hub.docker.com/r/b10c/fork-observer/tags automagically
* [x] "downgrade" from Typer to Click (which Typer is based on, but is more powerful/configurable)
* [x] use a config dir for each warnet (closes #86 ), generated with:
```python
def gen_config_dir(network: str) -> Path:
    """
    Determine a config dir based on network name
    """
    config_dir = os.getenv("XDG_CONFIG_HOME", os.path.expanduser("~/.warnet"))
    config_dir = Path(config_dir) / "warnet" / network
    return config_dir
```
* [x] updated README

Would love to see some testing on this from others, if possible, i'd recommend something like:

```bash
pip uninstall warnet
pip freeze | xargs pip uninstall -y
docker ps -aq | xargs docker stop

# WARNING WILL REMOVE ALL DOCKER EVERYTHINGS!
# Use other methods to prune old warnet images if you have other docker stuff you don't want to lose!
docker system prune -a

pip install -e .
```

Then README.md should guide you through the process :)

I plan to 

- [x] re-add the dns server
- [x] fluentd log manager (not working properly :'( )


willcl-ark:
The annoying thing about ruimarinho images is they are 2 versions out of date :'( But I plan to host our own warnet images (with tor and the other bits we want pre-installed), so can easily-enough make our own v24.1 and v25.0 images at that time I think...

willcl-ark:
I've opened https://github.com/ruimarinho/docker-bitcoin-core/pull/144 to try and fixup that repo, so we can use it here :)

willcl-ark:
Now includes some @pinheadmz fixups

PR #94, opened 2023-09-08
WIP: use dummy.invalid as dns seed

willcl-ark:
This fixes the DNS seed for regtest by re-routing the regtest dummy address: 

https://github.com/bitcoin/bitcoin/blob/238d29aff9b43234e340a9cf17742b2be5d1e97d/src/kernel/chainparams.cpp#L471

to query our dns-seed.


willcl-ark:
~~At the moment we are relying on a sleep hack to start bitcoind, otherwise it tries the dns before bind9 is started. This might be able to be scripted more efficiently if we can ping the dns-ssed and not stat up until we start getting responses...~~

Added this now

Issue #85, opened 2023-08-30
frontend:swtich ui library to react-flow??


Extheoisah:
d3.js is not as flexible as we would want it for the node and network mapping.
A good alternative which is being used by [Polar](https://github.com/jamaljsr/polar) is [react-flow](https://reactflow.dev/) which seems to have better documentation and is built for react, hence the flexibility.

PR #97, opened 2023-09-09
Feat: a react flow migration  from d3

0tuedon:
This closes the issue of #85  where the frontend should switch to react flow


Emmanuel-Develops (commented):
Looks good! Nice work.

a couple of bugs:
- editing and duplicating nodes
- can't generating node graph

@Extheoisah should we merge this then fix the bugs as a separate PR?
This PR is a bit bulky as it contains my changes and @0tuedon's conversion to react-flow

0tuedon:
> Looks good! Nice work.
> 
> a couple of bugs:
> 
> * editing and duplicating nodes
> * can't generating node graph
> 
> @Extheoisah should we merge this then fix the bugs as a separate PR? This PR is a bit bulky as it contains my changes and @0tuedon's conversion to react-flow

so for the generating of node graph, i did comment that out because i was not sure that would work with the change of node structure.  also @Extheoisah  said i should fix something about the home page disappearing

Extheoisah:
I'm fine with fixing the bugs in a follow up pr so this doesn't get too large than it already is. However, we can't merge without the homepage commit which is missing in this pr. I discussed that with @0tuedon already

0tuedon:
> I'm fine with fixing the bugs in a follow up pr so this doesn't get too large than it already is. However, we can't merge without the homepage commit which is missing in this pr. I discussed that with @0tuedon already

yh  @Extheoisah  as discussed i have returned the home page to the index and  move the main node pages to /start

PR #99, opened 2023-09-11
add container resource profiles

### LINKED ISSUE #26: Define node personas in the graph file
Such as:

* Raspi - low RAM, low CPU
* Wumbo node  - big CPU, big RAM
* bad latency
* regularly offline (pops up every now and again)

These are basically things that cannot be defined in bitcon.conf

willcl-ark:
Closes: #26

Adds a few demo profiles such as laptop, desktop, raspberry_pi, server to the graph config.

While we are using docker-compose, and not swarm, we cannot easily limit physical CPUS (withough downgrading to docker-compose v2), but we can get "relative resourse usage", and memory limitations.

This line:
```python
"cpu_shares": int(self.resource_profile.get("cpus", "0.5")) * 1024,
```

 is the temporary "hack" which converts our cpu number from a profile to a rough `cpu_shares` value (which docker-compose accepts).
 
 Link: https://docs.docker.com/config/containers/resource_constraints/

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ The graphml file has the following specification:
 * `y` specifies the node's y position when rendered in a GUI
 * `version` specifies the node's Bitcoin Core major version, or built branch
 * `bitcoin_config` is a comma-separated list of values the node should apply to it's bitcoin.conf, using bitcoin.conf syntax
+* `profile` is a profile name from the pre-built selection in src/warnet/resources.py
 
 `version` should be either a version number from the pre-compiled binary list on https://bitcoincore.org/bin/ **or** a built branch using `<user>/<repo>#<branch>` syntax.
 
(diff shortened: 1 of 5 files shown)


josibake:
Concept ACK

Need to look at this with the k8s backend, but I think it should work by having the resource profiles and then each backend translates that into something the backend can understand

willcl-ark:
If we want ot continue with tc_netem, then I think it just needs a big rebase.

If we want to pass it as actual pod resource profile to the cluster (can you even do that), then it might need a total re-write...

josibake:
You can set pod resource limits (CPU, memory) directly when defining the pod in k8s. But we could also use tc_netem throughout, because if I remember correctly that's happening inside a docker container? So it should work on both and iirc, it gives more options than just CPU/RAM.

Maybe we just stick with tc_netem through for now? Longterm, I can see value in having the profile be a "warnet" concept, and then each backend translates the profile into what the backend needs

PR #110, opened 2023-09-12
Persist bitcoin datadir locally using volume

willcl-ark:
This will enable better debugging, network snapshots, bringing networks up and
down, replaying scenarios etc.


pinheadmz:
Am learning that volumes might be one of our scaling bottlenecks and "bindmounts" might be better

willcl-ark:
Closing as I don't think we want/need this

PR #116, opened 2023-09-12
fix editing, adding, duplicating node frontend

### LINKED ISSUE #104: frontend:editing a node property is broken

Emmanuel-Develops:
closes #104

### DIFF
--- a/frontend/src/contexts/network-context.tsx
+++ b/frontend/src/contexts/network-context.tsx
@@ -25,10 +25,11 @@ export const networkContext = React.createContext<NetworkContext>(null!);
 const newNode = {
   id: "0",
   data: {
-  name: "new node",
-  label:"new node",
-  size: 10,
+    id: "0",
+    label:"new node",
+    size: 10,
   },
+  type: "draggable",
   position: {
     x: CANVAS_WIDTH / 2,
     y: CANVAS_HEIGHT / 2,
--- a/frontend/src/contexts/node-flow-context.tsx
+++ b/frontend/src/contexts/node-flow-context.tsx
@@ -33,7 +33,6 @@ export const NodeGraphFlowProvider = ({
   const closeDialog = () => {
     setIsDialogOpen(false);
     setNodeInfo(null);
-    // setSteps(-1);
   };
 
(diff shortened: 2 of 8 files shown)


frontend/src/contexts/network-context.tsx
```diff
@@ -25,10 +25,11 @@ export const networkContext = React.createContext<NetworkContext>(null!);
 const newNode = {
   id: "0",
   data: {
-  name: "new node",
-  label:"new node",
-  size: 10,
+    id: "0",
```
0tuedon, line 28: i see duplicates keys (id).

0tuedon (commented):
Nice work, left some comments.

frontend/src/contexts/network-context.tsx
Emmanuel-Develops: The `data` property, has a field `id` which is the `id` here. I made the `id` field mandatory (it was an optional type previously). The formatting makes it look like a duplicate, its a field under data.

Discussions on wether the field is redundant is a different matter. (I'm indifferent)
I think if we want separation of concern between node properties for the canvas (id, type, position) and node data itself, we can leave the second `id` field to reference the node.

There's a specific instance where separation of concern can simplify the editing fnc but its not a priority at the moment so we can leave the `id` as is for now.

frontend/src/contexts/node-flow-context.tsx
```diff
@@ -68,82 +63,81 @@ export const NodeGraphFlowProvider = ({
...
   const updateNodeInfo = (nodeProperty: any, value: any) => {
-    if (!nodeInfo) return
-    const duplNode = {...nodeInfo}
+    if (!nodeInfo) return;
+    const duplNode = JSON.parse(JSON.stringify(nodeInfo));
     //@ts-ignore partia will come back to it
-    duplNode.data[nodeProperty]  = value
-    setNodeInfo(duplNode)
-  }
+    duplNode.data[nodeProperty] = value;
```
Extheoisah, line 117: I guess we no longer need the ts ignore since node data is no longer a partial

Extheoisah (approved):

PR #117, opened 2023-09-12
Set config dir when using from_network classmethod

willcl-ark:
Stops fork_observer.toml being written to source dir on shutdown


src/warnet/warnet.py
```diff
@@ -90,6 +90,7 @@ def from_graph(cls, graph):
     def from_network(
         cls, config_dir: Path = Path(), network: str = "warnet", tanks=True
     ):
+        config_dir = gen_config_dir(network)
```
pinheadmz, line 93: this would override the passed-in argument. is it still a required param to from_network?

PR #123, opened 2023-09-13
Scenarios fix

pinheadmz:
1) remove gunicorn, refactor warnetd to class Server
2) refactor scenarios so we can track them in the server instance

[asciicast](https://asciinema.org/a/DUGDutULLkOz2S9FlcUArDfRO)


willcl-ark:
Looking nice and clean so far!

pinheadmz:
Pushed third commit which fixes the warnet from docker env issues and patched up a few other scenarios issues.

willcl-ark:
ACK from me. Got some additional cleanup commits here: https://github.com/bitcoin-dev-project/warnet/compare/scenarios-fix...willcl-ark:warnet:scenarios-fix-fixups

* remove a few unused imports
* we can remove `--network` from most scenario calls
* but we can add it to the `active` list so we know which network a scenario is running on
* `.terminate()` sends SIGTERM to the process, but, at least for me, it seemed like we were not catching it and they weren't _actually_ being stopped (according to htop). Now we handle the signal, gracefully calling `self.shutdown()` before exiting. Also, actually remove ourselves from the active list :)

^ this last one made me wonder whether we should also set `--nocleanup` as defaulting to `True`, so that we always save the logdir?

Feel free to cherry-pick in here, or I can open a followup PR

willcl-ark (approved):

josibake:
LGTM, I say we merge and then @willcl-ark can open a PR with his commits and I can rebase the scale-better branch on this

PR #124, opened 2023-09-14
A few fixups for the new scenarios

willcl-ark:


pinheadmz:
tested ACK and code review ACK and will is handsome DOUBLE-ACK

PR #127, opened 2023-09-14
fix and shorten miner wrap-around

willcl-ark:
For me (TM) without this miner dies after the first iteration when executing:

`warcli scenarios run miner_std --allnodes --interval=2`

### DIFF
--- a/src/scenarios/miner_std.py
+++ b/src/scenarios/miner_std.py
@@ -38,9 +38,7 @@ class MinerStd(WarnetTestFramework):
             block = self.generatetoaddress(self.nodes[current_miner], 1, addr)
             self.log.info(f"generated block from node {current_miner}: {block}")
             if self.options.allnodes:
-                current_miner = current_miner + 1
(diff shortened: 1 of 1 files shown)


willcl-ark:
@pinheadmz if this is an issue for you too (and this patch fixes it) then send it

pinheadmz:
Hm this is not an issue for me on main at fe031c7bfb0549df0a74455f7de2da7dab02c782
How could we be getting different behaviors here? Are you sure you were running on top of this commit? which fixed `self.num_nodes`

willcl-ark:
Yep, still dies on current main @ 215f4ca for me. So odd...

image

pinheadmz:
I see you have a netem setting for node 0. Could that be delaying the RPC responses?

willcl-ark:
Still dies with no netem @pinheadmz 

image

willcl-ark:
Cherry-picking this commit it wraps around just fine?

image

PR #128, opened 2023-09-14
fixup readme to match current operation

willcl-ark:
tiny commit to correctly apply tc-netem conditions too

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ Monitor and analyze the emergent behaviours of Bitcoin networks.
 ## Major functionality
 
 * Warnet uses docker to launch a bitcoin network of `n` nodes which are connected to each other according to a network topology from a specification file.
-* Nodes are by default assigned random IP addresses in the 100.0.0.0/8 subnet which ensures randomized `addrman` binning [by Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/8372ab0ea3c88308aabef476e3b9d023ba3fd4b7/src/addrman.h#L66), along with a Tor address.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -69,6 +70,8 @@ Or for a custom built branch with traffic shaping rules applied:
 </node>
 ```
 
+`x`, `y` and `tc_netem` datafields are optional for all nodes.
```
pinheadmz: so are `version` and `bitcoin_config`

pinheadmz (approved):

README.md
willcl-ark: What happens if `version` is not present? I feel like that should be (and on a quick look looks like it may be) UB?

willcl-ark: added that `bitcoin_config` is optional in force push

pinheadmz: check out the `__init__()` in tank.py, version starts off with a default of "25.0"

pinheadmz: (but if you think it should not have a default, go ahead and switch that to none in the constructor)

README.md
```diff
@@ -69,6 +70,8 @@ Or for a custom built branch with traffic shaping rules applied:
 </node>
 ```
 
+`x`, `y`, `bitcoin-config` and `tc_netem` datafields are optional for all nodes.
```
pinheadmz: nit `bitcoin_config`

willcl-ark: done

README.md
willcl-ark: done

pinheadmz (approved):
utACK (I didn't test out the netem code change) but README looks good now, merge at will, Will

PR #132, opened 2023-09-21
Generated graphml fix

Emmanuel-Develops:


Emmanuel-Develops:
cc: @Extheoisah @willcl-ark 

do we use `baseFee` anywhere?

Extheoisah:
> cc: @Extheoisah @willcl-ark
> 
> do we use `baseFee` anywhere?

don't think so. atleast we don't have it in the sample graphml file https://github.com/bitcoin-dev-project/warnet/blob/main/src/templates/example.graphml

PR #133, opened 2023-09-29
log: fix fluentd, add log-grepper

willcl-ark:
* Fluentd seems to only work listening on 0.0.0.0 and not an internal (warnet) network ip address.
* Implies a single docker daemon can only have a single flutend log driver.
* Search unified logs with `warcli grep-logs [regex pattern]`

### DIFF
--- a/src/services/fluentd.py
+++ b/src/services/fluentd.py
@@ -1,6 +1,7 @@
+import shutil
+from templates import TEMPLATES
 from .base_service import BaseService
 
-FLUENT_IP = "100.102.108.117"
 FLUENT_CONF = "fluent.conf"
 
 class Fluentd(BaseService):
@@ -9,16 +10,22 @@ class Fluentd(BaseService):
     def __init__(self, docker_network, config_dir):
         super().__init__(docker_network, config_dir)
         self.service = {
-            "image": "fluent/fluentd:v1.16-debian-1", # Debian version is recommended officially since it has jemalloc support.
+            "image": "fluent/fluentd:v1.16-debian-1",  # Debian version is recommended officially since it has jemalloc support.
             "container_name": f"{self.docker_network}_fluentd",
-            "ports": [f"{self.PORT}:{self.PORT}"],
+            "ports": [
+                f"{self.PORT}:{self.PORT}"
+            ],
             "volumes": [
                 f"{self.config_dir / FLUENT_CONF}:/fluentd/etc/{FLUENT_CONF}"
             ],
-            "command": ["/bin/sh", "-c", f"sleep 10 && fluentd -c /fluentd/etc/{FLUENT_CONF}"],
-            "networks": {
-                self.docker_network: {
-                    "ipv4_address": f"{FLUENT_IP}",
-                }
-            },
+            "command": ["/bin/sh", "-c", f"fluentd -c /fluentd/etc/{FLUENT_CONF}"],
+            "healthcheck": {
+                "test": ["CMD", "/bin/bash", "-c", f"cat < /dev/null > /dev/tcp/localhost/{self.PORT}"],
+                "interval": "5s",           # Check every 5 seconds
+                "timeout": "1s",            # Give the check 1 second to complete
+                "start_period": "2s",       # Start checking after 2 seconds
+                "retries": 3
+            }
         }
+
+        shutil.copy(TEMPLATES / FLUENT_CONF, config_dir)
--- a/src/templates/bitcoin.conf
+++ b/src/templates/bitcoin.conf
@@ -5,6 +5,7 @@ regtest=1
 # Debugging
 debug=1
 logips=1
+logtimemicros=1
 capturemessages=1
 debugexclude=libevent
 debugexclude=leveldb
--- a/src/templates/bitcoin.conf
+++ b/src/templates/bitcoin.conf
@@ -4,6 +4,7 @@ regtest=1
 
 # Debugging
 debug=1
+debuglogfile=0
 logips=1
 logtimemicros=1
 capturemessages=1
--- a/src/templates/fluent.conf
+++ b/src/templates/fluent.conf
@@ -4,6 +4,25 @@
   bind 0.0.0.0
 </source>
 
+<filter **>
+  @type record_transformer
+  enable_ruby
+  <record>
+    message ${record['container_name'].gsub(/^\//, '')}: ${record['log']}
+  </record>
+</filter>
+
 <match *>
   @type stdout
+  <format>
+    @type single_value
+    message_key message
+  </format>
 </match>
+
+<source>
+  @type http
+  port 24220
+  bind 0.0.0.0
+</source>
+
--- a/src/warnet/client.py
+++ b/src/warnet/client.py
@@ -1,6 +1,7 @@
 import concurrent.futures
 import logging
 import threading
+from datetime import datetime
 from typing import List, Optional, Any, Dict
 
 import docker
@@ -13,17 +14,18 @@ logger = logging.getLogger("warnet.client")
 
 def get_bitcoin_debug_log(network: str, index: int) -> str:
     warnet = Warnet.from_network(network)
-    tank = warnet.tanks[index]
-    subdir = "/" if tank.bitcoin_network == "main" else f"{tank.bitcoin_network}/"
-    data, stat = tank.container.get_archive(f"/home/bitcoin/.bitcoin/{subdir}debug.log")
-    out = ""
-    for chunk in data:
-        out += chunk.decode()
-    # slice off tar archive header
-    out = out[512:]
-    # slice off end padding
-    out = out[: stat["size"]]
-    return out
+    docker = warnet.docker
+    container_name = warnet.tanks[index].container_name
+    now = datetime.utcnow()
+
+    logs = docker.api.logs(
+        container=container_name,
+        stdout=True,
+        stderr=True,
+        stream=False,
+        until=now,
+    )
+    return logs.decode('utf-8')
 
 
 def get_bitcoin_cli(network: str, index: int, method: str, params=None) -> str:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,3 +31,11 @@ jobs:
       - run: pip install --upgrade pip
       - run: pip install -e .
       - run: ./test/v25_net_test.py
+  rpc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: sudo apt install -y docker docker-compose
+      - run: pip install --upgrade pip
+      - run: pip install -e .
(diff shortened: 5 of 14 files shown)


willcl-ark:
Sample:

image

Possibly to-decide:

* What format do we want these logs in? I've gone for the equivalent of 

```python
f"{bitcoind-log-line}, {docker_container_name}"
```

...which allows sorting by time, in case things arrived out of order, but open to other suggestions

pinheadmz:
Issues running this branch:


> 2023-10-03 13:36:35,558 - warnet - INFO - ERROR: for warnet_tank_000006  Cannot start service warnet_tank_000006: failed to create task for container: failed to initialize logging driver: dial tcp 127.0.0.1:24224: connect: connection refused

> 2023-10-03 13:36:35,699 - utils - ERROR - rpc error: 409 Client Error for http+docker://localhost/v1.43/containers/7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31/exec: Conflict ("Container 7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31 is not running")

> 2023-10-03 13:37:35,839 - warnet - ERROR - Exception Exception in function 'wrapper' with context (args=(<warnet.warnet.Warnet object at 0x107b353d0>,), kwargs={}, e=409 Client Error for http+docker://localhost/v1.43/containers/7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31/exec: Conflict ("Container 7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31 is not running"), func_name=wrapper, func=<function Warnet.connect_edges at 0x107a0aa20>): 409 Client Error for http+docker://localhost/v1.43/containers/7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31/exec: Conflict ("Container 7de684c019708ae36e966f8294442d2e8fea95f8a984ad2a4893ad965224cb31 is not running")

pinheadmz:
also this, fluentd container did not get deleted on dopcker compose down:

> 2023-10-03 13:43:54,779 - warnet - INFO - Found orphan containers (warnet_fluentd) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.

src/templates/fluent.conf
```diff
@@ -4,6 +4,18 @@
   bind 0.0.0.0
 </source>
 
+<filter **>
+  @type record_transformer
+  enable_ruby
+  <record>
+    message "${record['log'].split(' ')[0]}T${record['log'].split(' ')[1]} ${record['log'].split(' ')[2..-1].join(' ')}, ${record['container_name'].split('_')[1..-1].join('_')}"
```
pinheadmz: IMO container name should go first kinda like bitcoin test combined logs

pinheadmz:
this might be the source of my woes, none of the tanks are starting I assume because of the `depends_on`, and fluentd didnt start correctly?

> #0 dump an error event: error_class=RuntimeError error="failed to expand `%Q[\#{record['log'].split(' ')[0]}T\#{record['log'].split(' ')[1]} \#{record['log'].split(' ')[2..-1].join(' ')}, \#{record['container_name'].split('_')[1..-1].join('_')}]` : error = undefined method `split' for nil:NilClass" location="/usr/local/bundle/gems/fluentd-1.16.2/lib/fluent/plugin/filter_record_transformer.rb:310:in `rescue in expand'" tag="fluent.info" time=2023-10-03 17:50:55.957860299 +0000 record={"pid"=>14, "ppid"=>9, "worker"=>0, "message"=>"starting fluentd worker pid=14 ppid=9 worker=0"}

src/templates/fluent.conf
pinheadmz: simplify? works for me...

```diff
diff --git a/src/templates/fluent.conf b/src/templates/fluent.conf
index 8d99068..0af7898 100644
--- a/src/templates/fluent.conf
+++ b/src/templates/fluent.conf
@@ -8,7 +8,7 @@
   @type record_transformer
   enable_ruby
   <record>
-    message "${record['log'].split(' ')[0]}T${record['log'].split(' ')[1]} ${record['log'].split(' ')[2..-1].join(' ')}, ${record['container_name'].split('_')[1..-1].join('_')}"
+    message "${record['container_name']}: ${record['log']}"
   </record>
 </filter>
 

```

pinheadmz: https://gist.github.com/pinheadmz/477ea23854bf510d2acad929bf15a20a

src/services/fluentd.py
```diff
@@ -16,9 +18,6 @@ def __init__(self, docker_network, config_dir):
                 f"{self.config_dir / FLUENT_CONF}:/fluentd/etc/{FLUENT_CONF}"
             ],
             "command": ["/bin/sh", "-c", f"sleep 10 && fluentd -c /fluentd/etc/{FLUENT_CONF}"],
```
pinheadmz: don't need the `sleep 10`

pinheadmz:
bravo dude!
concept ack. just a few suggestions. very cool feature

src/templates/fluent.conf
willcl-ark: Done in [769dae9](https://github.com/bitcoin-dev-project/warnet/pull/133/commits/769dae9cab13f96a963f69446174538222d3a049)

src/services/fluentd.py
willcl-ark: yep, removed in [769dae9](https://github.com/bitcoin-dev-project/warnet/pull/133/commits/769dae9cab13f96a963f69446174538222d3a049)

willcl-ark:
I added a commit disabling log files from bitcoin Core, but anyway...

@pinheadmz @josibake IMO we still need some high-level conceptual review on this logging system, now that I have finally got it working, I feel less sure about it:

* Do we use this docker logger _driver_?

    My current understanding is that this limits us to one fluentd instance **per docker daemon**, and therefore if we _do_ want to use this, we should probably think about having warnet `Server` be the one who spins up the fluentd container (and maybe others), rather than each `Warnet`. We can then prepend `<network_name>` to the logs, collect alllllll the logs from alllllll the Warnets on a single container, and do filtering by network inside the RPCs.
    
    This might not actually want to be the way we want to scale though. I can imagine a single fluentd container listening to say, 3 large warnets, may start to get overwhelmed. If that was 5-10k nodes per warnet, you're listening, through docker-daemon networking magic, to 15-30k STDOUTs at the same time. I don't know how this will work.
    
    If our medium-term scaling plans include scaling to different physical hosts, it probably makes more sense to take advantage of their multiple cpus/disks and just write logs to (local) files, and minimise network contention. Promtail in theory can tail these files (in it's own time?) if we really want centralised aggregation. But perhaps the most scalabale of all is a combination of:
    
    i) Let the user access any direct log file (easy to copy, manual grep etc) on each container. we have an RPC for this, and they can SSH in or w/e
    ii) Provide a "search log aggregate" RPC which will under the hood despatch _individual (`docker exec`) `grep` commands to each container_, and just aggregate the results right before return.
    
    This way, if we scale with more physical hosts, the locally-despatched commands will be using different CPUs and disks...
   
Thoughts?

willcl-ark:
I did poke around with https://github.com/grafana/loki a little but, but not to a working state. But having a Promtail per container push its logs to a central Loki/grafana is another potential approach. Not really sure how it scales vs fluentd-as-a-logging-driver for example

src/templates/bitcoin.conf
```diff
@@ -4,7 +4,9 @@ regtest=1
 
 # Debugging
 debug=1
+debuglogfile=0
```
pinheadmz, line 7: This makes sense but breaks warcli debug_log. I think that's ok if maybe we have a new log category of commands that can use the same docker.api.log trick just target individual tanks

willcl-ark: ah yes, thank for thinking of that. If only we had a CI or something that could alert me of such cases.... :P

I think I'll push an additional commit using `docker.api.log` as you suggest

src/services/fluentd.py
willcl-ark: Going to revert this, seems to be some magic-fu which makes things work... Not sure why, not sure I have time to work out why...

src/templates/bitcoin.conf
willcl-ark: Ok now done this in [a77ac80](https://github.com/bitcoin-dev-project/warnet/pull/133/commits/a77ac80af62dc732284e5b6caf108abe6f1f9536)

src/warnet/client.py
```diff
@@ -14,17 +15,18 @@
...
+    docker = warnet.docker
+    container_name = warnet.tanks[index].container_name
+    now = datetime.utcnow()
+
+    logs = docker.api.logs(
+        container=container_name,
+        stdout=True,
+        stderr=True,
+        stream=False,
+        until=now,
```
pinheadmz, line 26: Ooh theres other good options for this: https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.logs

like `since=` (i.e. warcli option for logs "since `n` minutes ago")  and `timestamps=False` (since bitcoin core prints timestamps already?)

pinheadmz:
Ok this fixed everything for me. See https://docs.docker.com/config/containers/logging/fluentd/#fluentd-async

TLDR: Tanks depend on fluentd. Apparently `depends_on` only waits for the fluentd container to *exist*, so if the fluentd server isnt ready yet, the tanks just fail to build. Adding this async option buffers log output until the connection is ready. I tried it out and it does take a few seconds for fluentds combo-log to start tracking but at least docker-compose succeeds

```diff
diff --git a/src/warnet/tank.py b/src/warnet/tank.py
index f22da0d..f888ccc 100644
--- a/src/warnet/tank.py
+++ b/src/warnet/tank.py
@@ -246,7 +246,8 @@ class Tank:
                 "logging": {
                     "driver": "fluentd",
                     "options": {
-                        "tag": "{{.Name}}"
+                        "tag": "{{.Name}}",
+                        "fluentd-async": "true"
                     }
                 }
             }

```

willcl-ark:
> Ok this fixed everything for me. See [docs.docker.com/config/containers/logging/fluentd/#fluentd-async](https://docs.docker.com/config/containers/logging/fluentd/#fluentd-async)
> 
Yeah I had this in a few test branches, back when I couldn't get the thing to work. TBH I'm not that much of a fan of it, as I understand it, and I think I would rather have a healthcheck on fluentd, which I will investigate momentarily...

With the async stuff fluentd might just never start up?

pinheadmz:
rebased on main to run CI

pinheadmz:
added test and docs!
ACK from me, @willcl-ark check out my additions and merge at will, Will.

willcl-ark:
Nice, done!

PR #140, opened 2023-10-06
defuse unsafe xml

willcl-ark:


willcl-ark:
Lack of interest/need.

PR #141, opened 2023-10-06
test: always kill warnet server

willcl-ark:
Currently, if a test fails warnet server might not be properly killed and so remain running.

This is annoying for iteration; if the test fails we want everything gone ready to run again.

Also only try to bring the network down if we first brought it up.

### DIFF
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -1,4 +1,6 @@
 import atexit
+import os
+import signal
 from pathlib import Path
 from subprocess import Popen, run, PIPE
 from tempfile import mkdtemp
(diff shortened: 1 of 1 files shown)


test/test_base.py
```diff
@@ -17,47 +19,66 @@ def __init__(self):
...
-        print("\nRemaining server output:")
-        print(self.logfile.read())
-        self.logfile.close()
+        self._cleanup_network()
+        self._cleanup_server()
 
-        self.logfile = None
-        self.server.terminate()
-        self.server = None
+        if self.logfile:
```
willcl-ark, line 68: Don't 100% understand why this can be None, but it was often for me?

test/test_base.py
```diff
@@ -17,47 +19,66 @@ def __init__(self):
...
+            self.logfile.close()
+            self.logfile = None
 
 
     # Execute a warcli RPC using command line (always returns string)
     def warcli(self, str, network=True):
         cmd = ["warcli"] + str.split()
         if network:
             cmd += ["--network", self.network_name]
+        if "network" in cmd and "start" in cmd:
```
willcl-ark: Very rudimentary, also 11 lines below.

test/test_base.py
pinheadmz: Sometimes it gets called twice because of atexit. I want to garuntee it always gets run at least once when we quit

willcl-ark:
@pinheadmz FYI all tests passing now (just restarted to pass).

I also covered both startup CLI/RPC commands, i.e. `start`, and `up` for CLi , and `from_network` and `up` for RPC.

test/test_base.py
```diff
@@ -17,45 +19,63 @@ def __init__(self):
...
+                # If the warnet server exited prematurely then docker-compose down
+                # likely did not succeed or was never executed.
+                print(f"Error stopping server: {e}")
+                self._cleanup_docker()
+
+    def _cleanup_server(self):
+        # kill the server process by its PID if it still survives
+        if self.server:
+            print(f"killing server with pid {self.server.pid}")
+            os.killpg(os.getpgid(self.server.pid), signal.SIGTERM)
```
pinheadmz, line 59: this looks to me like the most useful part of this PR, I dont really understand the need for the other changes. Why couldn't this emergency termination just be part of the original `cleanup()` in the exception handler (i.e. if `warcli stop` fails?)

PR #142, opened 2023-10-13
build: switch to docker compose plugin by default

willcl-ark:
This should avoid docker-compose binary mismatches from system repos with old versions (looking at you Debian).

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -11,7 +11,7 @@ cp .env.sample .env
 2. Run the docker daemon
 
 ```bash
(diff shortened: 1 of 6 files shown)


pinheadmz:
thanks this works now on the new VPS. Gimme an hour or so to check it locally as well then we can merge

willcl-ark:
Cool. Thanks for the reminder as I had been meaning to switch it

pinheadmz (approved):
cool, works thanks!

PR #144, opened 2023-10-14
test: fix onion test variablility

willcl-ark:
An inelegant fix for the onion test randomly failing.

I would like to clear this up later(TM) in a move to pytest


pinheadmz:
LGTM thanks.

PR #145, opened 2023-10-17
updates to built branches

willcl-ark:
* add graph key `build_args` which will pass on build args to a built branch
* --enable-debug by default on built branches

@Emmanuel-Develops @Extheoisah  FYI changes to the graph file spec in this one.

note to self: will need corresponding update to the graph pull if this gets merged first

@pinheadmz obviously curious what you think of this :)

### DIFF
--- a/src/templates/example.graphml
+++ b/src/templates/example.graphml
@@ -2,6 +2,7 @@
   <key attr.name="version" attr.type="string" for="node" id="version"/>
   <key attr.name="bitcoin_config" attr.type="string" for="node" id="bitcoin_config"/>
   <key attr.name="tc_netem" attr.type="string" for="node" id="tc_netem"/>
+  <key attr.name="build_args" attr.type="string" for="node" id="build_args"/>
   <graph edgedefault="directed">
     <node id="0">
         <data key="version">25.0</data>
@@ -52,6 +53,7 @@
     <node id="11">
       <data key="version">vasild/bitcoin#relay_tx_to_priv_nets</data>
       <data key="bitcoin_config">sensitiverelayowntx=1,uacomment=sensitive_relay</data>
+      <data key="build_args">--disable-zmq --disable-fuzz</data>
     </node>
     <!-- provide addrs for 8 full + 2 block outbound connections -->
     <edge id="0" source="11" target="0"></edge>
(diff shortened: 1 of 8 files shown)


src/templates/example.graphml
```diff
@@ -52,6 +53,7 @@
     <node id="11">
       <data key="version">vasild/bitcoin#relay_tx_to_priv_nets</data>
       <data key="bitcoin_config">sensitiverelayowntx=1,uacomment=sensitive_relay</data>
+      <data key="build_args">--disable-zmq --disable-fuzz</data>
```
pinheadmz, line 56: if you put this in `test/data/build_v24_test.graphml` we could actually see the options in the test log:

https://github.com/bitcoin-dev-project/warnet/actions/runs/6544930178/job/17772400409?pr=145#step:6:318

src/warnet/tank.py
```diff
@@ -26,6 +26,8 @@
 
 
 class Tank:
+    DEFAULT_BUILD_ARGS = "--disable-tests --with-incompatible-bdb --without-gui --disable-bench --disable-fuzz-binary --enable-suppress-external-warnings --enable-debug "
```
pinheadmz, line 29: no real opinion on this, but why define this global inside the class instead of further up with the other globals?

src/warnet/warnet.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, config_dir):
         self.bitcoin_network: str = "regtest"
         self.docker_network: str = "warnet"
         self.subnet: str = "100.0.0.0/8"
-        self.graph = None
+        self.graph: Optional[networkx.Graph] = None
```
pinheadmz, line 39: Is this unrelated?

pinheadmz (commented):
concept ACK
a few questions

src/templates/example.graphml
willcl-ark: OK can do

src/warnet/tank.py
willcl-ark: Not really sure/no strong opinion either. Variables defined here exist once shared between all instances of the class, but only within the class namespace, which just seemed like exactly what we would want?

src/warnet/warnet.py
willcl-ark: Yup. I just fix the typechecker errors when I'm near by :)

src/templates/example.graphml
pinheadmz: I did.
Added a test

PR #146, opened 2023-10-17
warnet: generate as files

willcl-ark:
generates ASmap files and `COPY`s them into the containers

### DIFF
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -4,14 +4,17 @@
 
 import docker
 import logging
+import io
 import networkx
 import shutil
 import subprocess
+import sys
 import yaml
 from pathlib import Path
 from templates import TEMPLATES
 from typing import List, Optional
 
+import external.buildmap as buildmap
 from services.prometheus import Prometheus
 from services.node_exporter import NodeExporter
 from services.grafana import Grafana
@@ -21,8 +24,12 @@ from services.fluentd import Fluentd
 from warnet.tank import Tank
 from warnet.utils import parse_bitcoin_conf, gen_config_dir, bubble_exception_str, version_cmp_ge
 
-logger = logging.getLogger("warnet")
+
 FO_CONF_NAME = "fork_observer_config.toml"
+ASMAP_TXT_PATH = "asmap.txt"
+ASMAP_DAT_PATH = "asmap.dat"
+
+logger = logging.getLogger("warnet")
 logging.getLogger("docker.utils.config").setLevel(logging.WARNING)
 logging.getLogger("docker.auth").setLevel(logging.WARNING)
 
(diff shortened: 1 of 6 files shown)


willcl-ark:
@pinheadmz @josibake 

If y'all want ASmap, then we ready for review in here :)

src/warnet/warnet.py
```diff
@@ -38,7 +41,9 @@ def __init__(self, config_dir):
         self.graph: Optional[networkx.Graph] = None
         self.graph_name = "graph.graphml"
         self.tanks: List[Tank] = []
-
+        self.a_systems = set()
+        self.as_map_txt_path = config_dir / "asmap.txt"
+        self.as_map_dat_path = config_dir / "asmap.dat"
```
josibake: should probably get in the habit of not hardcoding files names. in the past, i think this has made it hard for us to pull things out / refactor

src/warnet/warnet.py
```diff
@@ -122,6 +127,43 @@ def tanks_from_graph(self):
                 )
             self.tanks.append(Tank.from_graph_node(node_id, self))
         logger.info(f"Imported {len(self.tanks)} tanks from graph")
+        self.generate_as_map()
+
+    def generate_as_map(self):
+        # Write AS mappings to file
+        with open(self.as_map_txt_path, "w") as f:
+            for tank in self.tanks:
+                f.write(f"{tank.ipv4}/32 AS{tank.autonomous_system}\n")
```
josibake, line 141: why "/32" here? will this /32 work regardless of where warnet is running?

src/warnet/warnet.py
```diff
@@ -122,6 +127,43 @@ def tanks_from_graph(self):
...
+        logger.info(f"AS map: Total bits: {len(ser)}")
+        with open(self.as_map_dat_path, "wb") as f:
+            f.write(bytes(buildmap.EncodeBytes(ser)))
+
+        # Reset sys.stdin to its original state
+        sys.stdin = sys.__stdin__
+
+        # load the file into each bitcoin node
+
+        # have each node read it on startup
```
josibake: not finished yet? or are these comments outdated?

josibake (commented):
left a few comments

src/warnet/warnet.py
willcl-ark: That's means "a single ip address"

src/warnet/warnet.py
willcl-ark: Oh, thought I took that out like I did in the remove volumes fixup

src/warnet/warnet.py
willcl-ark: Hmmm, the combo of this and the first comment makes me think my rebase went wrong. 

Will address after bathtime 🐤

willcl-ark: OK addressed in 08f4dc15bb71c6b4b3b646bb5d41bfd916f628ed

src/warnet/warnet.py
willcl-ark: fixed in 08f4dc15bb71c6b4b3b646bb5d41bfd916f628ed

willcl-ark:
Closing for #217

PR #147, opened 2023-10-17
tanks: remove mounted volumes and dns seeder

willcl-ark:
This saves mounting 2 (or 3 after ASmap changes) images per tank node, which won't scale nicely on k8s.

Also sort tank init code in classmethods, so that we can add more self.* variables during init (and not in two places each time)


willcl-ark:
Here we are @josibake, some more k8s-compatible tanks.

@pinheadmz the context here is that mounting multiple volumes per tank is not a best-practice on k8s, so we try to copy in all the required files for each tank.

Currently this leaves all the (tiny) files in config_dir, which I think is fine.

josibake:
> some more k8s-compatible tanks.

its actually necessary (not just preferred) to not using volumes for this stuff unless we want a separate process for creating and copying files over to `PersistentVolumes` in k8s.

Tested this with my `k8s-wip` branch and this works great. The only ones not working are forkobserver and fluentd, which are still using volumes, but seems like that could be a follow-up as those aren't strictly necessary for running scenarios in k8s.

src/warnet/tank.py
```diff
@@ -28,25 +32,29 @@
...
-        self.torrc_file = None
         self.netem = None
         self.rpc_port = 18443
         self.rpc_user = "warnet_user"
         self.rpc_password = "2themoon"
+        self.dockerfile_path = config_dir / "Dockerfile"
+        self.dns_path = config_dir / "check_dns.sh"
+        self.torrc_path = config_dir / "torrc"
+        self.warnet_entrypoint = config_dir / "warnet_entrypoint.sh"
+        self.docker_entrypoint = config_dir / "docker_entrypoint.sh"
```
josibake: feels like hardcoding warnet_entrypoint, docker_entrypoint, Dockerfile, etc as strings is gonna bite us in the ass. i see a few places where these are used in this commit, maybe we start using VARS, like with CONTAINER_PREFIX_BITCOIND, etc?

josibake (approved):
LGTM, left a comment regarding hardcoding string names; feels pretty brittle. Altho, this can probably be addressed later in a cleanup PR

src/warnet/tank.py
willcl-ark: Just seen this, and addressed in 0196682cb700a801b0f41a47537555669d84c9ef

PR #148, opened 2023-10-18
Use abstract base class for container operations

willcl-ark:


src/warnet/tank.py
```diff
@@ -116,30 +124,37 @@ def ipv4(self):
...
             )
         return self._exporter_name
 
     @property
     def container(self) -> Container:
+        # cast makes typecheckers happy
         if self._container is None:
             try:
-                self._container = docker.from_env().containers.get(self.container_name)
+                self._container = cast(Container, docker.from_env().containers.get(self.container_name))
```
pinheadmz: I was grepping for leftover `docker.from_env()` calls and I just wanna make sure that this line here is still abstractable

willcl-ark:
OK going to squash from 9057cef back

willcl-ark:
I didn't address the extra `services` in this one, but @josibake services are all inside `DockerInterface` so you shouldn't be bothered by them while working on getting tanks running on k8s (although without tor running some tests might not work).

I can add services to this one, or do in a followup (I'd prefer followup I think, but not strongly).

josibake:
LGTM, :shipit: 

> I can add services to this one, or do in a followup (I'd prefer followup I think, but not strongly).

followup :+1:

src/warnet/tank.py
josibake: nice catch, looks like this was fixed

PR #152, opened 2023-10-19
Move example & default graphs to src/graphs and document

### LINKED ISSUE #149: docs: make it clear that scenarios dont work without a network
We should reorder the docs or add a page between `install` and `warcli` or `scenarios` that explain how to start a network maybe with the example graphml or we could move `graph.md` earlier in the docs sequence

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/149

Moves our example.graphml to src/graphs/default.graphml. Also rearranges the docs so "gaphs" comes earlier to explain to users that nothing can happen without a network loaded from a graph first!


willcl-ark:
LGTM

PR #154, opened 2023-10-19
ci: factor out common action steps

pinheadmz:
Also switches ci from `docker-compose` to `docker compose` plugin


willcl-ark:
LGTM.

Don't worry about adding me as co-author :P

PR #156, opened 2023-10-20
abstract services

willcl-ark:


willcl-ark:
The service base classes now onyl really exist to specify which services are available.

Each service gets an `add_to_deployment()` method which should be implemented in the subclass for that containerisation/orchestration system.

willcl-ark:
Tests work on my machine; very cool docker!

PR #160, opened 2023-10-24
server: use sys.exit() to stop

pinheadmz:


willcl-ark:
Going to merge this for now as it helps CI

PR #163, opened 2023-10-24
WIP: more scaling fixes

willcl-ark:

PR #166, opened 2023-10-27
Clean up test operations

pinheadmz:
All the `test_base` commits from #163. I'm hoping I can see one PR pass CI tonight and not screw up my terminal output with escape codes. Let's narrow down what's going on over there.

### DIFF
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -47,7 +47,7 @@ class TestBase:
             print(f"Error stopping server: {e}")
             print("Attempting to cleanup docker network")
             wn = Warnet.from_network(self.network_name)
-            wn.docker_compose_down()
+            wn.warnet_down()
 
         self.server.terminate()
         self.server = None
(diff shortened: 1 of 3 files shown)


test/test_base.py
```diff
@@ -11,7 +12,10 @@ class TestBase:
     def __init__(self):
         # Warnet server stdout gets logged here
         self.tmpdir = Path(mkdtemp(prefix="warnet_test_"))
-        self.logfilepath = self.tmpdir / "warnet.log"
+
+        os.environ["XDG_STATE_HOME"] = f"{self.tmpdir}"
```
willcl-ark, line 16: Oh this is pretty lush! I was worried about setting this here (rather than reading it), but setting it is only local to the python process and its children, so this does get propagated to _server.py_ and used, but also doesn't actually set it in your shell. Neat!

willcl-ark (approved):
noice!

PR #171, opened 2023-10-27
gunicorn ❤️

willcl-ark:
Want to stop those pesky `warnet` instances from hanging around; use gunicorn! (use 1 worker thread only to keep "state")

Also, exponential backoff is occasionally locking us to sleep, fix that too.


willcl-ark:
Closing as @pinheadmz will never let me have my overly-complicated WSGI server 😢

PR #172, opened 2023-10-28
Use docker images

willcl-ark:
* Use remote docker images from registry for pre-built versions, but keep local builds for branches.
* Pre-built images and branches use same dockerfile.
* Images are multi-arch (amd64, arm64 and arhf for now, can be expanded manually if needed)
* Update supported tags
* Update default graphs to 25.1


willcl-ark:
Just testing the top 4 commits here, I'll drop them from this PR and make them dependent on it if they work ok

scripts/build_images.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "DOCKER_REGISTRY=$DOCKER_REGISTRY"
+echo "BITCOIN_URL=$BITCOIN_URL"
+
+# Map internal architectures to the format used in the URLs
+declare -A ARCH_MAP=(
+    ["amd64"]="x86_64-linux-gnu"
+    ["arm64"]="aarch64-linux-gnu"
+    ["armhf"]="arm-linux-gnueabihf"
+    # ["amd64-darwin"]="x86_64-apple-darwin"
+    # ["arm64-darwin"]="arm64-apple-darwin"
```
josibake, line 18: are these not working?

scripts/build_images.sh
```diff
@@ -0,0 +1,71 @@
...
+)
+
+# Tags and their supported architectures
+declare -A VERSION_ARCH_MAP=(
```
josibake, line 22: do we need this separate map? I can't think of a reason that a specific version wouldn't support all architectures? why not have a single variable for `ARCH`, regardless of version?

src/interfaces/docker_interface.py
```diff
@@ -229,58 +220,28 @@ def _write_docker_compose(self, warnet):
...
 
     def copy_configs(self, tank):
-        self.copy_torrc(tank)
-        self.copy_entrypoints(tank)
-        self.copy_dockerfile(tank)
+        import shutil
+        shutil.copyfile(TEMPLATES / DOCKERFILE_NAME, tank.config_dir / DOCKERFILE_NAME)
+        shutil.copyfile(TEMPLATES / TORRC_NAME, tank.config_dir / TORRC_NAME)
+        shutil.copyfile(TEMPLATES / ENTRYPOINT_NAME, tank.config_dir / ENTRYPOINT_NAME)
+        set_execute_permission(tank.config_dir / ENTRYPOINT_NAME)
```
josibake, line 237: this assumes the template directory and the tank.config_dir are on the same machine. while this has been true so far with docker, it's not necessarily always true and in the case of kubernetes it's definitely not true. probably fine for now, but just wanted to point it out and mainly note we will have to change this for k8s support.

src/interfaces/docker_interface.py
```diff
@@ -157,8 +157,9 @@ def get_messages(self, a_name: str, b_ipv4: str, bitcoin_network: str = "regtest
         return messages
 
 
-    def logs_grep(self, pattern: str, container_name: str):
+    def logs_grep(self, pattern: str, network: str):
         compiled_pattern = re.compile(pattern)
+        container_name = self.get_container(f"{network}_fluentd").name
```
josibake: assumes we are always using fluentd for the logger. also, would be really nice if anywhere we see a hardcoded string (`_fluentd`, in this case). in my WIP k8s branch I've had to hunt down all these places and modify these strings because k8s doesn't give the same names as docker compose.

josibake (commented):
looks good! would love to get the container image stuff merged and working with the docker hub registry, is it possible to split off the unrelated commits into their own PR?

willcl-ark:
Yeah I will do. Once I figure out which torrc setting I can bake in!

Hopefully first thing this am

josibake:
Seems like everything after https://github.com/bitcoin-dev-project/warnet/pull/172/commits/8ffe4599a89ff6fbb0d656aa105463acb8c26470 is unrelated. I also see you have a few other PRs building off https://github.com/bitcoin-dev-project/warnet/pull/172/commits/8ffe4599a89ff6fbb0d656aa105463acb8c26470, so seems like we should split this out and merge it?

scripts/build_images.sh
willcl-ark: Nope!

scripts/build_images.sh
willcl-ark: They all support different arches (going back in time), see for example this version:

https://bitcoincore.org/bin/bitcoin-core-0.20.2/

vs this version

https://bitcoincore.org/bin/bitcoin-core-25.1/

handy!

src/interfaces/docker_interface.py
willcl-ark: I know! :'(

My plan was to get this working, which I think I have now! Then I also plan to build tor Directory Authority and Tor Relay images, and we can just pull those and get rid of all these file copies... I think it's for a followup though, just based on how I've build more on top in #173

src/interfaces/docker_interface.py
willcl-ark: This is in the docker interface though? WYD in here on k8s?

scripts/build_images.sh
josibake: ah nice! TIL

willcl-ark: So it's available, but currently not configured.

So if someone really wants e.g. 0.20.2 they can add it to the map with available architectures, then fix all tghe broken docker building steps on their own :D

src/interfaces/docker_interface.py
josibake: sounds good. just reiterating my comment from the chat, I really think we would benefit from dropping Tor for now and adding it in later once we've hammered down scaling on docker / k8s and have done some code cleanup. seems like Tor is sucking up a lot of brain power and the consistent blocker is always "trying to figure out how to get bla bla with tor working"

willcl-ark: I think you can just not run tor services in your k8 interface? It's all set up in the docker interface and not strict requirement of a warnet. Although by default the bitcoindevproject images will run tor daemon... I should make that configurable before pushing new images.

Then we can turn it on in docker land, and k8s and not set it until they're ready?

src/interfaces/docker_interface.py
josibake: on the k8s interface, the fluentd service has a slightly different name so i just updated it inline. but my comment was more a general: "hardcoded strings are codesmell, we should be abstracting this into a config, and not specifying them in the interfaces"

that way, if something changes (e.g. we change the logging on the docker interface, or the naming changes on the k8s interface, etc), we have a single config file where were update the name once and then the rest of the code "just works"

scripts/build_images.sh
```diff
@@ -66,6 +63,6 @@ for VERSION in "${!VERSION_ARCH_MAP[@]}"; do
     done
 
     # Create the manifest list for each version under the same repository
-    MANIFEST_TAG="$DOCKER_REGISTRY/$IMAGE_NAME:$VERSION"
```
willcl-ark: Don't particularly understand why this worked on GitLab but not on DockerHub, but w/e

src/interfaces/docker_interface.py
josibake: I'm not worried about it for k8s, I'm worried about the dev time that its taking to get tor working on docker, which seems to be slowing everything down.

src/interfaces/docker_interface.py
willcl-ark: Hmmm, I guess I don't understand why your in this file still? But I agree re. hardcoded strings.

My master plan is actually to rip out fluentd, see middle commits here https://github.com/bitcoin-dev-project/warnet/pull/173/commits

It seems to use a lot of resources and so doesn't scale. Docker's stdout logging can work for us in this sense, so it might be best plan to just use whatever is native/easiest in k8s too? As it's an interface method you can do whatever you like with it on your k8s side.

I just changed how I grep all logs in the followup commit: [`55ff125` (#173)](https://github.com/bitcoin-dev-project/warnet/pull/173/commits/55ff125810c875a20a6186cd753d5581adebd62e)

I think the fact the log parser is in Ruby is the culprit, I often saw that maxing out one of my precious cores...

josibake: I was trying to get fluentd working on k8s, and IIRC ended up coping some stuff over from the docker interface. But agree that fluentd isn't great (i never did end up getting it working on k8s), and we will likely end up using something else for k8s. The comment was less about fluentd and more about pointing out places where we use string names in specific functions. Not critical to fix at this point, but more a "wouldn't it be nice" as we go forward.

willcl-ark:
I'll need some Tor commits to make this pass with the current images on bitcoindevproject 🥲

willcl-ark:
Let's see if these help. I think without these Tor changes -- specifically adding 10 dedicated relays using docker's `--scale`, that we won't build any tor circuits, [so onion is expected to fail](https://github.com/bitcoin-dev-project/warnet/actions/runs/6691714452/job/18179460449?pr=172).

With these 3 additional commits (which yes, also include some scaling stuff, sorry), then onion (and everything else) should pass.

willcl-ark:
So it fails here, but works on my_machine.gif yay docker!

Will re-run it once, but if we are seriously considering temporarily ripping out Tor, perhaps this is GTG?

willcl-ark:
OK so 8eab5a5a8c01536ca4365c1b933afd2f3a6ce671 does indeed pass as expected.

Going to update the docker images (removing tor) and push a final tor removal commit on top here.

If that passes too, we can choose to proceed without Tor in the docker images (and project), although I would really like the commits to remain in the history so we can undo them more easily in the future...

@josibake @pinheadmz

willcl-ark:
Now with Tor removed.

25.1 images are build and pushed, rest are on their way.

WDYT @josibake @pinheadmz ?

josibake:
LGTM, merge after CI passes?

willcl-ark:
> LGTM, merge after CI passes?

Well, that and if MZ confirms that it's not wanted for the first use-case!

.github/workflows/test.yml
```diff
@@ -19,12 +19,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions
       - run: ./test/scenarios_test.py
-  onion:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: ./.github/actions
-      - run: ./test/v25_net_test.py
```
pinheadmz, line 27: There are a few other checks in this test but ok

src/interfaces/docker_interface.py
```diff
@@ -328,7 +285,7 @@ def add_services(self, tank, services):
                     "test": ["CMD", "pidof", "bitcoind"],
                     "interval": "5s",            # Check every 5 seconds
                     "timeout": "1s",             # Give the check 1 second to complete
-                    "start_period": "5s",       # Start checking after 2 seconds
+                    "start_period": "5s",        # Start checking after 2 seconds
```
pinheadmz:
```suggestion
                    "start_period": "5s",        # Start checking after 5 seconds
```

src/interfaces/docker_interface.py
```diff
@@ -295,22 +255,19 @@ def add_services(self, tank, services):
...
-        # Add the bitcoind service
+            image = f"{DOCKER_REGISTRY}:{tank.version}"
+            services[tank.container_name]['image'] = image
+        # Add common bitcoind service details
         services[tank.container_name].update(
             {
                 "container_name": tank.container_name,
-                "build": build,
+                "environment": {
+                    "WARNET": "1",
```
pinheadmz: what is the `WARNET` arg for?

src/templates/bitcoin.conf
```diff
@@ -11,16 +9,11 @@ debugexclude=libevent
...
 rpcallowip=0.0.0.0/0
 rpcbind=0.0.0.0
-rest=1 # for forkmonitor
+rest=1
 
-# Wallet
 fallbackfee=0.00001000
 
-# P2P
 onion=127.0.0.1:9050
```
pinheadmz: mine as well remove this too to reduce log spam from:

```


2023-10-30T13:57:06.621686Z [tor] Error connecting to Tor control socket
2023-10-30T13:57:06.621789Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
2023-10-30T13:58:04.334333Z [tor] Error connecting to Tor control socket
2023-10-30T13:58:04.334405Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
2023-10-30T13:59:30.907407Z [tor] Error connecting to Tor control socket
2023-10-30T13:59:30.907456Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
2023-10-30T14:01:40.747036Z [tor] Error connecting to Tor control socket
2023-10-30T14:01:40.747156Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
2023-10-30T14:04:55.450443Z [tor] Error connecting to Tor control socket
2023-10-30T14:04:55.450495Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
2023-10-30T14:09:47.383646Z [tor] Error connecting to Tor control socket
2023-10-30T14:09:47.383896Z [tor] Not connected to Tor control port 127.0.0.1:9051, trying to reconnect
```

pinheadmz (approved):
okie doke!

scripts/build_images.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
pinheadmz: Needed this on macOS to load the newest version of bash, installed by homebrew. Otherwise `declare -A` is an error

```suggestion
#!/usr/bin/env bash
```

src/interfaces/docker_interface.py
willcl-ark: done

src/interfaces/docker_interface.py
willcl-ark: This is used in entrypoint.sh to optionally copy over our custom warnet torrc

https://github.com/willcl-ark/warnet/blob/f6cb1fc11e6e7b7fcf1c1e66c259abd3e6ca1b6b/src/templates/entrypoint.sh#L12-L14

This means that in the end, passing env var `TOR=1` to the tank will enable tor (regular) and `TOR=1 WARNET=1` will use warnet tor (I'm not done with it yet!)

willcl-ark: It doesn't matter, but it's incorrectly set here, as we're not using TOR=1

willcl-ark: removed!

scripts/build_images.sh
willcl-ark: done

src/templates/bitcoin.conf
willcl-ark: Also done

willcl-ark:
OK a final force-push with @pinheadmz nits addressed

https://github.com/bitcoin-dev-project/warnet/compare/f6cb1fc11e6e7b7fcf1c1e66c259abd3e6ca1b6b..ad3b5eb5de27a1c269118b7fc897edb9a32ecbed

pinheadmz:
cool thanks merge at will, Will

PR #173, opened 2023-10-29
Tame tor, add CAdvisor, Prometheus, Grafana, default dashboard, image building script

willcl-ark:
Built on top of #172

* New commits from:

> ABC: update grep_logs ABC to be more generic

### DIFF
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -1,0 +1,32 @@
+# Monitoring
+
+## Monitoring container resource usage
+
+When run in docker, a few additional containers are started up:
+
+* CAdvisor (container Monitoring)
+* Prometheus (log scraper)
+* Grafana (graphing/dashboard tool)
+
+## CAdvisor
+
+CAdvisor needs no additional setup, and can be accessed from the docker host at
+localhost:8080
+
+## Prometheus
+
+Prometheus should also not need any additional setup, and can be accessed from
+the docker host at localhost:9090
+
+## Grafana
+
+Grafana is provisioned with a single default dashboard, but alternative
+dashboards can be added or created.
+
+Grafana can be accessed on the docker host from localhost:3000 using username
+`admin` and password `admin` by default.
+
+The default dashboard is called "Docker Container & Host Metrics" and can be
+accessed via the "dashboards" tab, or from the bottom right of the home screen.
+
+Additional dashboards and datasources may be added in the future.
--- a/src/services/grafana-provisioning/dashboards/10619_rev2.json
+++ b/src/services/grafana-provisioning/dashboards/10619_rev2.json
@@ -1,0 +1,1260 @@
+{
+  "annotations": {
+    "list": [
+      {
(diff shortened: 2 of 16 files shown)


willcl-ark:
Now that I think about it, the three fluentd commits could be extracted into their own pull if desired:

d462886 fixup logs_grep in the docker-interface to not rely on fluentd
e4065f4 disable fluentd logging driver
526355e ABC: update grep_logs ABC to be more generic

willcl-ark:
Rebased on #172

pinheadmz:
running 250 nodes easily on this branch. would be nice to have some docs in data collection page about the tools. For the web-ui stuff we should expose those ports so for example I can see them runing on my VPS from a remote web browser

willcl-ark:
Strongly agree, will add some docs commits. We can also do with removing Tor DA from our homepage from the previous pull :'(

willcl-ark:
Cancelled tests until I add docs commit. The force push was just again on top of main

pinheadmz:
Screen Shot 2023-10-30 at 3 04 49 PM
 hot!

docs/monitoring.md
```diff
@@ -0,0 +1,32 @@
...
+When run in docker, a few additional containers are started up:
+
+* CAdvisor (container Monitoring)
+* Prometheus (log scraper)
+* Grafana (graphing/dashboard tool)
+
+## CAdvisor
+
+CAdvisor needs no additional setup, and can be accessed from the docker host at
+localhost:3000
```
pinheadmz: actually `8080`

src/services/grafana-provisioning/dashboards/10619_rev2.json
```diff
@@ -0,0 +1,1278 @@
...
+            "warnet_tank_000001",
+            "warnet_tank_000002",
+            "warnet_tank_000003",
+            "warnet_tank_000010",
+            "warnet_tank_000009",
+            "warnet_tank_000011",
+            "warnet_tank_000000",
+            "warnet_tank_000005",
+            "warnet_tank_000004",
+            "warnet_tank_000007"
```
pinheadmz: hard-coded tank names? This might work on network=warnet up to 12 tanks but what about other network names / number of tanks? Running default network (which should match this list) I don't see anything in the dashboard:
Screen Shot 2023-11-01 at 5 37 02 PM

willcl-ark: Damn, I didn't know that would be in there. It's supposed to use a formula!?

willcl-ark: Fixed now I think, and added an "all" (which is the default on).

FWIW you do need to wait a minute or so for some graphs to move off "no data"...

pinheadmz:
ACK, ACK, ACK, ACK, ACK and ACK. Although I couldn't verify that `build_script` is totally working without access to the public image repo. GOOD WORK WILL.

PR #175, opened 2023-10-31
fork-observer: reduce query interval

willcl-ark:
For S-C-A-L-I-N-G

Reduce query interval to every 60 seconds by default. Helps on larger networks.

### DIFF
--- a/src/templates/fork_observer_config.toml
+++ b/src/templates/fork_observer_config.toml
@@ -7,7 +7,7 @@ database_path = "db"
 www_path = "./www"
 
 # Interval for checking for new blocks
-query_interval = 10
+query_interval = 60
 
 # Webserver listen address
 address = "0.0.0.0:2323"


0xB10C:
lgtm

I might want to add some randomness to the query interval for each node in fork-observer so you don't end up firing 200 RPC calls exactly ever 60s.

0xB10C:
Also, keep in mind that fork-observer currently has a limit of 256 nodes per network. It's trivial to increase it but I haven't gotten to it yet. https://github.com/0xB10C/fork-observer/issues/22

willcl-ark:
Oh right, I think I will also add a commit here to disable if the graph is >256 nodes then

0xB10C:
I've fixed it in fork-observer: https://github.com/0xB10C/fork-observer/pull/24

The new docker image should allow Warnet to theoretically attach more than 256 nodes. Not sure if that's practical though.

pinheadmz:
Do we still need this now that fork observer was updated? There's also https://github.com/bitcoin-dev-project/warnet/pull/177

willcl-ark:
Yeah we want this. Updated FO lets us attach more nodes (256 limit), this queries the nodes less frequently. You know how you saw FO was using the most CPU/bandwidth on the graphs earlier... ;)

#177 is different

willcl-ark:
Would be even nicer (for us) if FO would query on a chill scheduler, e.g. If you configure 100 nodes and query every 60 seconds it could query one node every 0.6 seconds to keep resource usage as flat as possible. Vs currently where every n seconds it will query all nodes at the same time (or as fast as it can).

But I haven't thought about side effects of that for wider FO usage and that's a PR for another repo in any case!

pinheadmz:
can remove this commit though? https://github.com/bitcoin-dev-project/warnet/pull/175/commits/bf2e05873f690022059a05df3d5e60378da874c6

willcl-ark:
ohhhhh, sure

willcl-ark:
rebased and dropped commit

0xB10C:
> Would be even nicer (for us) if FO would query on a chill scheduler, e.g. If you configure 100 nodes and query every 60 seconds it could query one node every 0.6 seconds to keep resource usage as flat as possible. Vs currently where every n seconds it will query all nodes at the same time (or as fast as it can).

Agree, I made an issue for this https://github.com/0xB10C/fork-observer/issues/23 the other day.

PR #176, opened 2023-10-31
scripts: add generic docker multi arch build script

willcl-ark:
also rename build-bitcoin script


pinheadmz:
whats this for / how used?

willcl-ark:
Oh, sorry.

This is a more generic version of the Bitcoin Core multi-arch image builder, which you use with a passed-in dockerfile.

This could be used to generate images for any service. IIUC k8s basically wants an image for everything, so this would be useful. I was using it when testing out building Tor DA and Tor Relays as an image (while I was trying out having 6 dedicated Tor relays to try and reduce bandwidth usage). By using images here I was able to spin up 5 relays by just using docker compose's `--scale` option, which was nice.

I reckon it would be useful in the future though, even if unused now?

pinheadmz:
ok sure!

PR #177, opened 2023-10-31
fork-observer: extract from warnet

willcl-ark:
This is best implemented by the container implementation

@josibake Going to propose this for now. We should remove FO from `Warnet` in any case, this just lets you handle it in your own way more cleanly


josibake:
Concept ACK

If I'm understanding the description, this means each backend needs to decided how/if to implement fork observer?

willcl-ark:
Yeah that's right. 

I think we should really get it back, as it's a great tool, but maybe low prio until after we get some tests running stably on k8s

PR #178, opened 2023-10-31
Add k8s backend

josibake:
This *mostly* works, but is meant more as a PoC and really the final state. To test this PR:

1. setup a k8s cluster - I recommend digital ocean, as you can leave the cluster running and add/remove node pools as needed. Ive done all of my testing of the default graph with a basic node pool, where the cost is something like 10~30 bucks a month. This way, you can easily add a much larger node pool to test larger simulations, and then immediately delete when done. You could also use minikube, but eh. If not using minikube, make sure to install kubectl and connect it to your cluster
2. Deploy the resources in `src/templates/`, this sets up the correct permissions on the cluster (`rbac-config.yaml`) and deploys the warnet RPC server as a service + statefulset. This can be done with `kubectl apply -f '*.yaml'` from inside the src/templates directory.
3. Once the RPC server comes up, run `kubectl port-forward svc/rpc 9276:9276` - this allows you to communicate with the rpc server using `warcli`
4. run `warcli network start --force` ! 

I haven't fully implemented the interface, so you really should manually delete the pods in between runs using something like `kubectl delete pods --all`. You can also tail the logs on the rpc server using `kubectl logs rpc-0 -f`

### DIFF
--- a/Dockerfile_rpc
+++ b/Dockerfile_rpc
@@ -8,7 +8,7 @@ WORKDIR /root/warnet
 COPY . /root/warnet
 RUN pip install .
 
-# Make port 80 available to the world outside this container
+# Make port 9276 available to the world outside this container
 # Change the port if your server is running on a different port
 EXPOSE 9276
 
--- a/Dockerfile_rpc
+++ b/Dockerfile_rpc
@@ -1,5 +1,5 @@
 # Use an official Python runtime as the base image
-FROM python:3.9-slim
+FROM python:3.11-slim
 
 # Set the working directory in the container
 WORKDIR /root/warnet
--- a/Dockerfile_rpc
+++ b/Dockerfile_rpc
@@ -1,0 +1,16 @@
+# Use an official Python runtime as the base image
+FROM python:3.9-slim
+
+# Set the working directory in the container
+WORKDIR /root/warnet
+
+# Copy the current directory contents into the container at /app
+COPY . /root/warnet
+RUN pip install .
+
+# Make port 80 available to the world outside this container
+# Change the port if your server is running on a different port
+EXPOSE 9276
+
+# Run server.py when the container launches
+CMD ["warnet", "k8s"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
     "jsonschema",
     "jsonrpcserver==5.0.3",
     "jsonrpcclient==4.0.0",
+    "kubernetes",
     "networkx==3.1",
     "numpy==1.26.0",
     "rich==13.5.2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "jsonschema",
     "jsonrpcserver==5.0.3",
     "jsonrpcclient==4.0.0",
-    "kubernetes",
+    "kubernetes~=28.1.0",
     "networkx==3.1",
     "numpy==1.26.0",
     "rich==13.5.2",
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,4 +12,3 @@ python-decouple
 uuid
 pyjwt[crypto]
 email-validator
-kubernetes
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ python-decouple
 uuid
 pyjwt[crypto]
 email-validator
+kubernetes
--- a/src/backends/__init__.py
+++ b/src/backends/__init__.py
@@ -1,0 +1,3 @@
+from .backend_interface import BackendInterface
+from .compose_backend import ComposeBackend
+from .kubernetes_backend import KubernetesBackend
--- a/src/backends/__init__.py
+++ b/src/backends/__init__.py
@@ -1,3 +1,3 @@
-from .backend_interface import BackendInterface
+from .backend_interface import BackendInterface, ServiceType
 from .compose_backend import ComposeBackend
 from .kubernetes_backend import KubernetesBackend
--- a/src/backends/compose_backend.py
+++ b/src/backends/compose_backend.py
@@ -18,7 +18,7 @@ from services.grafana import Grafana
 from services.prometheus import Prometheus
 from templates import TEMPLATES
 from warnet.tank import Tank
-from warnet.status import Status
+from warnet.status import RunningStatus
 from warnet.lnnode import LNNode
 from warnet.utils import bubble_exception_str, parse_raw_messages, default_bitcoin_conf_args, set_execute_permission
 
@@ -119,20 +119,20 @@ class ComposeBackend(BackendInterface):
             return None
         return cast(Container, self.client.containers.get(container_name))
 
-    def get_status(self, tank_index: int, service: ServiceType) -> Status:
+    def get_status(self, tank_index: int, service: ServiceType) -> RunningStatus:
         container = self.get_container(tank_index, service)
         if container is None:
-            return Status.STOPPED
+            return RunningStatus.STOPPED
         match container.status:
             case 'running':
-                return Status.RUNNING
+                return RunningStatus.RUNNING
             case 'exited' | 'dead':
                 if container.attrs['State']['ExitCode'] == 0:
-                    return Status.STOPPED
+                    return RunningStatus.STOPPED
                 else:
-                    return Status.FAILED
+                    return RunningStatus.FAILED
             case _:
-                return Status.PENDING
+                return RunningStatus.PENDING
 
     def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "root") -> str:
         c = self.get_container(tank_index, service)
--- a/src/backends/compose_backend.py
+++ b/src/backends/compose_backend.py
@@ -10,15 +10,15 @@ from typing import cast, List, Tuple
 import docker
 from docker.models.containers import Container
 
-from .interfaces import ContainerInterface
-from warnet.utils import bubble_exception_str, parse_raw_messages
+from .backend_interface import BackendInterface, ServiceType
 from services import SERVICES
 from services.cadvisor import CAdvisor
 from services.fork_observer import ForkObserver
 from services.grafana import Grafana
 from services.prometheus import Prometheus
 from templates import TEMPLATES
-from warnet.tank import Tank, CONTAINER_PREFIX_BITCOIND
+from warnet.tank import Tank
+from warnet.status import Status
 from warnet.lnnode import LNNode
 from warnet.utils import bubble_exception_str, parse_raw_messages, default_bitcoin_conf_args, set_execute_permission
 
@@ -29,16 +29,20 @@ TORRC_NAME = "torrc"
 ENTRYPOINT_NAME = "entrypoint.sh"
 DOCKER_REGISTRY = "bitcoindevproject/bitcoin-core"
 GRAFANA_PROVISIONING = "grafana-provisioning"
+CONTAINER_PREFIX_BITCOIND = "tank-bitcoin"
+CONTAINER_PREFIX_LN = "tank-ln"
 
 logger = logging.getLogger("docker-interface")
 logging.getLogger("docker.utils.config").setLevel(logging.WARNING)
 logging.getLogger("docker.auth").setLevel(logging.WARNING)
 
 
-class DockerInterface(ContainerInterface):
-    def __init__(self, config_dir: Path) -> None:
+class ComposeBackend(BackendInterface):
+    def __init__(self, config_dir: Path, network_name: str) -> None:
         super().__init__(config_dir)
+        self.network_name = network_name
         self.client = docker.DockerClient = docker.from_env()
+
 
     @bubble_exception_str
     def build(self) -> bool:
@@ -100,19 +104,47 @@ class DockerInterface(ContainerInterface):
                 f"An error occurred while executing `{' '.join(command)}` in {self.config_dir}: {e}"
             )
 
-    def get_container(self, container_name: str) -> Container:
+    def get_container_name(self, tank_index: int, service: ServiceType) -> str:
+        match service:
+            case ServiceType.BITCOIN:
+                return f"{self.network_name}-{CONTAINER_PREFIX_BITCOIND}-{tank_index:06}"
+            case ServiceType.LIGHTNING:
+                return f"{self.network_name}-{CONTAINER_PREFIX_LN}-{tank_index:06}"
+            case _:
+                raise Exception("Unsupported service type")
+
+    def get_container(self, tank_index: int, service: ServiceType) -> Container:
+        container_name = self.get_container_name(tank_index, service)
+        if len(self.client.containers.list(filters={"name": container_name})) == 0:
+            return None
         return cast(Container, self.client.containers.get(container_name))
 
-    def exec_run(self, container_name: str, cmd: str, user: str = "root"):
-        c = self.get_container(container_name)
+    def get_status(self, tank_index: int, service: ServiceType) -> Status:
+        container = self.get_container(tank_index, service)
+        if container is None:
+            return Status.STOPPED
+        match container.status:
+            case 'running':
+                return Status.RUNNING
+            case 'exited' | 'dead':
+                if container.attrs['State']['ExitCode'] == 0:
+                    return Status.STOPPED
+                else:
+                    return Status.FAILED
+            case _:
+                return Status.PENDING
+
+    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "root") -> str:
+        c = self.get_container(tank_index, service)
         result = c.exec_run(cmd=cmd, user=user)
         if result.exit_code != 0:
             raise Exception(
-                f"Command failed with exit code {result.exit_code}: {result.output.decode('utf-8')}"
+                f"Command failed with exit code {result.exit_code}: {result.output.decode('utf-8')} {cmd}"
             )
         return result.output.decode("utf-8")
 
-    def get_bitcoin_debug_log(self, container_name: str):
+    def get_bitcoin_debug_log(self, tank_index: int):
+        container_name = self.get_container_name(tank_index, ServiceType.BITCOIN)
         now = datetime.utcnow()
 
         logs = self.client.api.logs(
@@ -129,10 +161,10 @@ class DockerInterface(ContainerInterface):
             cmd = f"bitcoin-cli -regtest -rpcuser={tank.rpc_user} -rpcport={tank.rpc_port} -rpcpassword={tank.rpc_password} {method} {' '.join(map(str, params))}"
         else:
             cmd = f"bitcoin-cli -regtest -rpcuser={tank.rpc_user} -rpcport={tank.rpc_port} -rpcpassword={tank.rpc_password} {method}"
-        return self.exec_run(tank.container_name, cmd, user="bitcoin")
+        return self.exec_run(tank.index, ServiceType.BITCOIN, cmd, user="bitcoin")
 
-    def get_file(self, container_name, file_path):
-        container = self.get_container(container_name)
+    def get_file(self, tank_index: int, service: ServiceType, file_path: str):
+        container = self.get_container(tank_index, service)
         data, stat = container.get_archive(file_path)
         out = b""
         for chunk in data:
@@ -143,21 +175,22 @@ class DockerInterface(ContainerInterface):
         out = out[: stat["size"]]
         return out
 
-    def get_messages(self, a_name: str, b_ipv4: str, bitcoin_network: str = "regtest"):
+    def get_messages(self, tank_index: int, b_ipv4: str, bitcoin_network: str = "regtest"):
         # start with the IP of the peer
         # find the corresponding message capture folder
         # (which may include the internal port if connection is inbound)
         subdir = (
             "/" if bitcoin_network == "main" else f"{bitcoin_network}/"
         )
-        dirs = self.exec_run(a_name, f"ls /home/bitcoin/.bitcoin/{subdir}message_capture")
+        dirs = self.exec_run(tank_index, ServiceType.BITCOIN, f"ls /home/bitcoin/.bitcoin/{subdir}message_capture")
         dirs = dirs.splitlines()
         messages = []
         for dir_name in dirs:
             if b_ipv4 in dir_name:
                 for file, outbound in [["msgs_recv.dat", False], ["msgs_sent.dat", True]]:
                     blob = self.get_file(
-                        a_name,
+                        tank_index,
+                        ServiceType.BITCOIN,
                         f"/home/bitcoin/.bitcoin/{subdir}message_capture/{dir_name}/{file}")
                     json = parse_raw_messages(blob, outbound)
                     messages = messages + json
@@ -288,9 +321,10 @@ class DockerInterface(ContainerInterface):
         shutil.copyfile(TEMPLATES / ENTRYPOINT_NAME, tank.config_dir / ENTRYPOINT_NAME)
         set_execute_permission(tank.config_dir / ENTRYPOINT_NAME)
 
-    def add_services(self, tank, services):
+    def add_services(self, tank: Tank, services):
         assert tank.index is not None
-        services[tank.container_name] = {}
+        container_name = self.get_container_name(tank.index, ServiceType.BITCOIN)
+        services[container_name] = {}
         logger.debug(f"{tank.version=}")
 
         # Setup bitcoind, either release binary or build from source
@@ -306,15 +340,15 @@ class DockerInterface(ContainerInterface):
                     "BUILD_ARGS": f"{tank.DEFAULT_BUILD_ARGS + tank.extra_build_args}",
                 },
             }
-            services[tank.container_name]['build'] = build
+            services[container_name]['build'] = build
             self.copy_configs(tank)
         else:
             image = f"{DOCKER_REGISTRY}:{tank.version}"
-            services[tank.container_name]['image'] = image
+            services[container_name]['image'] = image
         # Add common bitcoind service details
-        services[tank.container_name].update(
+        services[container_name].update(
             {
-                "container_name": tank.container_name,
+                "container_name": container_name,
                 "environment": {
                     "BITCOIN_ARGS": self.default_config_args(tank)
                 },
@@ -355,6 +389,8 @@ class DockerInterface(ContainerInterface):
         # }
 
     def add_lnd_service(self, tank, services):
+        ln_container_name = self.get_container_name(tank.index, ServiceType.LIGHTNING)
+        bitcoin_container_name = self.get_container_name(tank.index, ServiceType.BITCOIN)
         # These args are appended to the Dockerfile `ENTRYPOINT ["lnd"]`
         args = [
             "--noseedbackup",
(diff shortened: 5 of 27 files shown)


m3dwards:
K8s Interface is not completely implemented but this should be ready for early review / merge

src/interfaces/kubernetes_interface.py
```diff
@@ -14,7 +14,7 @@
 from warnet.tank import Tank, CONTAINER_PREFIX_BITCOIND
 from warnet.utils import parse_raw_messages, default_bitcoin_conf_args
 
-DOCKER_REGISTRY = "bitcoindevproject/warnet-bitcoin-core"
+DOCKER_REGISTRY = "bitcoindevproject/k8s-bitcoin-core"
```
willcl-ark: I guess we can delete https://hub.docker.com/repository/docker/bitcoindevproject/warnet-bitcoin-core after this is merged. So we don't end up with too many zombie repos

Dockerfile_rpc
```diff
@@ -0,0 +1,16 @@
...
+FROM python:3.9-slim
+
+# Set the working directory in the container
+WORKDIR /root/warnet
+
+# Copy the current directory contents into the container at /app
+COPY . /root/warnet
+RUN pip install .
+
+# Make port 80 available to the world outside this container
```
willcl-ark:
```suggestion
# Make port 9276 available to the world outside this container
```

requirements.txt
```diff
@@ -12,3 +12,4 @@ python-decouple
 uuid
 pyjwt[crypto]
 email-validator
+kubernetes
```
willcl-ark:
```suggestion

```

IIUC requirements.txt is (unconventionally) not matching `pyproject.toml`, and being used to store the frontend deps, so we don't need `kubernetes` in here

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "jsonschema",
     "jsonrpcserver==5.0.3",
     "jsonrpcclient==4.0.0",
+    "kubernetes",
```
willcl-ark: Think we should pin to current version

willcl-ark: We could also add docker/k8s as optional dependencies in their own sections:

```
[project.optional-dependencies]
docker = [
    "docker==6.1.3
]
kubernetes = [
    "kubernetes==28.1.0
]
```

Which could allow selective install like:

```sh
pip install -e .[docker]
#or 
pip install -e .[kubernetes]
```

src/interfaces/kubernetes_interface.py
```diff
@@ -0,0 +1,263 @@
...
+                    json = parse_raw_messages(blob, outbound)
+                    messages = messages + json
+
+        messages.sort(key=lambda x: x["time"])
+        return messages
+
+    def logs_grep(self, pattern: str, network: str):
+        compiled_pattern = re.compile(pattern)
+
+        # Fetch the logs from the pod
```
willcl-ark: Should we add a TODO that this should iterate search over all tanks?

src/interfaces/docker_interface.py
```diff
@@ -383,7 +383,7 @@ def warnet_from_deployment(self, warnet):
                 warnet.tanks.append(tank)
 
     def tank_from_deployment(self, service, warnet):
-        rex = fr"{warnet.network_name}_{CONTAINER_PREFIX_BITCOIND}_([0-9]{{6}})"
+        rex = fr"{warnet.network_name}-{CONTAINER_PREFIX_BITCOIND}-([0-9]{{6}})"
```
willcl-ark, line 462: I saw it being discussed, but not totally clear why this needs to be changed in this PR? It should be totally unrelated to k8s, right (and it shouldn't matter if docker and k8's name diffferently)?

src/interfaces/kubernetes_interface.py
```diff
@@ -0,0 +1,263 @@
...
+        for log_entry in log_stream:
+            log_entry_str = log_entry.decode('utf-8').strip()
+            if compiled_pattern.search(log_entry_str):
+                matching_logs.append(log_entry_str)
+
+        return '\n'.join(matching_logs)
+
+    def generate_deployment_file(self, warnet):
+        """
+        TODO: implement this
```
willcl-ark: I'm just skimming the code, but how does it work without this? As I recall josie was writing a docker-compose.yml file, then converting it, but even that could go in this override? Or something I'm missing?

src/interfaces/kubernetes_interface.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def warnet_from_deployment(self, warnet):
+        # Get pod details from Kubernetes deployment
+        pods = self.client.list_namespaced_pod(namespace="default")
+        for pod in pods.items:
+            tank = self.tank_from_deployment(pod, warnet)
+            if tank is not None:
+                warnet.tanks.append(tank)
+
+    def tank_from_deployment(self, pod, warnet):
```
willcl-ark: We changed docker-interface to use the deployment file (docker-compsoe.yml) for this at some point, and I think it might be a prefereable approach? @pinheadmz 

See: 

https://github.com/bitcoin-dev-project/warnet/blob/5535564f9dbbbe69993b0f85bcf1028c7866f3bb/src/interfaces/docker_interface.py#L385-L407

and callsite

https://github.com/bitcoin-dev-project/warnet/blob/5535564f9dbbbe69993b0f85bcf1028c7866f3bb/src/interfaces/docker_interface.py#L374-L383

src/interfaces/kubernetes_interface.py
```diff
@@ -0,0 +1,263 @@
...
+        # or one under the users control
+        # TODO: support custom builds
+        # TODO: pass a custom namespace , e.g. different warnet sims can be deployed into diff namespaces
+        return client.V1Pod(
+            api_version="v1",
+            kind="Pod",
+            metadata=client.V1ObjectMeta(name=tank.container_name, namespace="default"),
+            spec=client.V1PodSpec(
+                # Might need some more thinking on the pod restart policy, setting to Never for now
+                # This means if a node has a problem it dies
```
willcl-ark: I think `Never` is probably what we will always want? @pinheadmz

src/services/fluentd.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, docker_network, config_dir):
         super().__init__(docker_network, config_dir)
         self.service = {
             "image": "fluent/fluentd:v1.16-debian-1",  # Debian version is recommended officially since it has jemalloc support.
-            "container_name": f"{self.docker_network}_fluentd",
```
willcl-ark, line 14: I'm a bit surprised we (I!) didn't remove this file in either 9a883819a031b9fb831ecba1f75a8629668fcdf8 or 5b55940bbd4f61f434352ccdec609929ff958dfb

The change is fine though. Docker isn't using it, and I hope in the future k8s won't either

willcl-ark (commented):
Nice. Left a few nits and comments. Not tested it yet either.

src/interfaces/kubernetes_interface.py
m3dwards: I tried to remove the warnet-bitcoin-core images but I don't think I have the needed credentials

pyproject.toml
m3dwards: Cool feature. I've opted to just keep it one list for now. Used compatible version specifier "kubernetes~=28.1.0".

src/interfaces/docker_interface.py
m3dwards: I think perhaps Josie changed it to keep things consistent between k8s and docker compose?

src/interfaces/kubernetes_interface.py
m3dwards: See comment: https://github.com/bitcoin-dev-project/warnet/pull/178#discussion_r1403756754

src/interfaces/kubernetes_interface.py
m3dwards: Added a TODO

src/interfaces/kubernetes_interface.py
m3dwards: Right now there is no intermediate file describing the network. Up just creates k8s resources directly using the k8s python client.

In these comments Josie has suggested that perhaps we should create resource files

https://github.com/bitcoin-dev-project/warnet/blob/1bf977cf2948bfbaaf2ca14f1fadecfb620bb204/src/interfaces/kubernetes_interface.py#L233-L235

Personally I don't really understand the benefit of creating files and then deploying those files. I can see that you could parse those files to do the "warnet_from_deployment" without asking k8s for what it has running but that doesn't seem much of a problem to me.

Are there any other benefits to an intermediate file that I'm not thinking or aware of?

willcl-ark: Oh. I just thought I recalled Josie saying his workflow was:

* generate docker-compose.yml
* convert it to k8s using `kompose`
* some manual steps/other things
* ???
* working

m3dwards: I had not heard of using that strategy.

src/warnet/test_framework_bridge.py
```diff
@@ -45,10 +45,10 @@ def setup(self):
         ch.setFormatter(formatter)
         self.log.addHandler(ch)
 
-        self.warnet = Warnet.from_network(self.options.network)
+        self.warnet = Warnet.from_network(self.options.network, backend="docker")
```
pinheadmz: Scenarios run in subprocess so they will need to be explicitly told what backend to use by (I suppose) a command-line argument executed by `scenarios_run()` in `server.py`

src/backends/kubernetes_backend.py
```diff
@@ -0,0 +1,353 @@
...
+    def get_pod_name(self, tank_index: int) -> str:
+        return f"{self.network_name}-{POD_PREFIX}-{tank_index:06d}"
+
+    def get_pod(self, container_name: str) -> V1Pod:
+        return cast(V1Pod, self.client.read_namespaced_pod(name=container_name, namespace="default"))
+
+    # We could enhance this by checking the pod status as well
+    # The following pod phases are available: Pending, Running, Succeeded, Failed, Unknown
+    # For example not able to pull image will be a phase of Pending, but the container status will be ErrImagePull
+    def get_status(self, tank_index: int, service: ServiceType) -> RunningStatus:
```
pinheadmz, line 95: `warcli network status` error: `TypeError: Object of type V1PodStatus is not JSON serializable`

Stack:

```
2023-12-01 19:54:24,091 - warnet.server - ERROR - Exception on /api [POST]
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/usr/local/lib/python3.9/site-packages/flask/views.py", line 109, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)
  File "/usr/local/lib/python3.9/site-packages/flask/views.py", line 190, in dispatch_request
    return current_app.ensure_sync(meth)(**kwargs)
  File "/usr/local/lib/python3.9/site-packages/flask_jsonrpc/views.py", line 49, in post
    return make_response(jsonify(response), status_code, headers)
  File "/usr/local/lib/python3.9/site-packages/flask/json/__init__.py", line 170, in jsonify
    return current_app.json.response(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/flask/json/provider.py", line 215, in response
    f"{self.dumps(obj, **dump_args)}\n", mimetype=self.mimetype
  File "/usr/local/lib/python3.9/site-packages/flask/json/provider.py", line 180, in dumps
    return json.dumps(obj, **kwargs)
  File "/usr/local/lib/python3.9/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/local/lib/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/lib/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/lib/python3.9/site-packages/flask/json/provider.py", line 120, in _default
    raise TypeError(f"Object of type {type(o).__name__} is not JSON serializable")
TypeError: Object of type V1PodStatus is not JSON serializable
```

pinheadmz (commented):
Great work so far! I'm already impressed how many features are still working ;-)

In addition to the two issues below I also noticed some help docs will need to be ambiguated as well, e.g.:

```
def down(network: str):
    """
    Run 'docker compose down on a warnet named <--network> (default: "warnet").
    """
```

src/backends/kubernetes_backend.py
```diff
@@ -0,0 +1,353 @@
...
+        self.deploy_pods(warnet)
+        return True
+
+    def down(self, warnet) -> bool:
+        """
+        Bring an exsiting network down.
+            e.g. `k delete -f warnet-tanks.yaml`
+        """
+        for tank in warnet.tanks:
+            self.client.delete_namespaced_pod(tank.container_name, self.namespace)
```
pinheadmz: I don't think `tank.container_name` is defined any more. When I try `warcli network down` in k8s mode I get an error:

```
Exception Exception in function 'wrapper' with context (args=(<warnet.warnet.Warnet object at 0xffff7fa7e880>,), kwargs={}, e=This method isn't implemented yet, func_name=wrapper, func=<function 
Warnet.warnet_down at 0xffff7fd35670>): This method isn't implemented yet

```

src/warnet/test_framework_bridge.py
m3dwards: I've not spent a lot of time thinking about tests. Do we want to run all 5 tests on compose and on k8s?

src/backends/kubernetes_backend.py
m3dwards: Thanks, got this and a few more other bits on the k8s backend that aren't currently working or complete. `get_file` is untested and build custom image isn't even past concept stage.

src/warnet/test_framework_bridge.py
pinheadmz: To be clear, even though the file name is test_ something something, this module is actually used for scenarios which is a feature of warnet, not actual tests or CI

willcl-ark:
```
 Testing warcli network export
2023-12-04 20:17:38,711 - warnet - DEBUG - {'jsonrpc': '2.0', 'method': 'network_export', 'params': {'network': 'warnettestf4ogvb9l'}, 'id': 1}
2023-12-04 20:17:38,736 - warnet - ERROR - Exception occurred while exporting network: Exception in function 'wrapper' with context (args=(<warnet.warnet.Warnet object at 0x7fa7a5742ad0>, '/home/runner/.config/warnet/warnettestf4ogvb9l/simln'), kwargs={}, e='Warnet' object has no attribute 'backend', func_name=wrapper, func=<function Warnet.export at 0x7fa7adc1e440>): 'Warnet' object has no attribute 'backend'
2023-12-04 20:17:38,736 - werkzeug - INFO - 127.0.0.1 - - [04/Dec/2023 20:17:38] "POST /api HTTP/1.1" 200 -
  File "/home/runner/work/warnet/warnet/./test/ln_test.py", line 23, in <module>
    with open(path, "r") as file:
FileNotFoundError: [Errno 2] No such file or directory: "Exception Exception in function 'wrapper' with context \n(args=(<warnet.warnet.Warnet object at 0x7fa7a5742ad0>, \n'/home/runner/.config/warnet/warnettestf4ogvb9l/simln'), kwargs={}, e='Warnet' \nobject has no attribute 'backend', func_name=wrapper, func=<function \nWarnet.export at 0x7fa7adc1e440>): 'Warnet' object has no attribute 'backend'/sim.json"
```

willcl-ark:
Seems like `Warnet.backend` has gone missing?

How about something like this:

```diff
diff --git a/src/backends/backend_interface.py b/src/backends/backend_interface.py
index d49b2e0..ef80c99 100644
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -14,6 +14,14 @@ class BackendInterface(ABC):
         self.config_dir = config_dir
         self.client = None
 
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """
+        Name of the backend
+        """
+        raise NotImplementedError("This method should be overridden by child class")
+
     @abstractmethod
     def build(self) -> bool:
         """
@@ -100,4 +108,4 @@ class BackendInterface(ABC):
         """
         Rebuild a warnet object from an active deployment
         """
-        raise NotImplementedError("This method should be overridden by child class")
\ No newline at end of file
+        raise NotImplementedError("This method should be overridden by child class")
diff --git a/src/backends/compose_backend.py b/src/backends/compose_backend.py
index 85713e3..d87e68b 100644
--- a/src/backends/compose_backend.py
+++ b/src/backends/compose_backend.py
@@ -38,11 +38,15 @@ logging.getLogger("docker.auth").setLevel(logging.WARNING)
 
 
 class ComposeBackend(BackendInterface):
+
     def __init__(self, config_dir: Path, network_name: str) -> None:
         super().__init__(config_dir)
         self.network_name = network_name
         self.client = docker.DockerClient = docker.from_env()
 
+    @property
+    def name(self) -> str:
+        return "compose"
 
     @bubble_exception_str
     def build(self) -> bool:
diff --git a/src/backends/kubernetes_backend.py b/src/backends/kubernetes_backend.py
index ebcfd2f..6481f06 100644
--- a/src/backends/kubernetes_backend.py
+++ b/src/backends/kubernetes_backend.py
@@ -37,6 +37,10 @@ class KubernetesBackend(BackendInterface):
         self.logs_pod = logs_pod
         self.network_name = network_name
 
+    @property
+    def name(self) -> str:
+        return "kubernetes"
+
     def build(self) -> bool:
         # TODO: just return true for now, this is so we can be running either docker or k8s as a backend
         # on the same branch
diff --git a/src/warnet/warnet.py b/src/warnet/warnet.py
index 8b9a782..a33d219 100644
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -209,7 +209,7 @@ class Warnet:
 
     @bubble_exception_str
     def export(self, subdir):
-        if self.backend != "compose":
+        if self.container_interface.name != "compose":
             raise NotImplementedError("Export is only supported for compose backend")
         config = {"nodes": []}
         for tank in self.tanks:
```

m3dwards:
Just added a self.backend in the constructor of the warnet object.

src/backends/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .backend_interface import BackendInterface
```
willcl-ark:
```suggestion
from .backend_interface import BackendInterface, ServiceType
```

This will let all `from backend import x` work in other files without the relative `.`

src/backends/compose_backend.py
```diff
@@ -10,15 +10,15 @@
 import docker
 from docker.models.containers import Container
 
-from .interfaces import ContainerInterface
-from warnet.utils import bubble_exception_str, parse_raw_messages
+from .backend_interface import BackendInterface, ServiceType
```
willcl-ark:
```suggestion
from backend import BackendInterface, ServiceType
```

src/backends/kubernetes_backend.py
```diff
@@ -0,0 +1,365 @@
+from .backend_interface import BackendInterface, ServiceType
```
willcl-ark:
```suggestion
from backend import BackendInterface, ServiceType
```

src/warnet/cli/network.py
```diff
@@ -109,31 +109,13 @@ def status(network: str):
     """
     try:
         result = rpc_call("network_status", {"network": network})
-
-        # Preprocess the data to add color for "running" status
-        colored_result = []
```
willcl-ark, line 114: 😭  😋

src/warnet/lnnode.py
```diff
@@ -1,84 +1,31 @@
...
 import json
 import os
-from docker.models.containers import Container
 from warnet.utils import (
     exponential_backoff,
     generate_ipv4_addr
 )
-
-CONTAINER_PREFIX_LN = "ln"
+from backends.backend_interface import BackendInterface, ServiceType
```
willcl-ark:
```suggestion
from backends import BackendInterface, ServiceType
```

src/warnet/server.py
```diff
@@ -23,11 +23,13 @@
     create_graph_with_probability,
     gen_config_dir,
 )
+from backends.backend_interface import ServiceType
```
willcl-ark: Think this is not used?

src/warnet/tank.py
```diff
@@ -12,9 +12,10 @@
     sanitize_tc_netem_command,
     SUPPORTED_TAGS,
 )
+from backends.backend_interface import ServiceType
```
willcl-ark:
```suggestion
from backends import ServiceType
```

src/warnet/warnet.py
```diff
@@ -91,9 +92,9 @@ def _warnet_dict_representation(self) -> dict:
     @classmethod
     @bubble_exception_str
     def from_graph_file(
-        cls, base64_graph: str, config_dir: Path, network: str = "warnet"
```
willcl-ark, line 94: IMO if anything the formatting change here should be:

```python
    def from_graph_file(
        cls,
        base64_graph: str,
        config_dir: Path,
        network: str = "warnet",
        backend: str = "compose",
    ):
```

But I think after we merge I am going to run `ruff` over the entire codebase quickly before any followups are started, to unify everything.

test/test_base.py
```diff
@@ -145,13 +154,20 @@ def wait_for_all_tanks_status(self, target="running", timeout=20*60, interval=5)
         def check_status():
             tanks = self.wait_for_rpc("network_status", {"network": self.network_name})
             stats = {
-                "total": len(tanks)
```
willcl-ark, line 148: Why this change?

Is this so we account for lightning containers too? or something else

willcl-ark (approved):
Looks good to me.

Happy to merge this as a stepping stone towards full/tidy k8s compatibility to avoid the PR getting too monstrous.

Docker side of things all working nicely locally too. I am afriad I can't test much of the k8s stuff right now very easily.

If you and @pinheadmz are happy with that side of things then I'm happy.

test/test_base.py
m3dwards: Correct, it's because the total is number of containers which isn't 1 to 1 with number of tanks when we have lightning

m3dwards:
@willcl-ark thanks for your review. Implemented your suggestions.

willcl-ark:
Do custom builds in k8s work with this? i.e when someone provides a custom branch, are we still applying Josie's k8 networking hack to the custom branch build?

m3dwards:
@willcl-ark custom builds are not available yet with k8s.

src/interfaces/docker_interface.py
josibake: IIRC, k8s was choking on the underscores, but im not entirely sure this is necessary. aesthetically, I
m a staunch hyphen maxi :sweat_smile:

src/interfaces/kubernetes_interface.py
josibake: Originally, what @willcl-ark has listed was my initial thinking, but once I started implementing, it seemed like a very hamfisted way of doing it and introduced an extra dependency on "Kompose" for doing the conversion. I much prefer not making k8s depend on docker-compose and also think we should make our docker backend less reliant on docker-compose long term.

With regards to writing deployment files for k8s, I'm now ambivalent. It seems nice as a "break glass in case of fire" where something happens to the k8s cluster and we want to re-deploy warnet, either on a different cluster or the same cluster once it has recovered. Having the files makes it feel more "reproducible." However, I'm not convinced its worth the extra work? Curious what your thoughts are @willcl-ark , @m3dwards

josibake: another point for having the files: they could be written to a PV, which would allow us to shut the cluster down and then restart and re-deploy warnet? but perhaps there are other mechanisms for doing this in k8s that are more elegant?

src/warnet/test_framework_bridge.py
josibake: Great point, we should probably rename this file to avoid confusion @pinheadmz . IIRC, this "just works" right now because the warnet RPC server runs in its own container and the scenarios still run in subprocesses, all on the same container. I think this is fine for now but we might want to be more clever about this in the future. Could even run each scenario as its own container, but that seems a bit overkill?

src/backends/kubernetes_backend.py
josibake: I think these serializable errors can be fixed with a `cast` argument, btw. I seem to remember doing that in a few places where the errors were annoying me

src/backends/kubernetes_backend.py
josibake: hm, this error is because `warnet_down` is not implemented in the K8s backend class, but is present in the abstract base class.. but you might also be right about `container_name`

josibake:
> ```
>  Testing warcli network export
> 2023-12-04 20:17:38,711 - warnet - DEBUG - {'jsonrpc': '2.0', 'method': 'network_export', 'params': {'network': 'warnettestf4ogvb9l'}, 'id': 1}
> 2023-12-04 20:17:38,736 - warnet - ERROR - Exception occurred while exporting network: Exception in function 'wrapper' with context (args=(<warnet.warnet.Warnet object at 0x7fa7a5742ad0>, '/home/runner/.config/warnet/warnettestf4ogvb9l/simln'), kwargs={}, e='Warnet' object has no attribute 'backend', func_name=wrapper, func=<function Warnet.export at 0x7fa7adc1e440>): 'Warnet' object has no attribute 'backend'
> 2023-12-04 20:17:38,736 - werkzeug - INFO - 127.0.0.1 - - [04/Dec/2023 20:17:38] "POST /api HTTP/1.1" 200 -
> Traceback (most recent call last):
>   File "/home/runner/work/warnet/warnet/./test/ln_test.py", line 23, in <module>
>     with open(path, "r") as file:
> FileNotFoundError: [Errno 2] No such file or directory: "Exception Exception in function 'wrapper' with context \n(args=(<warnet.warnet.Warnet object at 0x7fa7a5742ad0>, \n'/home/runner/.config/warnet/warnettestf4ogvb9l/simln'), kwargs={}, e='Warnet' \nobject has no attribute 'backend', func_name=wrapper, func=<function \nWarnet.export at 0x7fa7adc1e440>): 'Warnet' object has no attribute 'backend'/sim.json"
> ```

this is for Sim-LN, right?

src/interfaces/kubernetes_interface.py
m3dwards: I can see some times when it would be useful. I kept waiting for a point where it was needed and I hadn't hit that yet so just kept it dynamic.

m3dwards:
Yes it's for simln. If you run that test as `./test/ln_test.py k8s` then it should skip the network export.

src/interfaces/kubernetes_interface.py
willcl-ark: Can be added as a followup (or issue) if not needed urgently

src/interfaces/kubernetes_interface.py
willcl-ark: ok

willcl-ark (approved):
ACK b6217546bdf4dbb8299057ee6bc958fb615f8324

josibake:
ACK https://github.com/bitcoin-dev-project/warnet/pull/178/commits/be51634ac928fd3297e0865976a40c794a232b9a

21 commits.. how fitting

willcl-ark (approved):
re-ACK be51634ac928fd3297e0865976a40c794a232b9a

PR #197, opened 2023-11-09
scenarios: add small sleep between sends

willcl-ark:
give single CPU setups (docker) time to context switch and keep up


willcl-ark:
Lets me run with 400 nodes pretty stably on my machine (TM). Not sure that it's _essential_, but seems like a good idea in any case..

willcl-ark:
Hmmm, perhaps we should bake this into rpc method instead...

willcl-ark:
Closing for now

PR #200, opened 2023-11-10
graph: switch to erdos renyi graphs

### LINKED ISSUE #199: graph create doesn't connect all nodes
See a 250 node graph I created: https://github.com/pinheadmz/warnet/blob/blockrelay-capacity/src/graphs/block_relay.graphml#L2858

You can see the edge sources are limited in range form node 0 up to 43. The destinations are properly mixed up and that looks ok but when I run this network there are many nodes with 0 or 1 peers

willcl-ark:
* Doesn't change default graph, only user-generated
* Avoid nodes with single edge connections.
* Erdos renyi provides good random starting point.
* Default to edge connection probabiliyt of p=0.2 if unspecified.
* Top up to 8 edges per node as a minimum

Closes: #199

### DIFF
--- a/docs/graph.md
+++ b/docs/graph.md
@@ -6,7 +6,7 @@ Before any scenarios or RPC commands can be executed, a Warnet network must be s
 See [warcli.md](warcli.md) for more details on these commands.
 
 To start a network called `"warnet"` from the [default graph file](../src/graphs/default.graphml)
-which consists of 12 Bitcoin Core v25.0 nodes connected in a ring:
+which consists of 12 Bitcoin Core v26.0 nodes connected in a ring:
 ```
 warcli network start
 ```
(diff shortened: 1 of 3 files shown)


pinheadmz:
I think we need to get the DNS seeder working again using the pynode approach we've discussed. As far as `graph create` I wonder if we should just ditch the networkx stuff and write our own thing. How hard could it be to just give each tank 12 random outgoing edges?

willcl-ark:
Obviously not hard. The whole reason to use networkx is so that it generates graphml files, and they can be interesting shapes? That's why I added a generic generator first...

Gimped to a single shape, we might as well write our own.

Personally I'd prefer having the generic generator back, but 🤷🏼‍♂️

willcl-ark:
This does not change the default graph (for now), but makes user-generated graphs more useful and updates `graphs.md` to be more current.

Also going to merge for now.

PR #201, opened 2023-11-20
cli: use rich.table Table to auto-format tables

willcl-ark:


willcl-ark:
Switched to use `rich.table Table` as we already have that dep, so no need for tabulate.

Kept server returning a dict and did table processing inside CLI, which kept the tests happier (and made more sense).

screenies

image

image

image

src/warnet/warnet.py
```diff
@@ -55,6 +55,39 @@ def __str__(self) -> str:
             f"Tanks:\n{tanks_str}"
         )
 
+    def _wn_repr(self) -> dict:
```
pinheadmz: what does "_wn_repr" mean?

pinheadmz (commented):
ACK 
looks good nice!
one optional nit, function name is confusing

willcl-ark:
> ACK 
> 
> looks good nice!
> 
> one optional nit, function name is confusing

Which, `print_repr`?

src/warnet/warnet.py
pinheadmz: @willcl-ark

willcl-ark: Warnet Representation

Its a python thing: https://docs.python.org/3/library/functions.html#repr

But I can change if its blocking?

willcl-ark: Actually, now that it returns a `dict` I don't think it's a "real" repr :P

In a mid-edit version I had this returning the `tabulate` tables as strings, but no longer...

willcl-ark: Updated in 807dab37497b535470c16dd50c3574cacc28b097

pinheadmz (approved):
|review|
|-----|
| ACK |

PR #202, opened 2023-11-21
Send graph via base64 rather than file path

m3dwards:
Supports scenarios where the RPC is running on a different machine and doesn't have access to the file paths

### DIFF
--- a/src/warnet/cli/network.py
+++ b/src/warnet/cli/network.py
@@ -1,3 +1,4 @@
+import base64
 from pathlib import Path
 
 import click
@@ -24,9 +25,12 @@ def start(graph_file: Path, force: bool, network: str):
     Start a warnet with topology loaded from a <graph_file> into <--network> (default: "warnet")
     """
     try:
+        encoded_graph_file = ""
+        with open(graph_file, "rb") as graph_file_buffer:
+            encoded_graph_file = base64.b64encode(graph_file_buffer.read()).decode("utf-8")
         result = rpc_call(
             "network_from_file",
-            {"graph_file": str(graph_file), "force": force, "network": network},
+            {"graph_file": encoded_graph_file, "force": force, "network": network},
         )
         print(result)
     except Exception as e:
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -2,6 +2,7 @@
   Warnet is the top-level class for a simulated network.
 """
 
+import base64
 import json
 import logging
 import networkx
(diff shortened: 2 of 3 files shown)


src/warnet/cli/network.py
```diff
@@ -24,9 +25,12 @@ def start(graph_file: Path, force: bool, network: str):
     Start a warnet with topology loaded from a <graph_file> into <--network> (default: "warnet")
     """
     try:
+        encoded_graph_file = ""
+        with open(str(graph_file), "rb") as graph_file_buffer:
```
willcl-ark:
```suggestion
        with open(graph_file, "rb") as graph_file_buffer:
```

Don't think we should need the `str()` here? `open` can handle `Path`

src/warnet/warnet.py
```diff
@@ -62,9 +63,11 @@ def from_graph_file(
         self = cls(config_dir)
         destination = self.config_dir / self.graph_name
         destination.parent.mkdir(parents=True, exist_ok=True)
-        shutil.copy(graph_file, destination)
```
willcl-ark, line 65: nit: could have used clearer naming in here

```diff
diff --git a/src/warnet/warnet.py b/src/warnet/warnet.py
index 9506c21..2255a85 100644
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -58,16 +58,16 @@ class Warnet:
     @classmethod
     @bubble_exception_str
     def from_graph_file(
-        cls, graph_file: str, config_dir: Path, network: str = "warnet"
+        cls, base64_graph: str, config_dir: Path, network: str = "warnet"
     ):
         self = cls(config_dir)
         destination = self.config_dir / self.graph_name
         destination.parent.mkdir(parents=True, exist_ok=True)
-        file_content = base64.b64decode(graph_file.encode("utf-8"))
+        graph_file = base64.b64decode(base64_graph.encode("utf-8"))
         with open(destination, "wb") as f:
-            f.write(file_content)
+            f.write(graph_file)
         self.network_name = network
-        self.graph = networkx.parse_graphml(file_content.decode("utf-8"), node_type=int)
+        self.graph = networkx.parse_graphml(graph_file.decode("utf-8"), node_type=int)
         self.tanks_from_graph()
         logger.info(f"Created Warnet using directory {self.config_dir}")
         return self
```

unrelated, seems odd that my typechecker complains about the `node_type=int` on the networkx parse. AFAICT this looks like a valid type specified. I guess an issue with mypy...

willcl-ark (approved):
Looks pretty good to me, thanks!

Left two nits, which you can choose to address or not. Will leave up to our lead maintainer @pinheadmz to ACK/merge

src/warnet/cli/network.py
m3dwards: Removed str()

src/warnet/warnet.py
m3dwards: Agree, naming could be better, accepted this suggestion

willcl-ark:
LGTM

PR #203, opened 2023-11-21
Specify thread to be daemonized

m3dwards:
Looks like these threads are set to daemon anyway but this way it is explicit


willcl-ark:
LGTM, thanks

PR #204, opened 2023-11-21
docs: auto-generate list of CLI commands and insert into docs/warcli.md

pinheadmz:
uses tabulate assuming merge of #201 

check it out: https://github.com/pinheadmz/warnet/blob/api-docs/docs/warcli.md

### DIFF
--- a/scripts/apidocs.py
+++ b/scripts/apidocs.py
@@ -17,7 +17,7 @@ def print_cmd(cmd, super=''):
         p['name'],
         p['type']['param_type'],
         p['required'],
-        p['default']
+        p['default'] if p['type']['param_type'] != "Path" else Path(p['default']).relative_to(Path.cwd())
     ] for p in cmd["params"]]
     doc += tabulate(data, headers=headers, tablefmt="github")
     doc += "\n\n"
--- a/scripts/apidocs.py
+++ b/scripts/apidocs.py
@@ -1,0 +1,44 @@
+import re
+import os
+from click import Context
+from warnet.cli.main import cli
+from tabulate import tabulate
+from pathlib import Path
+
+doc = ""
+
+def print_cmd(cmd, super=''):
+    global doc
+    doc += f"###{super} {cmd['name']}" + "\n"
+    doc += cmd["help"].strip() + "\n"
+    doc += "\noptions:\n"
+    headers = ["name", "type", "required", "default"]
+    data = [[
+        p['name'],
+        p['type']['param_type'],
+        p['required'],
+        p['default']
+    ] for p in cmd["params"]]
+    doc += tabulate(data, headers=headers, tablefmt="github")
+    doc += "\n\n"
+
+
+with Context(cli) as ctx:
+    info = ctx.to_info_dict()
+    for cmd in info["command"]["commands"].values():
+        if "commands" not in cmd:
+            print_cmd(cmd)
+        else:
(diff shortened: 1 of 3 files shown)


willcl-ark:
Looks awesome to me! Nice.

I'll fix up #201 first then we can do this one?

pinheadmz:
> I'll fix up #201 first then we can do this one?

cool, good plan. might tweak the style here and there on this as well

willcl-ark:
Code LGTM in general.

Just gave this a quick whizz rebase on top of #211 and it seemed to work well, except it uses my graph file path which needs manual change

image

It could be fixed with a patch like this, I believe:

```patch
diff --git a/src/warnet/cli/network.py b/src/warnet/cli/network.py
index 87f2d97..b235d3f 100644
--- a/src/warnet/cli/network.py
+++ b/src/warnet/cli/network.py
@@ -1,5 +1,5 @@
 import base64
-from pathlib import Path
+from pathlib import Path, PurePath
 
 import click
 from rich import print
@@ -8,7 +8,7 @@ from graphs import GRAPHS
 from warnet.cli.rpc import rpc_call
 
 
-DEFAULT_GRAPH_FILE = GRAPHS / "default.graphml"
+DEFAULT_GRAPH_FILE = PurePath(GRAPHS / "default.graphml")
 
 
 @click.group(name="network")
@@ -17,7 +17,7 @@ def network():
 
 
 @network.command()
-@click.argument("graph_file", default=DEFAULT_GRAPH_FILE, type=click.Path())
+@click.argument("graph_file", default=DEFAULT_GRAPH_FILE.relative_to(Path.cwd()), type=click.Path())
 @click.option("--force", default=False, is_flag=True, type=bool)
 @click.option("--network", default="warnet", show_default=True)
 def start(graph_file: Path, force: bool, network: str):
```

...which would use the relative path from where the script was run, to the graph file, instead.

pinheadmz:
thanks for reviewing Will, I modified the script to convert paths to relative when needed: 17da8a4b077a527a635f4656ea0b3db716efee5f

scripts/apidocs.py
```diff
@@ -0,0 +1,48 @@
...
+    global doc
+    doc += f"### `warcli{super} {cmd['name']}`" + "\n"
+    doc += cmd["help"].strip().replace("<", "\\<") + "\n"
+    doc += "\noptions:\n"
+    headers = ["name", "type", "required", "default"]
+    data = [[
+        p['name'],
+        p['type']['param_type'],
+        p['required'],
+        p['default'] if p['type']['param_type'] != "Path" else Path(p['default']).relative_to(Path.cwd())
```
pinheadmz, line 20: only caveat: script must be run from root directory. im ok with that

willcl-ark: I think that's fine too

willcl-ark (approved):
ACK 17da8a4b077a527a635f4656ea0b3db716efee5f

Re-ran after new relative path changes and got a 0 line diff 👍🏼 

```
will@ubuntu in ~/src/warnet on  api-docs [$?] is 📦 v1.0.0 via  v18.16.0 : 🐍 warnet
₿ python scripts/apidocs.py

will@ubuntu in ~/src/warnet on  api-docs [$?] is 📦 v1.0.0 via  v18.16.0 : 🐍 warnet
₿ git diff
```

PR #206, opened 2023-11-22
Added docker image build script for k8s

m3dwards:
Kubernetes default networking prevents bitcoin core from routing traffic between nodes. Until we have a better workaround for that problem these images are of a modified bitcoin core that will route to private IP addresses.

Unlike warnets standard bitcoin-core images which download bitcoin core binary, these images are built from bitcoin core source so that we can apply a patch.


willcl-ark:
Nice work!

> Until we have a better workaround for that problem these images are of a modified bitcoin core that will route to private IP addresses.

One nit in the desc, not for you but for future devs: Core always routes (traffic) to private IP addresses, but the patch marks the `CNetAddr` as `IsRoutable()` which in turn affects it's "reachability" (according to Core) and therefore whether they should be gossiped (or not) on the network in `addr` messages.

Will take a proper look later.

josibake:
this needs a few changes:

- following a discussion in the chat, we should just use the same image for both docker and k8s, since we are moving towards just using the patch for now
- we dont need to download binaries 
- we should figure out how to build older tagged versions on an ad hoc basis (low priority for now)

scripts/build-bitcoin-images-k8s.sh
```diff
@@ -0,0 +1,54 @@
...
+        echo "IMAGE_FULL_NAME=${IMAGE_FULL_NAME}"
+
+        # Use Buildx to build the image for the specified architecture
+        docker buildx build --platform linux/"${DOCKER_ARCH}" \
+            --provenance=false \
+            --build-arg REPO="${REPO}" \
+            --build-arg BRANCH="v${VERSION}" \
+            --build-arg BUILD_ARGS="${BUILD_ARGS}" \
+            --tag "${IMAGE_FULL_NAME}" \
+            . --push
```
willcl-ark, line 48: This was not specifying `--file Dockerfile_k8s` so (for me) it was using the default `Dockerfile`.

This could be why @josibake observed networking not working? (perhaps patch not applied?)

m3dwards: Could have been. I might have built and pushed the images and then renamed the Dockerfile but forgot to update it here. I might be misremembering but I thought I saw bitcoind compiling.

m3dwards: We should re-push the images though as there is doubt.

josibake: were the images re-pushed? if not no worries, was going to do it now as part of testing this PR

willcl-ark: I did not re-push anything

willcl-ark:
Removed my commits for now, as we are not sure if the patch actually works?

josibake:
> Removed my commits for now, as we are not sure if the patch actually works?

why not just keep the PR open for now with the commits? still works for docker, we just tell docker to assign IPs in a `/8` subnet and can still remove code where we are manually generating and assigning the IPs.

still need to do some thinking about k8s, but until then i dont see any point in remove commits from the PR

willcl-ark:
The pr is open?

josibake:
> The pr is open?

"..with the commits"

willcl-ark:
LGTM.
Let's get this in for now then.

willcl-ark:
I am pushing v26.0 to [docker hub](https://hub.docker.com/repository/docker/bitcoindevproject/k8s-bitcoin-core/general) too, so that main graph should work on k8s

PR #211, opened 2023-11-27
add persist option to network_down

### LINKED ISSUE #209: Prevent docker compose based warnet from downloading the same docker images so many times
This could be more of an issue for developers of warnet who are running frequent tests rather than users who will likely just pull once.

A suggestion could be to add a flag to request to keep the images when stopping (compose down) a network.

Alternatively, would developer documentation suggesting to manually pull commonly used images solve the problem as I think compose will only remove images that it pulled as part of it's up command.

willcl-ark:
Closes: #209

For docker, this enables keeping the images in local docker registry and not needing to pull them again.

Use with:

```
docker network down --persist
```

This is off by default, unsure whether on-by-default makes more sense? It's mainly going to be folks like us using it?...

### DIFF
--- a/src/interfaces/docker_interface.py
+++ b/src/interfaces/docker_interface.py
@@ -1,6 +1,7 @@
 import logging
 import re
 import shutil
+import shlex
 import subprocess
 import yaml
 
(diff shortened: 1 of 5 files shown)


src/interfaces/docker_interface.py
```diff
@@ -82,23 +83,27 @@ def up(self):
...
-                if process.stdout:
-                    for line in process.stdout:
-                        logger.info(line.decode().rstrip())
-        except Exception as e:
-            logger.error(
-                f"An error occurred while executing `{' '.join(command)}` in {self.config_dir}: {e}"
-            )
+    def down(self, persist):
+        if persist:
+            commands = [shlex.split("docker compose stop"), shlex.split("docker compose rm --force")]
```
pinheadmz, line 88: so, wait. if `persist` is `True` then we execute `rm --force` ?

pinheadmz: Also, do we really need another module import just to convert a string with spaces to a list? :-P

willcl-ark:
re-tested and `main` does not persist images for me

image

willcl-ark:
Doh, maybe I'm wrong. No idea what `ls` is doing here, but it seems the images _are_ in `docker images` (but not `docker images ls`)

image

willcl-ark:
Going to close for now as this seems fine

PR #214, opened 2023-12-05
Use ruff formatter

willcl-ark:
No more thinking about formatting. Also includes an action so we don't regress.

Run locally with the following (which will use the rules from pyproject.toml):

* linter: `ruff check .` or to auto-fix `ruff check . --fix`
* formatter: `ruff format .`

Docs: https://docs.astral.sh/ruff/


willcl-ark:
We should run a formatter over the codebase and enforce it going forward.

`black` or `ruff` should produce the same format IIUC, but seems like `ruff` is superior.

Issue #189, opened 2023-10-31
Fix IP addresses 


josibake:
Currently, we still have a method that randomly generates a fake IP address if `Tank._ipv4` is None. To get around this, I'm waiting for the pods to spin up and then getting the IP and setting it on the tank, which is pretty gross and will be super slow when spinning up 100s of nodes. We should move all the IP functions into the abstract base class , and grab the IP directly from the container when we need it (same as we were doing in docker land).

willcl-ark:
I think this would be moot with #303 

What did you want IP addresses for? We using hostnames now yo!

willcl-ark:
Closing as we use hostnames externally and patches on bitcoind

PR #215, opened 2023-12-07
move IP generators into an abstract base class

TChukwuleta:
A proposed fix for #189 

Created an abstract base class for the IP generator and let every other class that needs the fake IP generator including the Tank as well as LNNode inherit it.

@josibake  Kindly review. Thanks

### DIFF
--- a/src/warnet/ip_generator.py
+++ b/src/warnet/ip_generator.py
@@ -3,10 +3,6 @@ import ipaddress
 import random
 
 class IPV4AddressGenerator(ABC):
-    @abstractmethod
-    def get_ip(self):
-        pass
-
     def generate_ipv4_addr(self, subnet):
         """
         Generate a valid random IPv4 address within the given subnet.
--- a/src/warnet/ip_generator.py
+++ b/src/warnet/ip_generator.py
@@ -1,0 +1,49 @@
+from abc import ABC, abstractmethod
+import ipaddress
+import random
+
+class IPV4AddressGenerator(ABC):
+    @abstractmethod
+    def get_ip(self):
+        pass
+
+    def generate_ipv4_addr(self, subnet):
+        """
+        Generate a valid random IPv4 address within the given subnet.
+
+        :param subnet: Subnet in CIDR notation (e.g., '100.0.0.0/8')
+        :return: Random IP address within the subnet
+        """
+        reserved_ips = [
+            "0.0.0.0/8",
+            "10.0.0.0/8",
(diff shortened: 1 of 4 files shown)


josibake:
Hey @TChukwuleta ! Welcome to warnet :smile: 

Now that we are closer to having the kubernetes backend finished, I think we are going to move away from setting the IP addresses manually all together, and instead just apply a patch to bitcoin core that ignores the `IsRoutable()` function. This simplifies things quite a bit and means we will probably get rid of all of this IP related code altogether.

The patched docker image is being worked on here: https://github.com/bitcoin-dev-project/warnet/pull/206

I think we can keep this PR open for now until we have the patched docker images full ironed out, in case we need to back to generating the addresses.

TChukwuleta:
> Hey @TChukwuleta ! Welcome to warnet 😄
> 
> Now that we are closer to having the kubernetes backend finished, I think we are going to move away from setting the IP addresses manually all together, and instead just apply a patch to bitcoin core that ignores the `IsRoutable()` function. This simplifies things quite a bit and means we will probably get rid of all of this IP related code altogether.
> 
> The patched docker image is being worked on here: #206
> 
> I think we can keep this PR open for now until we have the patched docker images full ironed out, in case we need to back to generating the addresses.

Thanks Josie. would there be anything to do to assist on getting out the patched docker image? else, I'd look for another issue to hop on

willcl-ark:
If we use the k8s images (post https://github.com/bitcoin-dev-project/warnet/pull/241 ) in docker, we can get rid of all the ip address generation stuff from the repo.

Even if we want to add ASMap (https://github.com/bitcoin-dev-project/warnet/pull/217) we can simply map the assigned ip address to a new AS manually, as I did in the linked PR...

@m3dwards @josibake @pinheadmz

josibake:
> If we use the k8s images (post #241 ) in docker, we can get rid of all the ip address generation stuff from the repo.
> 
> Even if we want to add ASMap (#217) we can simply map the assigned ip address to a new AS manually, as I did in the linked PR...
> 
> @m3dwards @josibake @pinheadmz

Tend to agree as I don't think we'd even have the ability to set IP addresses on a managed k8s cluster. Also, with the isroutable patch for docker, I don't think we need to manually specify addresses, just need to pick the appropriate subnet. I'd say we close this for now and then can reopen if we do end up needing the IP addresses.

m3dwards:
Thanks for PR @TChukwuleta but I have to agree that it shouldn't be needed now.

PR #216, opened 2023-12-07
Add lightning cli command to warnet cli and warnet rpc

### LINKED ISSUE #213: LN RPC is missing from CLI and Warnet RPC
It would be useful to be able to run lightning rpc commands from warcli and warnet rpc.

This is needed to make the ln_test.py test run under k8s as right now it relies on creating an internal warnet object to run lightning rpc commands which is not feasible when warnet is running on another machine or k8s cluster.

m3dwards:
Fixes: #213

### DIFF
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
 from enum import Enum
+from typing import List
 
 
 class ServiceType(Enum):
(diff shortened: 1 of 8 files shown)


pinheadmz:
concept ACK, rebased on main bc we dont have arm images for 25.1 😭

pinheadmz (approved):
ACK
tested locally and fixed lint for CI.  if the light is green, merge it!

PR #217, opened 2023-12-12
ASMap v2

willcl-ark:
Another go at using ASMap.

This time do not rely on copying in asmap.dat when we build the container, instead (in Dockerland) mount it as a volume into each container.

Missing:

* K8s equivalent action to mount generated ASMap into pods
* Reloading tank AS from asmap.txt (can be done, not sure if necessary for server to ever know of this after initialisation?)
* Get rid of hardcoded variables in compose_backend.py volume mount
* Check that using AS numbers in range 1 - 64496 is correct
* Check that mapping an ip address /32 to an AS is correct

### DIFF
--- a/src/warnet/tank.py
+++ b/src/warnet/tank.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from warnet.lnnode import LNNode
 from warnet.utils import (
     exponential_backoff,
+    generate_as,
     generate_ipv4_addr,
     sanitize_tc_netem_command,
     SUPPORTED_TAGS,
@@ -41,6 +42,7 @@ class Tank:
         self.rpc_password = "2themoon"
         self._suffix = None
         self._ipv4 = None
+        self._a_system = None
         self._exporter_name = None
         self.extra_build_args = ""
         self.lnnode = None
@@ -111,6 +113,12 @@ class Tank:
     @property
     def status(self) -> RunningStatus:
         return self.warnet.container_interface.get_status(self.index, ServiceType.BITCOIN)
+
+    @property
+    def autonomous_system(self) -> int:
+        if self._a_system is None:
+            self._a_system = generate_as()
+        return self._a_system
 
     @exponential_backoff()
     def exec(self, cmd: str, user: str = "root"):
--- a/src/warnet/utils.py
+++ b/src/warnet/utils.py
@@ -515,3 +515,9 @@ def convert_unsupported_attributes(graph):
                 continue
             else:
                 edge_data[key] = str(value)
+
+
+def generate_as():
+    while True:
+        as_number = random.randint(1, 64496)  # I think these are not "reserved"
+        return as_number
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -3,21 +3,26 @@
 """
 
 import base64
+import io
 import json
 import logging
 import networkx
 import os
 import shutil
+import sys
 from pathlib import Path
 from templates import TEMPLATES
 from typing import List, Optional
 
+import external.buildmap as buildmap
 from backends import ComposeBackend, KubernetesBackend
 from warnet.tank import Tank
 from warnet.utils import gen_config_dir, bubble_exception_str, version_cmp_ge
 
 logger = logging.getLogger("warnet")
 FO_CONF_NAME = "fork_observer_config.toml"
+ASMAP_TXT_PATH = "asmap.txt"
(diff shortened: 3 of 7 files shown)


src/warnet/tank.py
```diff
@@ -112,6 +114,12 @@ def exporter_name(self):
     def status(self) -> RunningStatus:
         return self.warnet.container_interface.get_status(self.index, ServiceType.BITCOIN)
 
+    @property
+    def autonomous_system(self):
+        if self._a_system is None:
+            self._a_system = generate_as(self.warnet)
+        return self._a_system
+
```
m3dwards: Personally, I'm not a huge fan of passing the warnet object in to tanks, especially in this case as it's having it's `a_systems` property updated by the tank.

IMHO if the warnet object needs to keep track of what's been assigned then the logic belongs with the warnet object to assign the tanks a autonomous_system number.

m3dwards (commented):
Looks good.

Applying this to k8s will be slightly trickier as we are not assigning IP addresses up front like we are with compose.

Might be nice to include a bit of type hinting, especially on the `tank.autonomous_system` and `warnet.a_systems` which I think are just ints?

I would like to see k8s support before merging to prevent us from having a design that only works on compose unless there is a pressing need to get this into compose quickly.

willcl-ark:
Thanks, I'll try address this this evening. Agree it's not ready yet.

It's very annoying (with asmap) the workflow requires us to start the pods, wait for ip addresses, and then pass those ip addresses back to bitcoind *before it starts* :S

willcl-ark:
OK I updated those nits, but it's still not going to be any good to merge... Need to brainstorm how it would be possible to build an ASMap on k8s.

The only "sane" way I can think of doing it, which might work on both systems, is to override the default command (`bitcoind`) with `cat /dev/null`, wait for the containers to start, fetch the ip addresses, build the asmap, load it in, then use `exec` to start bitcoind.

This might have it's own problems though, for example if bitcoind crashes the container won't notify as the "main" process is now `cat` :(

Any other less insane ideas?

m3dwards:
In k8s there is a concept of [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) which I think might be worth considering for this. Init containers are part of the same pod, they run to completion before the main workload container starts.

How about:

1. Have a single volume for the ASMap files that the RPC pod and all tank pods can mount
2. After creating tank pods, the RPC container generates the required ASMap files and writes them to that volume, named by tank index. It can query the kubernetes service host to get IP addresses for each tank.
3. Tank pod has init container that simply blocks until the required ASMap file is present for it's tank
4. bitcoind container has ASMap file preconfigured in flag, by the time it starts the file will be present on the mount

The reason I think we will need a single volume that all pods mount is that I don't think it's possible to mount the volume after pod creation. As the RPC pod is long running it would have started before any tanks and their respective volumes were created.

src/warnet/utils.py
```diff
@@ -515,3 +515,9 @@ def convert_unsupported_attributes(graph):
                 continue
             else:
                 edge_data[key] = str(value)
+
+
+def generate_as():
+    while True:
+        as_number = random.randint(1, 64496)  # I think these are not "reserved"
```
josibake, line 522: magic number alert! do we use this value anywhere else? even if not, might be good to set it as a global const with a nice comment explaining what the number means

src/warnet/warnet.py
```diff
@@ -238,3 +245,35 @@ def export(self, subdir):
...
+        # Write AS mappings to file
+        with open(self.as_map_txt_path, "w") as f:
+            for tank in self.tanks:
+                f.write(f"{tank.ipv4}/32 AS{tank.autonomous_system}\n")
+
+        # Yes, read back into a string...
+        with open(self.as_map_txt_path, "r") as f:
+            file_content = f.read()
+
+        buffer = io.StringIO(file_content)
```
josibake, line 261: why the write / read? are we using the as_map_txt file anywhere else?

josibake (commented):
Did a quick glance, mostly to familiarize myself, but agree we should focus on getting k8s in before merging / more serious review. Happy to start reviewing/testing if you have a WIP k8s support branch?

willcl-ark:
Going to close this for now. As we are looking like we are going to fallback to orchestrator-assigned ip addresses, either with @pinheadmz nftables approach, or patched images a-la #261, ASMap will not work easily in either case.

We can revisit if needed.

PR #221, opened 2023-12-21
Re-enable prometheus exporter

pinheadmz:
This adds a new option to the graph file schema. When `<data key="exporter">true</data>` is present in a node definition, it will be paired with a bitcoin-prometheus-exporter container, and the exported data is visible in grafana. I also included a new default dashboard "Bitcoin RPC dashboard" which shows a little sample of the RPC data collected:

Screen Shot 2023-12-21 at 3 06 42 PM


m3dwards:
LGTM. Merging.

PR #222, opened 2023-12-21
scenarios: force new http connection for evey RPC call

pinheadmz:
This modification to the bitcoin test framework prevents re-using old HTTP connections for new RPC calls. I believe this will prevent "broken pipe" errors in long-running scenarios.


m3dwards:
I can understand the code but haven't faced the same issue. Tests pass and I can't see a big downside so if it fixed something for you then happy to merge.

PR #223, opened 2024-01-03
Add Loki log aggregation

m3dwards:
Compose only.

Adds log aggregation using Loki and Promtail and makes it available in Grafana.

Screenshot 2024-01-04 at 14 14 28

Also moves services directory to within the backends and specifically the compose backend. Services will be backend specific.

Suggested steps to test:

1. Run warnet network on docker compose
2. Open http://localhost:3000/explore
3. Select Loki datasource
4. Use the following query to see logs from tanks: `{container=~".+tank.+"}`
5. Click Run Query

PR #225, opened 2024-01-16
pass manual config from graph to tank service

m3dwards:
Compose only. Manual config specified in the graph file now makes it to the docker-compose file.

This PR depends on https://github.com/bitcoin-dev-project/warnet/pull/223 so please merge that one first.


m3dwards:
Bug fix and not controversial so self merging.

PR #229, opened 2024-01-18
fixup error message when config dir already exists

willcl-ark:
fixes an oversight when we try a custom print of a dict on an error string


willcl-ark:
I think really we should return `json` from all RPCs to make it unified...

But this is a small improvement

PR #232, opened 2024-01-19
Fixup RPC responses to always be json

willcl-ark:
May introduce some double-jsoning, but seems to be fine on my end.

Opening as draft to see if the CI is happy.

Also improved the `server stop` command to show something nice, and shutdown gracefully on the Flask side.


m3dwards:
Ci is happy!

PR #233, opened 2024-01-19
JSON-IFY the server properly

### LINKED ISSUE #198: RPC functions in server should not catch errors
This kind of thing causes type errors if the CLI module is expecting anything besides a `[str]`. The error is swallowed and then the return value (with the precious error message!) causes another error with a stack trace that's entirely inside Flask and the type checker. My opinion is that RPC functions should not catch errors, they should throw and that error should bubble all the way back out to the caller.

```python
        except Exception as e:
            return [f"Exception {e}"]
```

willcl-ark:
Use proper `ServerErrors` and things to neaten and homogenize the server interfaces.

Gets rid of most of the `try: except` stuff in all the cli modules. Allows us to raise custom JSONRPC exceptions using the native interface on the server side. Auto-handled on the CLI side (well, printed for now). Should generally make things more robust.

Fixes: #198

### DIFF
--- a/src/warnet/cli/rpc.py
+++ b/src/warnet/cli/rpc.py
@@ -1,17 +1,27 @@
 import requests
-from jsonrpcclient.responses import Ok, parse
+import sys
+from jsonrpcclient.responses import Error, Ok, parse
 from jsonrpcclient.requests import request
 from typing import Any, Dict, Tuple, Union, Optional
 from warnet.server import WARNET_SERVER_PORT
 
 
+class JSONRPCException(Exception):
+    def __init__(self, code, message):
+        try:
+            errmsg = f"{code} {message}"
+        except (KeyError, TypeError):
+            errmsg = ""
+        super().__init__(errmsg)
+
+
 def rpc_call(rpc_method, params: Optional[Union[Dict[str, Any], Tuple[Any, ...]]]):
     payload = request(rpc_method, params)
     response = requests.post(f"http://localhost:{WARNET_SERVER_PORT}/api", json=payload)
-    parsed = parse(response.json())
-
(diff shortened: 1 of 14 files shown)


willcl-ark:
Force-pushed to undo debug mode on the server :P

willcl-ark:
@m3dwards this is the version that we want I think!

If this passes then IMO it's RFM 👀

willcl-ark:
OK not quite ready yet, spotted one more thing...

src/warnet/cli/rpc.py
```diff
@@ -1,17 +1,27 @@
...
-    else:
-        error_message = getattr(parsed, "message", "Unknown RPC error")
-        raise Exception(error_message)
+    match parse(response.json()):
+        case Ok(result, _):
+            return result
+        case Error(code, message, _, _):
+            print(f"{code}: {message}")
+            sys.exit(1)
+            # raise JSONRPCException(code, message)
```
willcl-ark, line 27: Not sure what's preferable here, to raise a custom JSONRPC Exception (nasty stacktrace), or simply print the error message and code.

I vote the latter, but have left the code in for the former, in case there are strong opinions...

willcl-ark:
OK @m3dwards @pinheadmz 

If anyone wants to test/review/merge this, I think it's in decent shape now :)

willcl-ark:
Oh, one other difference, in typical python fashion this spits out the entire stacktrace to the logfile on a server-side exception now. It looks messier in the logs, but I actually think it makes sense here. Expecially as the `ServerError` is being returned (which doesn't contain the stacktrace).

m3dwards:
LGTM, merging

PR #236, opened 2024-01-23
Support custom images on kubernetes

willcl-ark:
Support building and using custom images (i.e. github branches) on K8s.

Unfortunately I've ended up with another multi-faceted pull here. Needed various un-related changes to makes things work.

* Add `warcli image build` command to build an image from a repo, apply networking patches, and push it to a registry.
* Handle using custom images in the graph file
* Switch `scenarios list` -->  `scenarios available` (list is a very bad function name in python)
* Introduce RPC versioning. K8s rpc is pulled from an image too. Ideally this needs to be version-controlled/in sync with the CLI. This adds a `/api/<version>/` field to the endpoint, so we can at least co-ordinate breaking changes in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+![Python Version from PEP 621 TOML](https://img.shields.io/python/required-version-toml?tomlFilePath=https://raw.githubusercontent.com/bitcoin-dev-project/warnet/main/pyproject.toml)
 # Warnet
 
 Monitor and analyze the emergent behaviors of Bitcoin networks.
--- a/docs/developer-notes.md
+++ b/docs/developer-notes.md
@@ -1,0 +1,23 @@
+# Developer notes
+
+## Kubernetes
+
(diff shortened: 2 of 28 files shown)


willcl-ark:
@pinheadmz not sure if you'd like to review/test this on MacOS? I will try and do it myself sometime soon if not.

@josibake @m3dwards would also appreciate feedback on this if/when you have spare cycles (or are back from holibobs!), particularly the rpc API versioning stuff, as I've never really worked on things like that before, but found myself needing to make changes here and test them (I couldn't understand for ages why my stuff wasn't working on k8s; doh!).

Note also, that as the original k8s-rpc image was not tagged it was being pulled via `latest` (or indeed no tag specified). I accidentally over-wrote it by pushing a (any) tagged image.

m3dwards:
Tested building and running a custom image on k8s. Nice!

PR #238, opened 2024-01-30
k8s: refactor pod creation

willcl-ark:
This makes bitcoind a service, which means that lnd -- in it's own pod -- can communicate with it using default K8s cluster IP services.

Includes a slight refactor to the pod creation code too.

untested


m3dwards:
Picking this one up

willcl-ark:
@m3dwards @josibake  rebased on master and undrafted. Curious on your thoughts if it's the correct approach etc.

m3dwards:
Yep, I think this is the right approach.

PR #239, opened 2024-01-30
WIP: Added cp file RPC method

m3dwards:
Untested and a draft for now

### DIFF
--- a/src/warnet/server.py
+++ b/src/warnet/server.py
@@ -152,6 +152,19 @@ class Server:
             self.logger.error(msg)
             raise ServerError(message=msg) from e
 
+    def tank_cp_file(self, network: str, node: int, src) -> str:
+        """
+        Copy file from <node>
+        """
+        wn = Warnet.from_network(network, self.backend)
+        try:
+            result = wn.container_interface.get_bitcoin_debug_log(wn.tanks[node].container_name)
+            return str(result)
+        except Exception as e:
+            msg = f"Error copying file: {e}"
+            self.logger.error(msg)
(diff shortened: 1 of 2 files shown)


src/warnet/server.py
```diff
@@ -152,6 +152,19 @@ def tank_debug_log(self, network: str, node: int) -> str:
             self.logger.error(msg)
             raise ServerError(message=msg) from e
 
+    def tank_cp_file(self, network: str, node: int, src) -> str:
+        """
+        Copy file from <node>
+        """
+        wn = Warnet.from_network(network, self.backend)
+        try:
+            result = wn.container_interface.get_bitcoin_debug_log(wn.tanks[node].container_name)
```
willcl-ark, line 161: should this really be `get_bitcon_debug_log`?

m3dwards: perhaps not..

josibake:
Can you add a motivation to the description? IIRC, at one point we discussed copying files *to* the RPC server, but I can't remember why we needed to copy from containers

m3dwards:
Copying files was required for tests to pass and I just wanted a handy way to test copying files manually so started adding it to the CLI. I had a random stash of working towards that so just pushed it up to not lose it.

willcl-ark:
I think this should work much more nicely-er with the fixes in #277 (to `get_file()`)

josibake:
Circling back to this, with helm and the new commanders approach, do we still need this?

willcl-ark:
I think this can prob be closed. If anything we'll add another wrapper + picker directly around kubectl

PR #240, opened 2024-01-31
devtools: add devmode to server, cleanup template dir, add justfile

willcl-ark:
First we add a `--dev` mode to the server, this permits hot-reloads and faster dev iteration.

Next we update the dockerfile to use the new `--backend` flag (rather than using `argv[1]`).

Afterwards we can clean house in the template dir: move Dockerfiles that were loitering in the top level down. Archive some currently unused files. Followed by fixing up the dockerfile to be run from its new location (updating the self-copy of the source dir path).

Update the k8s rpc docker image to remove the builder when complete (TODO: also remove upon failure/early exit).

Next we add a new Dockerfile (`Dockerfile_rpc_dev`) and statefulset yaml file. With minikube these can be used to have hot server reloading inside of a k8s cluster, rather than needing to push a new image. This requires the following setup:

```bash
# Mount source dir into minikube
minikube mount "$PWD":/mnt/src  # requires window left open while mounted

# Apply statefulset, using dev config
kubectl apply -f src/templates/rpc/rbac-config.yaml -f src/templates/rpc/warnet-rpc-service.yaml -f src/templates/rpc/warnet-rpc-statefulset-dev.yaml
```

And teardown:

```bash
# teardown statefulset
kubectl delete statefulset rpc

# switch to minikube mount command and use <Ctrl>+C

```

Address this inconvenience by providing a `[just](https://github.com/casey/just)` file which automates this process. Now you can run

```bash
Setup dev mode with k8s
just kd
```

```bash
# Teardown dev mode with k8s
just kdd
```


m3dwards:
Pushed a few commits to:

Make it bash 5 compatible for us poor mac users
On down, rather than delete the rpc, replace it with production version
Added commands for docker desktop users

Haven't looked into the CI failures yet but happy to do so.

m3dwards:
Ci is green, merging.

PR #241, opened 2024-02-02
patch: fix isroutable patch to work properly

willcl-ark:
This makes addresses relay properly in k8s.

before:
image

after:
image

I will rebuild all the k8s images after this PR is merged...


willcl-ark:
Why not use a hammer

PR #242, opened 2024-02-02
justfile: correct down commands, fmt

willcl-ark:


willcl-ark:
@m3dwards Not sure how we ended up using `apply` instead of `delete` to bring down the statefulset, but now it will work

(also applied auto format to justfile with `just --fmt --unstable`)

willcl-ark:
Actually, I now see [this](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/#complete-deletion-of-a-statefulset), so may make further updates

m3dwards:
I changed it to an apply in this commit: https://github.com/bitcoin-dev-project/warnet/commit/4718c56e3d095069d0e32e68ba4a67bbe9740051

The apply sets it from the debug version to a production version. 

Thinking about it now though perhaps this doesn't make sense without a just command to bring up a production warnet.

Instead perhaps it should destroy all the resources created in the create stage.

justfile
```diff
@@ -1,85 +1,85 @@
...
+    set -euxo pipefail
+    echo Tearing down kubernetes dev mode
+
+    # Check we're in home dir
+    if [ ! -f "$PWD/pyproject.toml" ]; then \
+      echo "Error: must run from project root." >&2; exit 1; \
+    fi
+
+    # Stop statefulset
+    kubectl delete -f src/templates/rpc/warnet-rpc-statefulset.yaml
```
m3dwards: This isn't the debug statefulset yaml file. If you do want to delete, it should be that one. If they have the same name then this will probably still work but I think we should set it to the right file.

We should probably delete all of the resources, not just the statefulset.

m3dwards:
Pushed a more sane version

PR #243, opened 2024-02-02
WIP: ci: add rpc-k8 job

pinheadmz:


willcl-ark:
Looks like its failing because the server is not set up yet. Perhaps try rebasing on main now with the new ready check?

.github/workflows/test.yml
```diff
@@ -25,6 +25,19 @@ jobs:
...
       - uses: ./.github/actions
       - run: ./test/rpc_test.py
+  rpc-k8s:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: extractions/setup-just@v1
+      - uses: medyagh/setup-minikube@master
+      - run: pip install --upgrade pip
+      - run: pip install -e .
```
willcl-ark, line 35: I don't think this will do anything, as `just kd` is what mounts the (code) directories, so there will be nothing to install?

willcl-ark:
OK I give up. I think the latest push will work, but for some reason the CI has "gone away" (won't run). Perhaps we exhausted it?

Now we write an (empty) pidfile when we start the server. and the liveness check checks for that. It should be reliable.

willcl-ark:
Still think we must have exhausted free CI or something, CI is working on my [branch](https://github.com/willcl-ark/warnet/tree/tmp-ci-k8s):  https://github.com/willcl-ark/warnet/actions/runs/7767660281/job/21184732801

pinheadmz:
closing in favor of https://github.com/bitcoin-dev-project/warnet/pull/250

PR #248, opened 2024-02-05
Fix exec on k8s backend

rtreffer:
The official k8s stream method by default decodes the payload if it looks like JSON.
Upstream bug report: https://github.com/kubernetes-client/python/issues/1032

This _breaks_ `json.load` of basically all lightning and bitcoin cli outputs. 
E.g. `lncli newaddress p2wkh` would output (address mangled to avoid any copy/paste)
```
{
    "address": "___qmzz850w3t572tuwj9f68sc3jkku4py6qvpn___"
}
```
this would turn into
```
{
    'address': '___qmzz850w3t572tuwj9f68sc3jkku4py6qvpn___'
}
```
which fails on json.loads (single quotes vs. double quotes).

To avoid this we are disabling content parsing and manually let the [ws_client](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/stream/ws_client.py) run.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -149,6 +149,8 @@ class KubernetesBackend(BackendInterface):
             stdin=False,
             stdout=True,
             tty=False,
+            # Avoid preloading the content to keep JSON intact
+            _preload_content=False,
         )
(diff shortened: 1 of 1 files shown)


m3dwards:
Nice catch!

PR #250, opened 2024-02-05
rtreffer k8s fixes, pinheadmz ci fixes, and more...

willcl-ark:
Jam-packed with commits from what should, in all fairness, probably be about 5 PRs or more. Including #243 

Many cleanups from rtreffer, pinheadmz' ci work, and some of my own fixes and changes.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -377,6 +377,7 @@ class KubernetesBackend(BackendInterface):
     def create_lnd_container(self, tank, bitcoind_service_name) -> client.V1Container:
         # These args are appended to the Dockerfile `ENTRYPOINT ["lnd"]`
         bitcoind_rpc_host = f"{bitcoind_service_name}.{self.namespace}.svc.cluster.local"
+        lightning_dns = f"lightning-service-{tank.index}.{self.namespace}.svc.cluster.local"
         args = [
             "--noseedbackup",
             "--norest",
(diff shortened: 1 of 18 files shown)


src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -382,13 +446,20 @@ def create_pod_object(
         )
 
     def create_bitcoind_service(self, tank) -> client.V1Service:
-        service_name = f"bitcoind-service-{tank.index}"
+        service_name = f"warnet-{POD_PREFIX}-{tank.index:06d}"
```
m3dwards: Being as the lightning service is called "lightning-service-01" shouldn't we keep the bitcoind service named "bitcoind-service-01" too?

willcl-ark: Yeah I was actually thinking this too...

m3dwards (approved):
Nice PR! I think namespaces are long overdue.

Some great enhancements all round on this one. Let's merge before it grows any bigger as it already covers a few different concepts.

src/warnet/server.py
```diff
@@ -31,6 +32,7 @@
 SERVER_VERSION = "0.1"
 WARNET_SERVER_PORT = 9276
 CONFIG_DIR_ALREADY_EXISTS = 32001
+PID_FILE_PATH = "/tmp/warnet.pid"
```
pinheadmz: I feel like this should be in the warnet data directory and not hard coded

pinheadmz (approved):
ACK 5c38d9ba2abc80a59b5a9734687533003b10bb60

one question about pid file location but otherwise this is great work

src/warnet/server.py
willcl-ark: Thanks, this was a hangover from the previous pidfile liveness check. I thought it was cleaner (and more reliable) to add `pgrep` to the image and use that instead, so this can be removed. This is done in 7f432993402334575cfe6370e697890f8f6ba371

PR #254, opened 2024-02-06
Remove ip generator code

### LINKED ISSUE #189: Fix IP addresses

### LINKED ISSUE #253: tank IPs are still 100.0.0.0/8 in K8s
these IPs are not accurate:

```
--> warcli network start --force
Warnet:
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Temp dir                    ┃ Bitcoin network ┃ Docker network ┃ Subnet      ┃ Graph                              ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ /root/.warnet/warnet/warnet │ regtest         │ warnet         │ 100.0.0.0/8 │ DiGraph with 12 nodes and 11 edges │
└─────────────────────────────┴─────────────────┴────────────────┴─────────────┴────────────────────────────────────┘

Tanks:
┏━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ Index ┃ Version ┃ IPv4            ┃ bitcoin conf   ┃ tc_netem ┃
┡━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━┩
│ 0     │ 26.0    │ 100.157.38.141  │ -uacomment=w0  │ None     │
│ 1     │ 26.0    │ 100.240.177.228 │ -uacomment=w1  │ None     │
│ 2     │ 25.1    │ 100.3.110.4     │ -uacomment=w2  │ None     │
│ 3     │ 26.0    │ 100.170.225.26  │ -uacomment=w3  │ None     │
│ 4     │ 26.0    │ 100.213.234.13  │ -uacomment=w4  │ None     │
│ 5     │ 26.0    │ 100.183.221.239 │ -uacomment=w5  │ None     │
│ 6     │ 26.0    │ 100.43.160.138  │ -uacomment=w6  │ None     │
│ 7     │ 26.0    │ 100.17.43.236   │ -uacomment=w7  │ None     │
│ 8     │ 26.0    │ 100.192.174.3   │ -uacomment=w8  │ None     │
│ 9     │ 26.0    │ 100.49.199.126  │ -uacomment=w9  │ None     │
│ 10    │ 26.0    │ 100.135.210.163 │ -uacomment=w10 │ None     │
│ 11    │ 26.0    │ 100.208.150.101 │ -uacomment=w11 │ None     │
└───────┴─────────┴─────────────────┴────────────────┴──────────┘
```

willcl-ark:
Closes: #253
Closes: #189

* Remove the ip generator stuff
* Add helper methods to fetch ip address of a tank from the backend directly

### DIFF
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -107,3 +107,10 @@ class BackendInterface(ABC):
         Rebuild a warnet object from an active deployment
         """
         raise NotImplementedError("This method should be overridden by child class")
+
+    @abstractmethod
+    def fetch_ip_address(self, tank) -> bool:
+        """
+        Populate the tank ipv4 address
+        """
+        raise NotImplementedError("This method should be overridden by child class")
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -114,3 +114,11 @@ class BackendInterface(ABC):
         Populate the tank ipv4 address
         """
         raise NotImplementedError("This method should be overridden by child class")
+
+    @abstractmethod
+    def get_container_name(self, tank_index: int, service: ServiceType) -> str:
+        """
+        Fetch a container name
+        """
+        raise NotImplementedError("This method should be overridden by child class")
+
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -50,6 +50,7 @@ class ComposeBackend(BackendInterface):
         super().__init__(config_dir)
         self.network_name = network_name
         self.client = docker.DockerClient = docker.from_env()
+        self.running_ips: dict | None = None
 
     def build(self) -> bool:
         command = ["docker", "compose", "build"]
@@ -139,6 +140,8 @@ class ComposeBackend(BackendInterface):
 
     def exec_run( self, tank_index: int, service: ServiceType, cmd: str) -> str:
         c = self.get_container(tank_index, service)
+        if not c:
+            return f"Could not get_container {tank_index=:} {service=:}"
         result = c.exec_run(cmd=cmd)
         if result.exit_code != 0:
             raise Exception(
@@ -174,6 +177,8 @@ class ComposeBackend(BackendInterface):
 
     def get_file(self, tank_index: int, service: ServiceType, file_path: str):
         container = self.get_container(tank_index, service)
+        if not container:
+            return f"Could not get_container {tank_index=:} {service=:}"
         data, stat = container.get_archive(file_path)
         out = b""
         for chunk in data:
@@ -280,7 +285,6 @@ class ComposeBackend(BackendInterface):
             "networks": {
                 warnet.network_name: {
                     "name": warnet.network_name,
-                    "ipam": {"config": [{"subnet": warnet.subnet}]},
                 }
             },
             "volumes": {"grafana-storage": None},
@@ -378,11 +382,7 @@ class ComposeBackend(BackendInterface):
                 # logging with json-file to support log shipping with promtail into loki
                 "logging": {"driver": "json-file", "options": {"max-size": "10m"}},
                 "environment": {"BITCOIN_ARGS": self.config_args(tank)},
-                "networks": {
-                    tank.network_name: {
-                        "ipv4_address": f"{tank.ipv4}",
-                    }
-                },
+                "networks": [tank.network_name],
                 "labels": {"warnet": "tank"},
                 "privileged": True,
                 "cap_add": ["NET_ADMIN", "NET_RAW"],
@@ -400,7 +400,7 @@ class ComposeBackend(BackendInterface):
             services[container_name]["labels"].update({"collect_logs": True})
 
         if tank.lnnode is not None:
-            self.add_lnd_service(tank, services)
+            self.add_lnd_service(tank, services, container_name)
 
         # Add the prometheus data exporter in a neighboring container
         if tank.exporter:
@@ -408,7 +408,7 @@ class ComposeBackend(BackendInterface):
                 "image": "jvstein/bitcoin-prometheus-exporter:latest",
                 "container_name": tank.exporter_name,
                 "environment": {
-                    "BITCOIN_RPC_HOST": tank.ipv4,
+                    "BITCOIN_RPC_HOST": container_name,
                     "BITCOIN_RPC_PORT": tank.rpc_port,
                     "BITCOIN_RPC_USER": tank.rpc_user,
                     "BITCOIN_RPC_PASSWORD": tank.rpc_password,
@@ -416,7 +416,7 @@ class ComposeBackend(BackendInterface):
                 "networks": [tank.network_name],
             }
 
-    def add_lnd_service(self, tank, services):
+    def add_lnd_service(self, tank, services, bitcoind_container_name):
         ln_container_name = self.get_container_name(tank.index, ServiceType.LIGHTNING)
         bitcoin_container_name = self.get_container_name(tank.index, ServiceType.BITCOIN)
         # These args are appended to the Dockerfile `ENTRYPOINT ["lnd"]`
@@ -430,10 +430,10 @@ class ComposeBackend(BackendInterface):
             "--bitcoin.node=bitcoind",
             f"--bitcoind.rpcuser={tank.rpc_user}",
             f"--bitcoind.rpcpass={tank.rpc_password}",
-            f"--bitcoind.rpchost={tank.ipv4}:{tank.rpc_port}",
-            f"--bitcoind.zmqpubrawblock=tcp://{tank.ipv4}:{tank.zmqblockport}",
-            f"--bitcoind.zmqpubrawtx=tcp://{tank.ipv4}:{tank.zmqtxport}",
-            f"--externalip={tank.lnnode.ipv4}",
+            f"--bitcoind.rpchost={bitcoind_container_name}:{tank.rpc_port}",
+            f"--bitcoind.zmqpubrawblock=tcp://{bitcoind_container_name}:{tank.zmqblockport}",
+            f"--bitcoind.zmqpubrawtx=tcp://{bitcoind_container_name}:{tank.zmqtxport}",
+            # f"--externalip={tank.lnnode.ipv4}",
             f"--rpclisten=0.0.0.0:{tank.lnnode.rpc_port}",
             f"--alias={tank.index}",
         ]
@@ -441,15 +441,10 @@ class ComposeBackend(BackendInterface):
             "container_name": ln_container_name,
             "image": "lightninglabs/lnd:v0.17.0-beta",
             "command": " ".join(args),
-            "networks": {
-                tank.network_name: {
-                    "ipv4_address": f"{tank.lnnode.ipv4}",
-                }
-            },
+            "networks": [tank.network_name],
             "labels": {
                 "tank_index": tank.index,
                 "tank_container_name": bitcoin_container_name,
-                "tank_ipv4_address": tank.ipv4,
             },
             "depends_on": {bitcoin_container_name: {"condition": "service_healthy"}},
             "restart": "on-failure",
@@ -458,7 +453,6 @@ class ComposeBackend(BackendInterface):
             {
                 "labels": {
                     "lnnode_container_name": ln_container_name,
-                    "lnnode_ipv4_address": tank.lnnode.ipv4,
                     "lnnode_impl": tank.lnnode.impl,
                 }
             }
@@ -488,7 +482,6 @@ class ComposeBackend(BackendInterface):
 
         index = int(match.group(1))
         tank = Tank(index, warnet.config_dir, warnet)
-        tank._ipv4 = service["networks"][tank.network_name]["ipv4_address"]
         if "image" in service:
             tank.version = service["image"].split(":")[1]
         else:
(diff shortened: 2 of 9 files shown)


src/backends/backend_interface.py
```diff
@@ -107,3 +107,10 @@ def warnet_from_deployment(self, warnet):
         Rebuild a warnet object from an active deployment
         """
         raise NotImplementedError("This method should be overridden by child class")
+
+    @abstractmethod
+    def fetch_ip_addresses(self, tank) -> bool:
```
josibake: in "backend: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/pull/254/commits/50107b7f1dd88a1f0d393ed0e64b2ef8b19fddc7)

nit: shouldn't this be `fetch_ip_address`, since it only operates on a single tank? also not sure if this matters in python if the abstractmethod return type differs from the implementation, but if we're returning a bool, maybe we should call this `set_ip_address`?

src/backends/compose/compose_backend.py
```diff
@@ -82,10 +82,12 @@ def up(self, *_):
                 if process.stdout:
                     for line in process.stdout:
                         logger.info(line.decode().rstrip())
+            return True
         except Exception as e:
             logger.error(
                 f"An error occurred while executing `{' '.join(command)}` in {self.config_dir}: {e}"
             )
+            return False
```
josibake: in "backend: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/commit/50107b7f1dd88a1f0d393ed0e64b2ef8b19fddc7):

seems unrelated? (probably fine to keep it, just making sure im not missing something :sweat_smile: )

src/backends/compose/compose_backend.py
```diff
@@ -278,7 +286,6 @@ def _write_docker_compose(self, warnet):
             "networks": {
                 warnet.network_name: {
                     "name": warnet.network_name,
-                    "ipam": {"config": [{"subnet": warnet.subnet}]},
```
josibake, line 283: in "backend: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/commit/50107b7f1dd88a1f0d393ed0e64b2ef8b19fddc7):

I'm not sure about this.. IIRC, with out patch bitcoin core will see non-routable addresses as routable (great!). But if the addresses are still in too small of a subnet, the bucketing logic wont kick in. For a compose backend, I think we might still want to set an ipam network so we can say ip addresses should be assigned in the `172.*` ranges, right?

On the other hand, if we are going to fully handle that with ASMAP (for docker and k8s backends), then yes we should nuke the network part.

src/backends/compose/compose_backend.py
```diff
@@ -436,26 +439,21 @@ def add_lnd_service(self, tank, services):
             "--bitcoin.node=bitcoind",
             f"--bitcoind.rpcuser={tank.rpc_user}",
             f"--bitcoind.rpcpass={tank.rpc_password}",
-            f"--bitcoind.rpchost={tank.ipv4}:{tank.rpc_port}",
-            f"--bitcoind.zmqpubrawblock=tcp://{tank.ipv4}:{tank.zmqblockport}",
-            f"--bitcoind.zmqpubrawtx=tcp://{tank.ipv4}:{tank.zmqtxport}",
-            f"--externalip={tank.lnnode.ipv4}",
+            f"--bitcoind.rpchost={bitcoind_container_name}:{tank.rpc_port}",
+            f"--bitcoind.zmqpubrawblock=tcp://{bitcoind_container_name}:{tank.zmqblockport}",
+            f"--bitcoind.zmqpubrawtx=tcp://{bitcoind_container_name}:{tank.zmqtxport}",
```
josibake, line 435: in "backend: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/commit/50107b7f1dd88a1f0d393ed0e64b2ef8b19fddc7):

This works because every bitcoind node is behind a service, right?

src/warnet/warnet.py
```diff
@@ -220,3 +220,10 @@ def export(self, subdir):
         config_path = os.path.join(subdir, "sim.json")
         with open(config_path, "a") as f:
             json.dump(config, f)
+
+    def populate_ip_addresses(self):
+        while not all(tank.ipv4 is not None for tank in self.tanks):
```
josibake, line 225: in "backend: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/commit/50107b7f1dd88a1f0d393ed0e64b2ef8b19fddc7):

oof.. this is gonna be fun with 500 nodes. but not sure if there's a cleaner way

src/warnet/server.py
```diff
@@ -322,10 +322,8 @@ def network_up(self, network: str = "warnet") -> str:
         def thread_start(wn):
             try:
                 wn.container_interface.up()
-                # Update warnet from docker here to get ip addresses
-                wn = Warnet.from_network(network, self.backend)
-                wn.populate_ip_addresses()
                 wn.apply_network_conditions()
+                wn.fetch_ip_addresses()
```
josibake: in "remove ip generator" (https://github.com/bitcoin-dev-project/warnet/pull/254/commits/e4099d7b297f5d320e22875f01cba72809b4d839):

seems like these changes could be squashed into the last commit?

josibake (commented):
Kill it with fire! Left a few nits but seems good. I need to do my homework, but I see we have a `ServiceType.[BITCOIN|LIGHTNING]`. Does this mean lighting and bitcoin are both running behind the same service but in different containers, and would then have the same IP address but different ports? Or are they two separate services, but belong to the same "tank" ?

src/backends/backend_interface.py
willcl-ark: Agree

src/backends/compose/compose_backend.py
willcl-ark: It is, drive-by fixes not helping anything much

src/backends/compose/compose_backend.py
willcl-ark: Damn you are right about the add bucketing... That is a pain. Even if I rework the asmap stuff the only way to have it loaded and effective is to know the ip addresses beforehand

src/warnet/warnet.py
willcl-ark: I think I can make the loop cleaner.

Also, we should be able to fetch an entire `inspect network` (or equivalent), and parse that in a oner

src/backends/compose/compose_backend.py
josibake: coupla ideas:

1. leave the addrman bucketing alone for now until we sort out some of the other stuff and at least get the k8s backend completely built out
2. leave this ipam network bit in for now so we can use it with the docker backend. this means addrman will work for docker with patch + defining a subnet
3. get the ASMAP stuff working where it runs after the ip addresses have been assigned (same for docker and k8s)

prefer 1) but technically thats a regression for the docker backend. 2 seems like not too much work, and 3 is gross as I feel we'll have too many moving parts. WDYT?

src/warnet/warnet.py
josibake: dunno about docker, but i think the k8s api gives us some nice endpoints that should give everything we need in one response

willcl-ark:
Drafting while we #261

willcl-ark:
Rebased on top of #261 @ 66444ff for CI check

src/backends/backend_interface.py
willcl-ark: done

src/backends/compose/compose_backend.py
willcl-ark: removed

src/backends/compose/compose_backend.py
willcl-ark: revolved in #261

src/backends/compose/compose_backend.py
willcl-ark: Correct!

willcl-ark:
This needs the json decorator from #270, so I've pulled those commits in here for CI purposes

src/warnet/warnet.py
willcl-ark: On the compose side I at least cache the result of the first lookup, which seems to work.

https://github.com/bitcoin-dev-project/warnet/pull/254/files#diff-adeb9c7bdada05ccdf2c802012d7c32c0737f526b7ed1d34443e7678cb81236aR499-R516

src/backends/compose/compose_backend.py
```diff
@@ -500,4 +493,37 @@ def tank_from_deployment(self, service, warnet):
         if "lnnode_impl" in labels:
             tank.lnnode = LNNode(warnet, tank, labels["lnnode_impl"], self)
             tank.lnnode.ipv4 = labels.get("lnnode_ipv4_address")
+        self.fetch_ip_address(tank)
         return tank
+
+    def _get_running_containers_ips(self):
+        self.running_ips = {
```
josibake, line 500: in "compose: remove ipv4 from compose backend" (https://github.com/bitcoin-dev-project/warnet/pull/254/commits/a80917d332bad84d6b5b896babbbbcc4b51e95b5):

hm, caching scares me a lil in that a container could die and restart with a different IP, which im guessing invalidates the cache? i dont think this is a big issue for docker, since the containers tend to fall over less, but could lead to some weird/hard to troubleshoot problems if running docker with not enough resources / too many containers.

src/warnet/server.py
```diff
@@ -376,13 +389,15 @@ def network_from_file(
...
 
-        def thread_start(wn, lock: threading.Lock):
-            with lock:
+        def thread_start(server: Server, network):
+            with server.image_build_lock:
                 try:
+                    wn = server.get_warnet(network)
                     wn.generate_deployment()
-                    # wn.write_fork_observer_config()
+                    wn.write_fork_observer_config()
```
josibake, line 397: in "server: optimistic warnet cache on server" (https://github.com/bitcoin-dev-project/warnet/pull/254/commits/be145a400d0a4ed43685558d005a1f93eb1f8937):

did you mean to add fork observer back in here? this whole commit seems to be unrelated to the IP changes, might be safer to drop it for now so we can merge just getting rid of the IP code?

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -545,6 +545,20 @@ def create_lightning_service(self, tank) -> client.V1Service:
...
         return service
 
+    def fetch_ip_address(self, tank) -> bool:
+        pod_ip = None
+        while not pod_ip:
+            pod_name = self.get_pod_name(tank.index, ServiceType.BITCOIN)
+            pod = self.get_pod(pod_name)
+            if pod is None or pod.status is None or getattr(pod.status, "pod_ip", None) is None:
+                print(f"Waiting for tank {tank.index} response or tank IP...")
+                time.sleep(3)
```
josibake, line 559: in "k8s: remove ipv4 from k8s backend" (https://github.com/bitcoin-dev-project/warnet/pull/254/commits/1a3223d0accd8398954f36d402e17844b33be8db):

I don't have a good sense of how likely we are to hit this, but if we hit it on every single pod while spinning up a 500 node network, that would be `(3 sec * 500) / 60 sec = 25 mins`. maybe not a blocker for now, but wanted to mention it as i think this approach will need reworking for k8s.

josibake (commented):
LGTM, left a few comments. I'm not quite sure what's going on with the optimistic cache commit, perhaps we drop that one for now and merge this as is?

src/backends/compose/compose_backend.py
willcl-ark: mmmmm, yeah. The previous behaviour is that we must look up the ip address on (almost) every server call... Perhaps I can refactor this to look it up dynamically each time, but only when it's relevant.

Will investigate...

src/warnet/server.py
willcl-ark: Yeah I see why it's not obvious, sorry my bad.

Current behaviour is that we rebuild the warnet from deployment file on *every* command run against the server. This doesn't work when we remove the hardcoded IP addresses because we are now retrieving them from the backend. This meant that every single server RPC would cause lookup of every ip from the backend on every call. Yuck.

To combat this I attempted the warnet caching, alongside the ip address caching in the previous comment. I think these are still a bit naive though, as you point out. I am trying to work out how we can totally switch to hostnames, which I think should be doable in most cases, and then we will neither need to cache or lookup ips all the time...

willcl-ark: It actually looks like it could work quite nicely by only modifying lnnode.py ... which would be nice

PR #255, opened 2024-02-06
patch: addrman buckets ipv4 based on /32

willcl-ark:
This adds a(nother) new dockerfile for building small alpine linux images, with both isroutable and addrman bucketing fixes specific to warnet.

If all goes well, we should be able to use this dockerfile for building our default (tagged) images, and to use in the custom branch building code.

The script should also be cross-purpose for building bitcoind Docker images and warnet-rpc images.

This will enable us to remove from the codebase:

```bash
src/templates/Dockerfile
src/templates/Dockerfile_k8
scripts/build-bitcoin-images-k8s.sh
scripts/ build-bitcoin-images.sh
scripts/build-k8s-rpc.sh
```

and rework all builds to use a single script, and both docker and k8s to use the unified image

### DIFF
--- a/scripts/build-bitcoind.sh
+++ b/scripts/build-bitcoind.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+# Run from the src/templates dir with e.g.:
+#
+# $ DOCKER_REGISTRY=bitcoindevproject/demo TAG=26.0 ../../scripts/build-k8s-rpc.sh Dockerfile_rpc_alpine
+
+
+BUILDER_NAME="warnet-bitcoind-builder"
+# Cleanup function to remove the Docker builder
+cleanup() {
+  echo "Removing builder..."
+  docker buildx rm "$BUILDER_NAME" || echo "Failed to remove builder $BUILDER_NAME"
+}
+
+# Set up trap to call cleanup function on EXIT signal
+trap cleanup EXIT SIGINT SIGTERM
+
+set -ex
+
+# Create a new builder to enable building multi-platform images
+docker buildx create --name "$BUILDER_NAME" --use
+
+# Read DOCKER_REGISTRY from the environment
+: "${DOCKER_REGISTRY?Need to set DOCKER_REGISTRY}"
+: "${TAG?Need to set TAG}"
+
+# Architectures for building
+ARCHS="linux/amd64,linux/arm64"
+
+# Read Dockerfile from the first argument
+DOCKERFILE_PATH=$1
+if [[ ! -f "$DOCKERFILE_PATH" ]]; then
+  echo "Dockerfile does not exist at the specified path: $DOCKERFILE_PATH"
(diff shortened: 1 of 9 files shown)


src/templates/addrman.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -43,11 +43,7 @@ std::vector<unsigned char> NetGroupManager::GetGroup(const CNetAddr& address) co
+     } else if (!address.IsRoutable()) {
+         // all other unroutable addresses belong to the same group
+     } else if (address.HasLinkedIPv4()) {
+-        // IPv4 addresses (and mapped IPv4 addresses) use /16 groups
+-        uint32_t ipv4 = address.GetLinkedIPv4();
+-        vchRet.push_back((ipv4 >> 24) & 0xFF);
+-        vchRet.push_back((ipv4 >> 16) & 0xFF);
+-        return vchRet;
++        nBits = 32;
```
josibake, line 14: note for posterity's sake: if this patch becomes unmaintainable, we could propose a refactor in core to have these "magic" numbers be consts defined in `src/addrman.cpp` with nice comments explaining what they do. that means our patch would work for any old versions and our new patch would be as simple as changing a constant.

scripts/build-bitcoind.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Run from the src/templates dir with e.g.:
+#
+# $ DOCKER_REGISTRY=bitcoindevproject/demo TAG=26.0 ../../scripts/build-k8s-rpc.sh Dockerfile_rpc_alpine
+
+set -ex
+
+# Create a new builder to enable building multi-platform images
+BUILDER_NAME="warnet-rpc-builder"
```
josibake: in "images: generic builder" (https://github.com/bitcoin-dev-project/warnet/pull/255/commits/7c818eceeb56d328906f63112f13a523e4349f64):

seems incorrect? but afaict doesn't cause any problems since it is only used by the builder

test/data/v25_x_12.graphml
```diff
@@ -4,7 +4,7 @@
   <key attr.name="tc_netem" attr.type="string" for="node" id="tc_netem"/>
   <graph edgedefault="directed">
     <node id="0">
-        <data key="version">26.0</data>
+        <data key="image">bitcoindevproject/demo:26.0</data>
```
josibake, line 8: this is causing the CI failure, fixed in #256

josibake (commented):
LGTM! rebasing on #256 should fix the CI failures

josibake:
Rebased on #256

josibake:
ACK https://github.com/bitcoin-dev-project/warnet/pull/255/commits/e766f7947fbc52f583ec447b1df64bfae2109427

willcl-ark:
I think Ci still broken with the alpine image for now...

PR #257, opened 2024-02-06
ci: use matrix for backends

pinheadmz:
Cleans up `test.yml`: for each test file we have one job that uses a matrix to run that test in both backends. The compose option applies the the docker installation in `actions/action.yml` and the k8s option installs minikube with other github action layers. The test command is run explicitly with the backend option passed as an argument.

I also cut down the timeouts on LN and Build tests, since they don't pass on k8s and would otherwise hang forever.

Finally, I fixed the `graph_test.py` test for k8s. The `graph create` command already returns the graphml file as a big string so in the k8s mode the client will have to write that string to a file, then run `network start` from that file, which sends it back  to the server as a base64 string.

I feel like we should merge this even with the two failing k8s tests. @rtreffer is already working on fixing LN in k8s, and can then rebase his work.


josibake:
CI seems to be failing because it can’t find the generated graph file:

```
-32000: Error generating graph: [Errno 2] No such file or directory: '/tmp/tmpparntl5k/4b4ebef8-9837-48ca-be86-a808f51f90b1.graphml'
```

The predicate truth error is a bit misleading, has to scroll way up to find the actual failure

willcl-ark:
LGTM, nice!

I actually am going to advocate for removing the branch_build test in the very near future, and the related building code. 

But I will do it elsewhere :P

rtreffer:
LN on k8s passing: #259

PR #259, opened 2024-02-06
Wait for lightning balances to be distributed

rtreffer:
Otherwise we might try to open channels from nodes that can't spend any fund yet.

[EDIT] rebased on main after #257, ln on k8s is passing.


willcl-ark:
Excellent, thanks!

PR #261, opened 2024-02-07
patch addrman bucketing, remove build

### LINKED ISSUE #212: Use networking patch in docker too
We should use the same josie-patched version of Core in both docker and k8s:

Advantages:

* single image / registry
* single image building pipeline
* can be built with guix:
  * smaller image sizes
  * no libs needed
  * improved performance (lower ram usage)

Disadvantages:

* no downsides?
* prevents a "firewalled" setup (one node with a single LAN connection to a "firewall" public-facing node, add unsupported feature to docs?

willcl-ark:
Closes: #212 

Yay, another PR which should be 3!

This turned out longer than I wanted, as after patching addrman bucketing I needed to update the images. These are currently based on a plethora of dockerfiles created using various scripts, with different versions for k8s and compose. By using the same image for both backends we can in the future drop our ipv4 generator code from compose and just have everything routable and bucketable on both. We also only need to rebuild half the number of images...

In undertaking this, I thought we could take the opportunity to switch to alpine-based images, as they're (much) smaller. This neccesitated some minor changes to `exec` methods in the backends as there is no `gosu` in alpine (but there is in LND pods, debian-based?)

Finally, the knock-on effect of this was breakage of the `build` test. In my opinion this should be removed altogether and we should require users to simply run:

```bash
warcli image build --registry=<my_registry>/<name> --repo=<gh_user>/<repo> --branch=<branch_name> --arches=amd64 --tag=<my_tag>
```

...and then supply image info to the graph. This is a bit more manual than auto-building, but it's not clear currently how that would work with k8s in any case. This ends up therefore removing compose-specific code only, and makes a few code sections much simpler. We could simplify even further by removing `version` and only permitting image, using our registry as the default. But I've left that for now.

In order to proceed with this change, if folks want things past the patch, we should first update https://hub.docker.com/r/bitcoindevproject/bitcoin-core with new unified (and patched) images, then remove the NO MERGE commit(s), both if we  want to keep image building on compose, otherwise only the penultimate.

We can then remove the following registries too for extra cleanliness:

https://hub.docker.com/r/bitcoindevproject/k8s-bitcoin-core
https://hub.docker.com/r/bitcoindevproject/warnet-bitcoin
https://hub.docker.com/r/bitcoindevproject/warnet-bitcoin-core


willcl-ark:
Also worth considering: I cannot claim to fully understand the ramifications of this addrman patch. I've observed different buckets and positions being used rather than a single bucket, so I know that much works. However it's unclear to me yet whether there are any currently-unforseen consequences of bucketing by /32...

It occurred to me that we may have unwittingly altered the total number of buckets, but there may be other side-effects too.

josibake:
> Also worth considering: I cannot claim to fully understand the ramifications of this addrman patch. I've observed different buckets and positions being used rather than a single bucket, so I know that much works. However it's unclear to me yet whether there are any currently-unforseen consequences of bucketing by /32...

Agree, and I think we should clearly document this (at least for now). I still think this is the best route forward, but we should run a live warnet scenario for awhile and compare the raw addrman against some real nodes to get a sense of how closely the change mirrors mainnet. If this ends up introducing side effects, we'll need to go back to the drawing board, but for now this seems to unblock us.

willcl-ark:
FWIW, some image size comparisons:

Release image, current, debian-slim: 159MB
Built image, current, debian-slim: > 300MB!
(estimate built could be reduced to about 175MB but not tested)

Unified, proposed, alpine: 35.3MB

src/backends/compose/compose_backend.py
```diff
@@ -135,8 +135,10 @@ def get_status(self, tank_index: int, service: ServiceType) -> RunningStatus:
             case _:
                 return RunningStatus.PENDING
 
-    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "root") -> str:
+    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user="bitcoin") -> str:
```
josibake: in "images: unified alpine-based dockerfile" (https://github.com/bitcoin-dev-project/warnet/pull/261/commits/4dc71ff74d4267b9c3f45530c1490702bd81c188):

```suggestion
    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "bitcoin") -> str:
```

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -157,13 +157,12 @@ def get_status(self, tank_index: int, service: ServiceType) -> RunningStatus:
                 break
         return RunningStatus.RUNNING if ready else RunningStatus.PENDING
 
-    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "root"):
-        # k8s doesn't let us run exec commands as a user, but we can use su
-        # because its installed in the bitcoin containers. we will need to rework
-        # this command if we decided to remove gosu from the containers
-        # TODO: change this if we remove gosu
+    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user="root"):
```
josibake: in "images: unified alpine-based dockerfile" (https://github.com/bitcoin-dev-project/warnet/commit/4dc71ff74d4267b9c3f45530c1490702bd81c188):

```suggestion
    def exec_run(self, tank_index: int, service: ServiceType, cmd: str, user: str = "root"):
```

src/utils/image_build.py
```diff
@@ -51,10 +47,9 @@ def build_and_upload_images(
         print("Directory src/templates does not exist.")
         print("Please run this script from the project root.")
         return False
-    os.chdir("src/templates")
 
     # Setup buildkit
-    builder_name = "warnet-builder"
+    builder_name = "bitcoind-builder"
```
josibake, line 53: in "fixup! images: unified alpine-based dockerfile" (https://github.com/bitcoin-dev-project/warnet/pull/261/commits/6070ea73f07dbccd0e08a1baec44ebe974f94e89):

in the previous commit this was called `warnet-bitcoind-builder`, not sure if that makes a difference. Also noticed you're removing the script in this commit, might be worth just dropping the script entirely from the PR?

josibake:
> Finally, the knock-on effect of this was breakage of the build test. In my opinion this should be removed altogether and we should require users to simply run: <script>

I do agree this should be the normal mode of operation, but I don't think we should disable the build script altogether, in that I think it's still useful for CI. Needing to push images first to a registry before the CI can pass seems error prone and a PITA. AFAIK, if we are using minikube in the CI, we can build the images and use them directly in minikube, so maybe we keep the building and CI stuff to where it doesn't rely on hitting the registry? Or both: one that builds and tests the images, and another that runs with the prebuilt images, more as a regression test?

There may also be users who want to use warnet locally with docker or minikube that would benefit from the build script for local work.

src/backends/compose/compose_backend.py
willcl-ark: taken

src/backends/kubernetes/kubernetes_backend.py
willcl-ark: taken

src/utils/image_build.py
willcl-ark: Which script do you mean? This PR attempts to clean a load of them out already?

willcl-ark:
> I do agree this should be the normal mode of operation, but I don't think we should disable the build script altogether, in that I think it's still useful for CI. Needing to push images first to a registry before the CI can pass seems error prone and a PITA.
> AFAIK, if we are using minikube in the CI, we can build the images and use them directly in minikube, so maybe we keep the building and CI stuff to where it doesn't rely on hitting the registry? Or both: one that builds and tests the images, and another that runs with the prebuilt images, more as a regression test?

Hmmm, I somewhat agree. let me see if the script can be modified to use `--load`, then we can use it to build and load into local registry, before using in minikube on CI. Then I'd argue that's equivalent to what we have now, and the rest can still go?


> There may also be users who want to use warnet locally with docker or minikube that would benefit from the build script for local work.

I think this would be addressed by the above too

src/utils/image_build.py
josibake: I could have sworn you were adding `build-bitcoind.sh` in an earlier commit and then removing it in this one.. but I don't see it now? Either I'm losing my marbles or you already removed it haha

josibake:
> Hmmm, I somewhat agree. let me see if the script can be modified to use --load, then we can use it to build and load into local registry, before using in minikube on CI. Then I'd argue that's equivalent to what we have now, and the rest can still go?

Sounds excellent!

m3dwards (approved):

willcl-ark:
Sorry for many rebases, somehow broke the scenarios test during the work...

willcl-ark:
OK, split the big commits for easier review, and got all the tests passing again.

It sounds like we might now be waiting on @pinheadmz magic formula before considering merging this? Alternatively, we could merge this and simply update the (single) docker images, removing the patches and swapping with his (non-bitcoin) patches, if he gets it working.

> AFAIK, if we are using minikube in the CI, we can build the images and use them directly in minikube, so maybe we keep the building and CI stuff to where it doesn't rely on hitting the registry?

For now, I've disabled branch-building on k8s backend _in CI only_. It might work on your local machine though...

willcl-ark:
It also just occurred to me that with a unified image we could have a workflow to (re)build the docker images when we tag in this repo (or whatever)

src/utils/image_build.py
```diff
@@ -21,11 +21,7 @@ def build_and_upload_images(
     arches: str,
 ):
     if not build_args:
-        build_args = (
-            '"--disable-tests --with-incompatible-bdb --without-gui --disable-bench '
-            "--disable-fuzz-binary --enable-suppress-external-warnings "
-            '--without-miniupnpc --without-natpmp"'
```
willcl-ark, line 27: Don't need these as we don't install the deps for them in the image, so they don't get build automatically.

src/templates/entrypoint.sh
```diff
@@ -16,20 +16,10 @@ if [ "${TOR:-0}" -eq 1 ]; then
     mkdir -p /home/debian-tor/.tor/keys
     chown -R debian-tor:debian-tor /home/debian-tor
     # Start tor in the background
-    gosu debian-tor tor &
+    su-exec debian-tor:debian-tor tor &
```
josibake, line 19: do we still need su-exec, now that we are always running as the root user?

src/templates/Dockerfile_unified_alpine
```diff
@@ -0,0 +1,86 @@
...
+RUN addgroup bitcoin --gid ${GID} --system \
+    && adduser --uid ${UID} --system bitcoin --ingroup bitcoin
+RUN --mount=type=cache,target=/var/cache/apk sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/https\:\/\/alpine.global.ssl.fastly.net/g' /etc/apk/repositories \
+    && apk --no-cache add \
+    bash \
+    libevent \
+    libzmq \
+    shadow \
+    sqlite-dev \
+    su-exec
```
josibake: can remove, yeah?

src/utils/image_build.py
```diff
@@ -66,25 +61,27 @@ def build_and_upload_images(
...
         f"docker buildx build"
         f" --platform {platforms}"
-        f" --provenance=false"
         f" --build-arg REPO={repo}"
         f" --build-arg BRANCH={branch}"
         f" --build-arg BUILD_ARGS={build_args}"
         f" --tag {image_full_name}"
-        f" --file Dockerfile_k8 ."
+        f" --file src/templates/Dockerfile ."
         f" --push"
```
josibake: would be nicer if you added the {action} arg in this commit and then updated the justfile in the next, but not a big deal, was just very confused by the next commit haha

src/cli/image.py
```diff
@@ -17,11 +17,12 @@ def image():
...
 @click.option("--build-args", required=False, type=str)
 @click.option("--arches", required=False, type=str)
-def build(repo, branch, registry, tag, build_args, arches):
+@click.option("--action", required=False, type=str)
+def build(repo, branch, registry, tag, build_args, arches, action="load"):
     """
-    Build bitcoind and bitcoin-cli from <repo>/<branch> and deploy to <registry> as <tag>
-    This requires docker and buildkit to be enabled.
+    Build bitcoind and bitcoin-cli from <repo>/<branch> as <registry>:<tag>.
+    Optionally deploy to remote registry using --action=push, otherwise image is loaded to local registry.
```
josibake, line 24: note for a follow up , would be nice if we gave examples in our docs of how to use this script, just to make it more clear

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -18,6 +18,7 @@
```
josibake, line 18: tbh im not sure this will work on k8s. generally in k8s, the pattern for building images is you have a custom image with the correct permissions (Kaniko is the most popular one), so you pass of a build context to a container running in the cluster, and that container returns the image. this has some additional nice properties of allowing you to build multiple images in parallel (i think?). either way, i think this will take some more finessing to get working.

josibake (commented):
looks good, left a few comments

src/backends/kubernetes/kubernetes_backend.py
willcl-ark: I think this will just have the RPC container build the image no?

But I can't see what exactly you're commenting on here (thank GH)

src/utils/image_build.py
willcl-ark: yeah. But rebasing is tRiCkY

src/templates/entrypoint.sh
willcl-ark: leaving for Tor (for now)

src/templates/Dockerfile_unified_alpine
willcl-ark: leaving for Tor (for now)

willcl-ark:
OK final force push done.

_Now_ if this gets approved, I'll update images before removing the top commit and merging

josibake:
ACK https://github.com/bitcoin-dev-project/warnet/commit/d3cc379f2b5e00c7bfa8abdedbaba0aea276cf1d

looks RFM, will let @m3dwards and @pinheadmz take a look first tho

m3dwards:
ACK https://github.com/bitcoin-dev-project/warnet/commit/66444fff44256e45dc69feff8d28cf1cab9604e7

I think this is great.

josibake:
reACK https://github.com/bitcoin-dev-project/warnet/commit/66444fff44256e45dc69feff8d28cf1cab9604e7

diff looks good!

.github/workflows/test.yml
```diff
@@ -11,17 +11,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        backend: [compose, k8s]
+        backend: [compose]
     steps:
       - uses: actions/checkout@v3
-      - if: matrix.backend == 'compose'
```
pinheadmz, line 17: Did you have to remove this "if"? If we can just keep everything the same and remove "k8s" from the matrix on tests we wanna skip, that'll be cleaner and easier to undo

willcl-ark: When we add K8s back here, we'll likely need docker installed to build images. So we'd need to remove the if then. (Unless we find a weird way of building)

pinheadmz:
post post post post post post post post merge nit: https://github.com/bitcoin-dev-project/warnet/pull/376

PR #264, opened 2024-02-07
Make k8s DNS service lookup non-absolute

rtreffer:
The cluster domain name can be changed. This is particular useful when running multiple cluster or meshing clusters just via DNS + VPNs.

It is common to change this for production clusters.

See [DNS for Services and Pods](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/) for details on DNS search domains.


m3dwards:
Thanks for this! This is much better.

PR #268, opened 2024-02-08
Disable debug default

### LINKED ISSUE #266: Turn off debug=1 by default
Users can add logging on the nodes they're interested in in the graph file

willcl-ark:
Closes: #266 

This logging is overly verbose, disable it by default.
Add a demo debug category to default graph config.

Also include a commit to have k8s actually use bitcoind options from the graph (doh!). We now indirectly test this by having debug logging disabled and _requiring_ debug=validation to come from a test graph.

Finally add some logging to the k8s backend so we can see what's going on better

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -1,4 +1,5 @@
 import io
+import logging
 import re
 import time
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


m3dwards:
@willcl-ark this may have broken the rpc test?

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -368,9 +368,9 @@ def create_bitcoind_container(self, tank) -> client.V1Container:
         else:
             container_image = f"{DOCKER_REGISTRY_CORE}:{tank.version}"
 
-        container_env = [
-            client.V1EnvVar(name="BITCOIN_ARGS", value=self.default_bitcoind_config_args(tank))
-        ]
+        bitcoind_options = self.default_bitcoind_config_args(tank)
+        bitcoind_options += tank.conf
```
willcl-ark: This is the critical missing line in this commit

m3dwards:
Nice! Looks good.

PR #270, opened 2024-02-09
util: add json-parsing decorator

### LINKED ISSUE #249: Better debugging for exec / json parsing
While debugging the exec issue in https://github.com/bitcoin-dev-project/warnet/pull/248 I've mostly been greeted by generic JSON parse errors.
I would have expected some debug output about the actual issue.

Examples of failing code paths:
- `lnnode.py`
```python
    def getnewaddress(self):
        res = json.loads(self.lncli("newaddress p2wkh"))
        return res["address"]
```
- `ln_test.py`
```
inv = json.loads(base.warcli("lncli 2 addinvoice --amt=1234"))["payment_request"]
```

To improve debugging it would be great to have a "execute + parse json" method that
1. runs the command
2. tries to parse the output
3. on error: log the command + output and re-throw an error

Alternately it could be added to all code paths, but given that json output is the most common output I would expect it to be very redundant.

willcl-ark:
... & use in lnnode.

Fixes: #249 

I think this should somewhat fix #249, or at least provide a start for it.

### DIFF
--- a/src/warnet/lnnode.py
+++ b/src/warnet/lnnode.py
@@ -1,8 +1,7 @@
-import json
 import os
 
 from backends import BackendInterface, ServiceType
-from warnet.utils import exponential_backoff, generate_ipv4_addr
+from warnet.utils import exponential_backoff, generate_ipv4_addr, handle_json
 
 from .status import RunningStatus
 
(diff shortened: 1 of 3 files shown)


willcl-ark:
@josibake @rtreffer was this roughly what you had in mind?

rtreffer:
Oh nice, yeah

src/warnet/lnnode.py
```diff
@@ -22,28 +21,28 @@ def status(self) -> RunningStatus:
...
-        res = json.loads(self.lncli("getinfo"))
+        res = self.lncli("getinfo")
         return res["uris"][0]
 
     def get_wallet_balance(self):
-        res = json.loads(self.lncli("walletbalance"))
+        res = self.lncli("walletbalance")
         return res
 
+    @handle_json
```
josibake: why do you need the extra @handle_json here? isn't it already being handled on the `self.lncli()` call?

josibake (commented):
Nice!

src/warnet/lnnode.py
willcl-ark: Oh whoops, first i added it to all functions, before extracting it onto `lncli`, nice catch

josibake:
LGTM, thanks!

PR #273, opened 2024-02-13
Add prometheus to logging stack

m3dwards:
To test:

```shell
just start(d)
warcli network start
just connectlogging
```

Browse to localhost:3000 (username: admin, password: password)

See that we have Loki and Prometheus data sources already configured and that we have data for Loki and Prometheus. Prometheus bitcoin metrics can take a couple of minutes to appear.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -276,6 +276,7 @@ class KubernetesBackend(BackendInterface):
             try:
                 log_stream = self.client.read_namespaced_pod_log(
                     name=pod.metadata.name,
+                    container=BITCOIN_CONTAINER_NAME,
                     namespace=self.namespace,
                     timestamps=True,
                     _preload_content=False,
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -9,6 +9,7 @@ import yaml
 from backends import BackendInterface, ServiceType
 from cli.image import build_image
 from kubernetes import client, config
+from kubernetes.dynamic import DynamicClient
 from kubernetes.client.models.v1_pod import V1Pod
 from kubernetes.client.rest import ApiException
 from kubernetes.stream import stream
@@ -17,12 +18,15 @@ from warnet.status import RunningStatus
 from warnet.tank import Tank
 from warnet.utils import default_bitcoin_conf_args, parse_raw_messages
 
+
 DOCKER_REGISTRY_CORE = "bitcoindevproject/bitcoin"
 DOCKER_REGISTRY_LND = "lightninglabs/lnd:v0.17.0-beta"
 LOCAL_REGISTRY = "warnet/bitcoin-core"
+
 POD_PREFIX = "tank"
 BITCOIN_CONTAINER_NAME = "bitcoin"
 LN_CONTAINER_NAME = "ln"
+MAIN_NAMESPACE = "warnet"
 
 
 logger = logging.getLogger("KubernetesBackend")
@@ -35,6 +39,7 @@ class KubernetesBackend(BackendInterface):
         # running inside a k8s cluster as a statefulset
         config.load_incluster_config()
         self.client = client.CoreV1Api()
+        self.dynamic_client = DynamicClient(client.ApiClient())
         self.namespace = "warnet"
         self.logs_pod = logs_pod
         self.network_name = network_name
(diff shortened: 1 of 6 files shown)


src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -415,6 +420,37 @@ def create_bitcoind_container(self, tank) -> client.V1Container:
...
+                client.V1EnvVar(name="BITCOIN_RPC_HOST", value="localhost"),
+                client.V1EnvVar(name="BITCOIN_RPC_PORT", value=str(tank.rpc_port)),
+                client.V1EnvVar(name="BITCOIN_RPC_USER", value=tank.rpc_user),
+                client.V1EnvVar(name="BITCOIN_RPC_PASSWORD", value=tank.rpc_password),
+            ],
+        )
+
+    def apply_prometheus_scrape_configs(self, tanks):
+        for tank in tanks:
+            if not tank.exporter:
```
willcl-ark, line 441: Really like how this is only for enabled nodes 👍🏼

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -415,6 +420,37 @@ def create_bitcoind_container(self, tank) -> client.V1Container:
...
+                continue
+
+            bitcoind_service_name = self.get_service_name(tank.index)
+            target = f"{bitcoind_service_name}.{self.namespace}:9332"
```
willcl-ark: in a4d4ee1319cd135f519210a2be75924cd05dc5cd "Add prometheus to logging stack":
can this `9332` magic number be extracted into a variable somewhere? It's also used on line 554.

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -415,6 +420,37 @@ def create_bitcoind_container(self, tank) -> client.V1Container:
...
+            scrape_config = {
+                "apiVersion": "monitoring.coreos.com/v1alpha1",
+                "kind": "ScrapeConfig",
+                "metadata": {"name": "test", "namespace": "prometheus"},
+                "spec": {"scrapeInterval": "5s", "staticConfigs": [{"targets": [target]}]},
```
willcl-ark: in a4d4ee1319cd135f519210a2be75924cd05dc5cd "Add prometheus to logging stack":
perhaps (in the future) this `5s` constant should be configurable in the graph? Unclear to me currently.

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -487,8 +523,11 @@ def create_pod_object(
             ),
         )
 
+    def get_service_name(self, tank_index: int) -> str:
```
willcl-ark, line 552: Glad to see this here, I have it as an unrelated commit in a different branch. Hope to use this instead

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -557,6 +597,8 @@ def deploy_pods(self, warnet):
             bitcoind_pod = self.create_pod_object(
                 tank, bitcoind_container, self.get_pod_name(tank.index, ServiceType.BITCOIN)
             )
+            prometheus_container = self.create_prometheus_container(tank)
```
willcl-ark: in a4d4ee1319cd135f519210a2be75924cd05dc5cd "Add prometheus to logging stack":

We are only scraping selected containers per the filter on `tank.exporter`:

```python
    def apply_prometheus_scrape_configs(self, tanks):
        for tank in tanks:
            if not tank.exporter
```

Would it not make more sense to not add a prometheus container to them under the same condition?

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -26,7 +26,7 @@
 POD_PREFIX = "tank"
 BITCOIN_CONTAINER_NAME = "bitcoin"
 LN_CONTAINER_NAME = "ln"
-LOGGING_NAMESPACE = "warnet-logging"
+MAIN_NAMESPACE = "warnet"
```
willcl-ark, line 28: in 95610e44f38e2210d93fdffac73cc51a0ae0ed97 Add prometheus to logging stack:

this is added in the previous commit then removed here?

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -629,8 +637,9 @@ def deploy_pods(self, warnet):
                 self.client.create_namespaced_service(
                     namespace=self.namespace, body=lightning_service
                 )
-            # add metrics scraping for tanks configured to export metrics
-            self.apply_prometheus_scrape_configs(warnet.tanks)
+
+        # add metrics scraping for tanks configured to export metrics
+        self.apply_prometheus_service_monitor(warnet.tanks)
```
willcl-ark: in 95610e44f38e2210d93fdffac73cc51a0ae0ed97 Add prometheus to logging stack:

There seems to be many things modified here which were only added in the previous commit. Perhaps just squash them together? Alternatively just leave; it's not worth splitting and rebasing in any case.

willcl-ark (commented):
Changes look pretty much great! Nice work.

I did notice quite a few renames/re-touches between the first and second commits, not sure if its just worth squashing those?

One main Q regarding only starting node-exporter for those nodes which prometheus is configured for, but otherwise a tACK from me

src/backends/kubernetes/kubernetes_backend.py
m3dwards: Scrapeconfig is no longer used, bad (no) squashing on my part.

src/backends/kubernetes/kubernetes_backend.py
m3dwards: squashed them together

src/backends/kubernetes/kubernetes_backend.py
m3dwards: squashed

src/backends/kubernetes/kubernetes_backend.py
m3dwards: Added this

src/backends/kubernetes/kubernetes_backend.py
m3dwards: Extracted

m3dwards:
@willcl-ark I've addressed all your comments.

I still want to make the logging stack optional but that can come in on it's own PR if this gets merged first.

willcl-ark:
ACK 1850e31a9ee435ae4ef36256ab7ba758fe41b3e7

Working nicely now, only on configured pods.

PR #274, opened 2024-02-14
Include exec for entryfile on image

m3dwards:


src/templates/entrypoint.sh
```diff
@@ -41,7 +41,7 @@ fi
 
 if [ "$1" = "bitcoind" ] || [ "$1" = "bitcoin-cli" ]; then
   echo
-  "$@"
+  exec "$@"
```
willcl-ark: Want to just remove the block entirely now we're root?

willcl-ark:
ACK 3a302a18e78d0a5516a5d15b39163e7346ba25df

Do you want to do the images or shall I? It's a shame I can't do the AMD64 ones, and you the ARM64 ones, then we merge into a multiarch image on docker hub...

willcl-ark:
> ACK [3a302a1](https://github.com/bitcoin-dev-project/warnet/commit/3a302a18e78d0a5516a5d15b39163e7346ba25df)
> 
> Do you want to do the images or shall I? It's a shame I can't do the AMD64 ones, and you the ARM64 ones, then we merge into a multiarch image on docker hub...

Seems like we might be able to do this with `docker manifest` [command](https://docs.docker.com/engine/reference/commandline/manifest/)... maybe not worth it though

m3dwards:
I wondered if new images were being made anyway whether to just wait and let them get replaced in time. It's not a mission critical failure?

PR #275, opened 2024-02-14
Make logging stack optional

m3dwards:
Now logging stack will not be created with `just start` and the kubernetes backend will be just fine with that and not install the prometheus exporters.

How to test:

```shell
just stop(d)
k delete crds --all
just start(d)
warcli network start --force
```

Observe that the containers do not have prometheus exporters and everything is happy

```shell
warcli network down
just installlogging
warcli network start --force
just connectlogging
```

Now when you browse to localhost:3000 (admin/password) Loki and prometheus should be present and working.

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,7 +23,6 @@ start:
     kubectl apply -f src/templates/rpc/rbac-config.yaml
     kubectl apply -f src/templates/rpc/warnet-rpc-service.yaml
     kubectl apply -f src/templates/rpc/warnet-rpc-statefulset-dev.yaml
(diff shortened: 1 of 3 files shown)


willcl-ark:
OK, I think we're going to need a series of `doc/` PRs very soon. Lots more on the K8s/logging stuff, image gen stuff (from me) ...Lots of new command combos becoming available!

willcl-ark:
LGTM. Logging still slow to shutdown for me, so happy to have this in 👍🏼

PR #276, opened 2024-02-14
Only collect logs if graph has collect_logs set

m3dwards:
How to test:

```shell
just start(d)
just installlogging
warcli network start --force
warcli rpc 2 'logging [\"all\"]'
warcli rpc 7 'logging [\"all\"]'
just connectlogging
```

Browse to `localhost:3000` (admin/password) and explore the loki data source. Using the label explorer, observe that many logs are available for `{pod="warnet-tank-000002"}` but none are available for `{pod="warnet-tank-000007"}` even though both have been set to log everything. This is because tank 02 has collect_logs set to true in it's graph but tank 07 does not.


willcl-ark:
This also LGTM. Going to let main CI finish before merging.

PR #277, opened 2024-02-15
fixup get_messages to use tank indexes

willcl-ark:
Adds a (compose) get_ipv4_address func and then switches up `get_messages()` to use two tank indexes, rather than an ipv4 address.

Also fixup k8s `get_file()` while we're here and need to use it.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -1,4 +1,4 @@
-import io
+import base64
 import logging
 import re
 import time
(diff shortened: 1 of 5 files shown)


willcl-ark:
Note that I "had" to use base64 again for get_file. Seems like K8s python SDK doesn't like bytes...

I did try to mimic docker (which tars it on remote and untars it locally?), but didn't succeed. It might be possible though...

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -78,7 +78,7 @@ def get_file(self, tank_index: int, service: ServiceType, file_path: str):
         Read a file from inside a container
         """
         pod_name = self.get_pod_name(tank_index, service)
-        exec_command = ["cat", file_path]
+        exec_command = ['sh', '-c', f'cat "{file_path}" | base64']
```
willcl-ark, line 81: @m3dwards what's you'r professional opinion on bas64-ing files out of k8s? Kinda nasty, but we are forced into returning `str`, so I don't really see a better way...

m3dwards: When I last looked into it, I think I read that some people were TARing but I don't know how that helps with the bytes problem as it would just be bytes of a tar?

If base64 works then we can keep it like that.

PR #279, opened 2024-02-15
ci: only run build when all others succeed

willcl-ark:
Only run build job if other (faster) jobs succeed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,27 +7,6 @@ on:
   pull_request:
 
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


willcl-ark:
I feel like  it makes sense not to run the heavy build job unless the faster-failing ones complete. We could even rewrite them all in a dependant chain based on test speed, to save resources and get faster red dots...

See below how the `build` job was only triggered by all others completing.

image

pinheadmz (approved):
LGTM smart move thanks

josibake (approved):
Nice, ACK

PR #281, opened 2024-02-16
server: make all connections with addnode

willcl-ark:
using `addpeeraddress` sounds better, but it does not guarantee that all nodes have connections made to them in the end. This causes issues on large graphs. We have observed on n=100 graphs, often there exist ~5 nodes which end up with no connections.

First, add a more accurate healthcheck for compose backend, which will not return true until `bitcoind` RPC server is listening.

Next add a wait for healthy to startup, on k8s this is currently inactive (although notable k8s does use a "port listen" healthcheck already).

Finally remove the `addpeeraddress` connection type.

### DIFF
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -49,7 +49,8 @@ class ComposeBackend(BackendInterface):
     def __init__(self, config_dir: Path, network_name: str) -> None:
         super().__init__(config_dir)
         self.network_name = network_name
-        self.client = docker.DockerClient = docker.from_env()
+        self.client: docker.DockerClient = docker.from_env()
+        self._apiclient: docker.APIClient = docker.APIClient(base_url='unix://var/run/docker.sock')
 
     def build(self) -> bool:
         command = ["docker", "compose", "build"]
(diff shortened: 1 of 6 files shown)


src/backends/compose/compose_backend.py
```diff
@@ -388,7 +388,7 @@ def add_services(self, tank: Tank, services):
                 "privileged": True,
                 "cap_add": ["NET_ADMIN", "NET_RAW"],
                 "healthcheck": {
-                    "test": ["CMD", "pidof", "bitcoind"],
+                    "test": ["CMD-SHELL", f"nc -z localhost {tank.rpc_port} || exit 1"],
```
willcl-ark, line 395: For k8s we are already checking port "liveness" using:

```python
readiness_probe=client.V1Probe(
    failure_threshold=1,
    initial_delay_seconds=0,
    period_seconds=1,
    timeout_seconds=1,
    tcp_socket=client.V1TCPSocketAction(port=tank.rpc_port),
),
```

willcl-ark:
@pinheadmz do you have any objection to moving away from `addpeeraddress`? It appears to fix the behaviour we observed on large clusters.

PR #283, opened 2024-02-21
backend: add debug exec output

willcl-ark:
Better debug logs for exec actions


pinheadmz:
untested, concept ACK

PR #284, opened 2024-02-21
Add graph schema

willcl-ark:
Long enough we have had not canonical graph schema with which to validate graphs against.

Before adding more properties to it for lightning (and perhaps circuitbreaker), create a schema and validation code.

Also includes a bonus traceback touch up, to finally log tb of unhandled exceptions and avoid tedious debug stepping.

### DIFF
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -332,10 +332,10 @@ class ComposeBackend(BackendInterface):
             dirs_exist_ok=True,
         )
 
-    def config_args(self, tank):
+    def config_args(self, tank: Tank):
         args = self.default_config_args(tank)
-        if tank.conf is not None:
-            args = f"{args} -{tank.conf.replace(',', ' -')}"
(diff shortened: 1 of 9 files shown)


src/backends/compose/compose_backend.py
```diff
@@ -363,7 +363,7 @@ def add_services(self, tank: Tank, services):
             # it's a git branch, building step is necessary
             repo, branch = tank.version.split("#")
             services[container_name]["image"] = f"{LOCAL_REGISTRY}:{branch}"
-            build_image(repo, branch, LOCAL_REGISTRY, branch, tank.DEFAULT_BUILD_ARGS + tank.extra_build_args, arches="amd64")
+            build_image(repo, branch, LOCAL_REGISTRY, branch, tank.DEFAULT_BUILD_ARGS + tank.build_args, arches="amd64")
```
willcl-ark, line 368: FWIW `tank` properties names MUST now exactly match the graph schema, as they are parsed and set according to it.

Should probably document that...

PR #285, opened 2024-02-21
Replace connect_edges() with bitcoind -addnode args 

pinheadmz:
Goal:

Replace `connect_edges()` which was slow with `-addnode=...` args passed to bitcoind at startup

Method:

Add `init_peers` list to `Tank` objects which contains destination tank indexes imported from the graph file.

Requires:

- Since we don't know any tank IP addresses before we start up a network (especially in Kubernetes) we need to pass internal-network domain names to `addnode` instead.
- Since `connect_edges()` was also blocking, a new rpc is added and used in tests to determine if all expected graph edges have been established. The command is `warcli network connected` and it only ever returns `true` or `false`. Right now it checks that the number of `tank.init_peers` equals the number of bitcoin peers connected with type `"manual"`. As discussed, this RPC can be softened in a follow-up PR to accept some kind of threshold argument.

Fixes:
- In Kubernetes the *service* name is how we connect to bitcoin peers so the services need to forward the regtest p2p port 18444.
- This also means that internally each peer apparently has two IP addresses (the bitcoind container IP which may be unreliable and the service IP which is always resolved by the service name). So far I think the only problem with the 2-IPs thing is how we retrieve p2p messages with `warcli messages <a> <b>`, but we should also be aware in the future of connecting twice to the same node.
- During shutdown I noticed we did not handle missing prometheus exporters well because that method uses a different client with a different error type. Probably has just not been an issue before because the exporters were always there during shutdown. I'm not sure why that would be different now unless it has to do with tests running faster?
- `graph_generate` was adding edges to the graph that already existed in the opposite direction. That was breaking the `connect_edges` check since bitcoind would never actually make that connection

### DIFF
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -352,6 +352,9 @@ class ComposeBackend(BackendInterface):
         defaults += f" -rpcport={tank.rpc_port}"
         defaults += f" -zmqpubrawblock=tcp://0.0.0.0:{tank.zmqblockport}"
         defaults += f" -zmqpubrawtx=tcp://0.0.0.0:{tank.zmqtxport}"
+        # connect to initial peers as defined in graph file
+        for dst_index in tank.init_peers:
+            defaults += f" -addnode={self.get_container_name(dst_index, ServiceType.BITCOIN)}"
         return defaults
 
     def copy_configs(self, tank):
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -161,26 +161,20 @@ class Warnet:
                 raise Exception(
                     f"Node ID in graph must be incrementing integers (got '{node_id}', expected '{len(self.tanks)}')"
                 )
-            self.tanks.append(Tank.from_graph_node(node_id, self))
+            tank = Tank.from_graph_node(node_id, self)
+            # import edges as list of destinations to connect to
+            for edge in self.graph.edges(data=True):
+                (src, dst, data) = edge
+                if "channel" in data:
+                    continue
+                if src == node_id:
+                    tank.init_peers.append(int(dst))
+            self.tanks.append(tank)
         logger.info(f"Imported {len(self.tanks)} tanks from graph")
 
     def apply_network_conditions(self):
         for tank in self.tanks:
             tank.apply_network_conditions()
-
-    def connect_edges(self):
-        if self.graph is None:
-            return
-
-        for edge in self.graph.edges(data=True):
-            (src, dst, data) = edge
-            if "channel" in data:
-                continue
-            src_tank = self.tanks[src]
-            dst_ip = self.tanks[dst].ipv4
-            cmd = f"bitcoin-cli -regtest -rpcuser={src_tank.rpc_user} -rpcpassword={src_tank.rpc_password} addnode {dst_ip}:18444 onetry"
-            logger.info(f"Using `{cmd}` to connect tanks {src} to {dst}")
-            src_tank.exec(cmd=cmd)
 
     def warnet_build(self):
         self.container_interface.build()
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -221,3 +221,16 @@ class Warnet:
 
     def wait_for_health(self):
         self.container_interface.wait_for_healthy_tanks(self)
+
+    def network_connected(self):
+        for tank in self.tanks:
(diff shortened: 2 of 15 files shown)


src/backends/compose/compose_backend.py
```diff
@@ -345,6 +345,9 @@ def default_config_args(self, tank):
         defaults += f" -rpcport={tank.rpc_port}"
         defaults += f" -zmqpubrawblock=tcp://0.0.0.0:{tank.zmqblockport}"
         defaults += f" -zmqpubrawtx=tcp://0.0.0.0:{tank.zmqtxport}"
+        # connect to initial peers as defined in graph file
+        for dst_index in tank.init_peers:
```
willcl-ark, line 356: We could print a warning here if we exceed n==8 so that someone would understand why not all their graph connections are being made (and in the k8s equivalent)

src/warnet/warnet.py
```diff
@@ -159,27 +159,21 @@ def tanks_from_graph(self):
                 raise Exception(
                     f"Node ID in graph must be incrementing integers (got '{node_id}', expected '{len(self.tanks)}')"
                 )
-            self.tanks.append(Tank.from_graph_node(node_id, self))
+            tank = Tank.from_graph_node(node_id, self)
+            # import edges as list of destinations to connect to
+            for edge in self.graph.edges(data=True):
```
willcl-ark, line 166: Seems like it should be possible to do something like:

```python
for edge in self.graph.nodes[i].edges:
```

willcl-ark: In fact, this is changed again in the subsequent commit. Perhaps just squash?

willcl-ark (commented):
Light code review ACK from me so far. Not done any testing and will try to look more closely later.

src/backends/compose/compose_backend.py
pinheadmz: Could that be added to the graph schema validate RPC?

src/warnet/warnet.py
pinheadmz: nice suggestion thanks ill take that

pinheadmz: added in https://github.com/bitcoin-dev-project/warnet/pull/285/commits/d51d0832662d2965c235401c209c56daa7c14719

m3dwards:
Couple of thoughts:

The threshold for connected peers is mentioned but unless I'm not reading correctly doesn't seem to be implemented?

The `base.wait_for_all_edges()` function would be very useful to be able to call from a scenario and not just a test file. And should it also be a function where you pass in how many peer connections you are looking for again rather than all? Do we want to use graphy language like edges rather than what they represent, i.e peers?

willcl-ark:
```cpp
    /**
     * We open manual connections to addresses that users explicitly requested
     * via the addnode RPC or the -addnode/-connect configuration options. Even if a
     * manual connection is misbehaving, we do not automatically disconnect or
     * add it to our discouragement filter.
     */
    MANUAL,
```

I think it's worth noting here that these manual connections do not have the regular eviction logic applied to them, which could be of consequence in some networks or tests.

It's also worth noting that (IIUC) DNS seeding is incompatible with regtest.

Therefore it's important to understand (and perhaps should be documented) that edges described on the graph are unlikely to be disconnected.

pinheadmz:
> It's also worth noting that (IIUC) DNS seeding is incompatible with regtest.

I thought we got `dummyseed.invalid` to resolve? We could also use `-seednode`

pinheadmz:
> The threshold for connected peers is mentioned but unless I'm not reading correctly doesn't seem to be implemented?

That's correct - what kind of threshold do you think we need to support? It could be a float (percent of edges listed in graph) or int (total number of connected edges) or something on a per-peer basis (every tank must have *n* peers).


 
> The `base.wait_for_all_edges()` function would be very useful to be able to call from a scenario and not just a test file. 

Added and covered by test in b6b9de05d876ec70d4e3e160bbee1152e1015a60

pinheadmz:
also rebased on main to take advantage of #290

willcl-ark (approved):
ACK b6b9de05d876ec70d4e3e160bbee1152e1015a60

I like this, it works nicely and should provide much faster startup connection as all nodes will autonomously make their connections in parallel.

The CLI command is also nice. Part of me can't help but think it'd be nice to poll this on startup and log a "connected" log message once we are connected, but it's likely not worth the effort (and might strangle large nertworks with requests).

PR #287, opened 2024-02-26
Allow custom LN images

### LINKED ISSUE #226: Allow Custom Branches of LND Nodes
Add the ability to run a custom version of LND. 

No strong preferences as to how this is achieved, could be: 
* Point to a branch + warnet builds
* Point to a docker image that's built using LND's [standard dockerfile](https://github.com/lightningnetwork/lnd/blob/master/Dockerfile)

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/226

Adds to the graphml schema a new data element for tanks: `ln-image`.
If the element `ln` is missing, then `ln-image` is ignored and the tank will not get an ln node.
If `ln-image` is missing, the current default value (set in lnnode.py) is `"lightninglabs/lnd:v0.17.0-beta"`

Example of two tanks, one using the default image and the other specifying v15.5

```
    <node id="0">
        <data key="version">26.0</data>
        <data key="bitcoin_config">-uacomment=w0</data>
        <data key="ln">lnd</data>
        <data key="collect_logs">true</data>
    </node>
    <node id="1">
        <data key="version">26.0</data>
        <data key="bitcoin_config">-uacomment=w1</data>
        <data key="ln">lnd</data>
        <data key="ln-image">lightninglabs/lnd:v0.15.5-beta</data>
        <data key="collect_logs">true</data>
    </node>
```

### DIFF
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -443,7 +443,7 @@ class ComposeBackend(BackendInterface):
         ]
         services[ln_container_name] = {
             "container_name": ln_container_name,
-            "image": "lightninglabs/lnd:v0.17.0-beta",
+            "image": tank.lnnode.image,
             "command": " ".join(args),
             "networks": {
                 tank.network_name: {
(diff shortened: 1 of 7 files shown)


m3dwards:
LGTM

PR #289, opened 2024-02-28
Added a new scenario cli command "run-file"

m3dwards:
This will base64 up the file and throw it over to the RPC where it can be run.

The existing scenarios and commands are unchanged but I'm not sure it makes sense to have both run and run-file.

Proposing this as an immediate enhancement to be able to run modified scenarios on k8s clusters without having to rebuild the RPC image and to spark debate on how scenarios should run in the future.

To test:

```
warcli scenarios run-file src/scenarios/tx_flood.py
warcli scenarios stop <PID>`
```

### DIFF
--- a/src/cli/scenarios.py
+++ b/src/cli/scenarios.py
@@ -1,3 +1,4 @@
+import base64
 import sys
 
 import click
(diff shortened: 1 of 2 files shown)


src/cli/scenarios.py
```diff
@@ -49,6 +50,26 @@ def run(scenario, network, additional_args):
     print(rpc_call("scenarios_run", params))
 
 
+@scenarios.command(context_settings={"ignore_unknown_options": True})
+@click.argument("scenario_path", type=str)
+@click.argument("additional_args", nargs=-1, type=click.UNPROCESSED)
+@click.option("--network", default="warnet", show_default=True)
+def run_file(scenario_path, network, additional_args):
+    """
+    Run <scenario_path> from the Warnet Test Framework on <--network> with optional arguments
```
josibake, line 59: nit: not sure if `scenario_path` is enough of a distinction to let a user know how this command is different from the command above. also have a slight aversion to `_` in the command verb names.

josibake (approved):
Looks good! didn't test, but seems harmless enough since this doesn't change any existing code (and I want this for my testing tomorrow :smile: ).

Left a comment regarding the naming/help text but moreso for posterity as I imagine we won't always have two commands once we settle on the best way to run scenarios.

src/cli/scenarios.py
m3dwards: Agree with the nit. Hopefully we can have a chat about how we want scenarios to be in the future and make it cleaner / clearer. I just wanted something now to unblock testing.

PR #288, opened 2024-02-27
Add circuitbreaker to lnd

pinheadmz:
Closes: https://github.com/bitcoin-dev-project/warnet/issues/228
Requires (based on) https://github.com/bitcoin-dev-project/warnet/pull/287

Adds one more data element to tanks in the graphml file:

```
 <data key="ln-cb-image">carlakirkcohen/circuitbreaker:endorsement-experiment</data>
```

This adds a container running circuit breaker, and connects it to the corresponding lnd container. In K8s, both containers are in the same pod so they can interact via `localhost` and share a volume at `/root/.lnd` in both containers. In docker they also share a volume with that same path (this is needed to pass circuitbreaker the credentials generated by lnd).

~I'm not sure how else to test circuit breaker~, but the logs look promising. It keeps retrying to connect until it finally does:

```
2024-02-27 12:55:38 2024-02-27T17:55:38.313Z    INFO    Circuit Breaker starting        {"version": ""}
2024-02-27 12:55:38 2024-02-27T17:55:38.313Z    INFO    Opening database        {"path": "/root/.circuitbreaker/circuitbreaker.db"}
2024-02-27 12:55:38 2024-02-27T17:55:38.314Z    ERROR   Unexpected exit {"err": "open /root/.lnd/tls.cert: no such file or directory"}
2024-02-27 12:55:38 main.main
2024-02-27 12:55:38     /src/main.go:144
2024-02-27 12:55:38 runtime.main
2024-02-27 12:55:38     /usr/local/go/src/runtime/proc.go:250
2024-02-27 12:55:44 2024-02-27T17:55:44.878Z    INFO    Circuit Breaker starting        {"version": ""}
2024-02-27 12:55:44 2024-02-27T17:55:44.878Z    INFO    Opening database        {"path": "/root/.circuitbreaker/circuitbreaker.db"}
2024-02-27 12:55:44 2024-02-27T17:55:44.884Z    INFO    Press ctrl-c to exit
2024-02-27 12:55:44 2024-02-27T17:55:44.884Z    INFO    Grpc server starting    {"listenAddress": "127.0.0.1:9234"}
2024-02-27 12:55:44 2024-02-27T17:55:44.884Z    INFO    CircuitBreaker started
2024-02-27 12:55:44 2024-02-27T17:55:44.884Z    INFO    HTTP server starting    {"listenAddress": "127.0.0.1:9235"}
[...]

PR #290, opened 2024-02-29
No more `from_deployment()`

pinheadmz:
The graphml has everything we need, except IP addresses. We can get those from the backend, but that's only actually needed for scenarios currently.

#254 and #278 are already cleaning up IP stuff... ultimately we should be able to get rid of a lot of the metadata we dump in the compose file in `environment` variables and `labels` and in the kubernetes deployment several `V1EnvVar` are set just for the sake of recovering "from deployment".

This also means initiatives like #285 and #288 will be easier to build since we wont need to add extra metadata in the deployments.


willcl-ark:
ACK. Nice cleanup!

PR #292, opened 2024-02-29
include schema and templates in non-editable images

willcl-ark:
previously this relied on `pip install -e .` style install, and was not being copied across in a regular `pip install .`

Fix that by putting the schema in it's own package, and explicitly including that in `MANIFEST.in`, along with other files needed from `templates/`.


willcl-ark:
This also switches the RPC statefulset to use the `latest` tag by default. IMO we should now think about adding a GH action to build merges to `main`, which will push the new `latest` image to docker hub.

WDYT @m3dwards @pinheadmz @josibake ?

willcl-ark:
BTW I did test this PR (image tag 0.4/latest on dockerhub) using a digital ocean cluster and... it works for me (TM)

src/templates/rpc/warnet-rpc-statefulset.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
       - name: warnet-rpc
         imagePullPolicy: Always
-        image: bitcoindevproject/warnet-rpc:0.2
+        image: bitcoindevproject/warnet-rpc:latest
```
m3dwards, line 20: Just a note to say that this means if we push semantic versioned images we will also need to push an image with the tag `latest` explicitly for this to work.

m3dwards:
Looks good to me. I especially like the .dockerignore.

src/templates/rpc/warnet-rpc-statefulset.yaml
m3dwards: and I see you have the option for that in the build script.

willcl-ark: My intention was that we could push any tag we like, and then optionally also specify it as `latest`? There is also a `latest` tag there [now](https://hub.docker.com/layers/bitcoindevproject/warnet-rpc/latest/images/sha256-8f620c5a24352daf75030b779c12960aaaf4198682bf834afe609b3d844d90b3?context=explore)

PR #296, opened 2024-03-04
update test_framework to v26.0 and add a miniwallet prop to test_node

willcl-ark:
This updates the test_framework to Bitcoin Core v26.0, which can be used in e.g. double_tx_relay tests.
Also add a miniwallet property to test_node.


willcl-ark:
@pinheadmz you want to check this before merge?

pinheadmz (approved):
LGTM 👍 
Confirmed no diff when manually importing v26.0 test_framework except for addiiton of miniwallet in test base. CI is green, the trap is clean.

PR #301, opened 2024-03-05
ci: build RPC image automagically

willcl-ark:
This will build a warnet-rpc image when merging into main branch

### DIFF
--- a/.github/workflows/build-rpc-images.yml
+++ b/.github/workflows/build-rpc-images.yml
@@ -1,0 +1,42 @@
+name: build-docker-rpc-images
+
+on:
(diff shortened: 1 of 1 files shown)


willcl-ark:
I hate workflows

willcl-ark:
Ok this does work, but currently it will just push a new "latest" tag when we merge things into `main` branch, see [here](https://hub.docker.com/r/will8clark/warnet-rpc/tags) for example.

I wanted to have each pr get it's own tag, but thats difficult when PRs come from forks (you have to use a GitHub "environment", and I cannot get it to read secrets out of the environment correctly).

PR #304, opened 2024-03-05
Cache warnets on the server

### LINKED ISSUE #155: Cache active networks on server
Avoid recreating them on every call, when not necessary.

willcl-ark:
Avoids repeated rebuilding of the warnet when calling commands

Closes #155

### DIFF
--- a/src/warnet/server.py
+++ b/src/warnet/server.py
@@ -64,6 +64,7 @@ class Server:
         self.setup_global_exception_handler()
         self.setup_logging()
         self.setup_rpc()
+        self.warnets: dict = dict()
         self.logger.info("Started server")
 
         # register a well known /-/healthy endpoint for liveness tests
@@ -179,13 +180,27 @@ class Server:
         # Logs
         self.jsonrpc.register(self.logs_grep)
 
+    def get_warnet(self, network: str) -> Warnet:
+        """
+        Will get a warnet from the cache if it exists.
+        Otherwise it will create the network using from_network() and save it
+        to the cache before returning it.
+        """
+        if network in self.warnets:
+            return self.warnets[network]
+        wn = Warnet.from_network(network, self.backend)
+        if isinstance(wn, Warnet):
+            self.warnets[network] = wn
+            return wn
+        raise ServerError(f"Could not find warnet {network}")
+
     def tank_bcli(
         self, node: int, method: str, params: list[str] | None = None, network: str = "warnet"
     ) -> str:
         """
         Call bitcoin-cli on <node> <method> <params> in [network]
         """
-        wn = Warnet.from_network(network, self.backend)
+        wn = self.get_warnet(network)
         try:
             return wn.container_interface.get_bitcoin_cli(wn.tanks[node], method, params)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


src/warnet/server.py
```diff
@@ -269,7 +278,8 @@ def network_export(self, network: str) -> str:
         Export all data for sim-ln to subdirectory
         """
         try:
-            wn = Warnet.from_network(network)
+            wn = self.get_warnet(network)
+            self.warnets[network] = wn
```
pinheadmz: why would you need to set this `wn` again?

willcl-ark: That is a good question, I think it's a mistake.

src/warnet/server.py
```diff
@@ -424,7 +434,7 @@ def scenarios_list_running(self) -> list[dict]:
         return running
 
     def network_up(self, network: str = "warnet") -> str:
-        def thread_start(wn):
+        def thread_start(server: Server, network):
```
pinheadmz, line 443: don't think `server` is used in the function here? Also im not sure but there is a `self.backend` -- i think that `self` *IS* the server?

src/warnet/server.py
```diff
@@ -440,7 +450,8 @@ def thread_start(wn):
 
         try:
             wn = Warnet.from_network(network, self.backend)
-            t = threading.Thread(target=lambda: thread_start(wn))
+            self.warnets[network] = wn
```
pinheadmz: here, too ?

src/warnet/server.py
```diff
@@ -456,11 +467,12 @@ def network_from_file(
         Run a warnet with topology loaded from a <graph_file>
         """
 
-        def thread_start(wn, lock: threading.Lock):
-            with lock:
+        def thread_start(server: Server, network):
```
pinheadmz, line 472: here too, too. Just not sure why server needs to be passed in

src/warnet/server.py
```diff
@@ -480,11 +492,13 @@ def thread_start(wn, lock: threading.Lock):
...
-            wn = Warnet.from_graph_file(graph_file, config_dir, network, self.backend)
-            t = threading.Thread(target=lambda: thread_start(wn, self.image_build_lock))
+            self.warnets[network] = Warnet.from_graph_file(
+                graph_file, config_dir, network, self.backend
+            )
+            t = threading.Thread(target=lambda: thread_start(self, network))
             t.daemon = True
             t.start()
-            return wn._warnet_dict_representation()
+            return self.warnets[network]._warnet_dict_representation()
```
pinheadmz, line 503: okay... or you could keep the local `wn` variable and add it to `self.warnets` right after `start()`

pinheadmz (commented):
concept ACK, havent tested yet but this is super duper important and badly needed THANK YOU.

One additional comment: if `--force` is passed to `network start` we MUST clear the cached warnet with the same name!

update: okay i think this is already happening in `network_from_file`

willcl-ark:
Thanks for the review, I will re-view it now and incorporate any needed changes.

willcl-ark:
@pinheadmz wdyt of the changes now? I think I addressed all comments

src/warnet/server.py
willcl-ark: Now using server (I find it makes more sense in a thread than using `self` from the outer scope?)

src/warnet/server.py
willcl-ark: I think it's the same thing? Just referencing the dict here

pinheadmz: it is, its fine thought it looked prettier without the extra dict reference

pinheadmz (approved):
👍 
🚢 IT

PR #303, opened 2024-03-05
Remove ipv4 v3

willcl-ark:
Remove the ipv4 stuff (and fixed subnet from compose).

Closes #253

PR #306, opened 2024-03-06
Internal Tor (compose only, for now)

pinheadmz:
Motivation: local testing of https://github.com/bitcoin/bitcoin/pull/29415

Adds two new properties to the graph file. Both are booleans labeled "tor". At the graph level, this will deploy a Tor DA container on the network. At the tank level, it will start the tor daemon in that particular tank. This keeps the resource-hungry Tor as optional as possible. Of course you will also need to configure bitcoind in the tanks to use Tor once it's up.

```xml
  <graph edgedefault="directed">
    <data key="tor">true</data>
    <node id="0">
        <data key="image">pinheadmz/bitcoin:26.0_with_tor</data>
        <data key="bitcoin_config">-uacomment=w0 -onion=localhost:9050</data>
        <data key="tor">true</data>
    </node>
```

Still using hard coded IP `100.20.15.18` (20-15-18 is "t-o-r" ;-) ) for the Tor DA, but I think with some clever entrypoint and environment variable stuff we can make this more flexible and eventually deploy in k8s as well.

Uses as a healthcheck the Tor daemon endpoint  `status/bootstrap-phase"` (see https://spec.torproject.org/control-spec/commands.html) and similar to LN integration, I wrote a scenario `onion_init` which waits for DA health, and then creates a ring shaped network graph from all the tanks' onion addresses. Shaping this graph or using edges configured in the graphml file is a future TODO.


willcl-ark:
Hardcoded ip will conflict with #303

pinheadmz:
> Hardcoded ip will conflict with #303

I think as long as we keep the network subnet in compose `100.0.0.0/8` the Tor DA static IP will continue to work ok

pinheadmz:
Soft balling the onion test. Only waiting until `num_nodes / 2` total connections are established since they take so damn long to connect over onion and also appear to drop in and out 😬 

Locally this test passes like this:
node 0: 1 in
node 1: 1 in, 1 out
node 2: 1 in, 1 out
node 3: 1 out
nodes 4-8: bubkis

pinheadmz:
something works!

```
Bitcoin Core client v26.1.0rc1 regtest - server 70016/Satoshi:26.1.0(w5)/

<->   type   net  mping   ping send recv  txn  blk  hb addrp addrl  age id address                                                              version
out   full onion     94     94    6    6                              0  0 qbbdgs5myqcbbff34f2vrhif4eqpnevpu53oxw5dsqz74gwvvsj77nyd.onion:18444 70016/Satoshi:26.1.0(w4)/
                     ms     ms  sec  sec  min  min                  min

         ipv4    ipv6   onion   total   block
in          0       0       0       0
out         0       0       1       1       0
total       0       0       1       1

Local addresses
0.0.0.0                                                            port  18444    score      2
100.170.48.88                                                      port  18444    score      1
::                                                                 port  18444    score      2
ce6t3pwbm3hmxehqxlijghfbtrxds5fg3pvksbf7qq2qxdz43mwxnsqd.onion     port  18444    score      4
```

pinheadmz:
closing, will revisit eventually in k8s

PR #315, opened 2024-03-12
support channel policies in LN edges

### LINKED ISSUE #227: Support Channel Policies in Lightning Edges
To allow more realistic mimicking of the real world lightning graph, add support for setting channel policies on lightning edges. 

In lightning, one channel `A ---- B` will have two different sets of channel policies: 
* Set by `A`: applies to all payments sent `A ----> B`
* Set by `B`: applies to all payments sent `B ----> A`

These can _in theory_ be set on open, but it's difficult to do for the non-opening party (ie, when A opens to B, setting B's channel policy requires a complex interception API in LND. I'd suggest opening the channel fully and then using [UpdateChannelPolicy](https://lightning.engineering/api-docs/api/lnd/lightning/update-channel-policy) to set each side of the channel's policy. One exception to this is the `cltv_expiry_delta`, which applies across channels. This can be set on a node level, and set in LND's config using `--bitcoin.timelockdelta=`. 

Channel level policies -> [Update Request API](https://lightning.engineering/api-docs/api/lnd/lightning/update-channel-policy#lnrpcpolicyupdaterequest)
- Base fee rate -> `base_fee_msat`
- Proportional fee rate -> `fee_rate_ppm`

These will need to be specified for the `source` and `target` node in the XML file. 

Out of Scope/Bonus points: 
Channel policies in LN take some time to propagate through the network. We can use [GetChanInfo](https://lightning.engineering/api-docs/api/lnd/lightning/get-chan-info/index.html) to check that the whole network has seen the update. This is more trouble that it's worth initially, but would certainly be a nice to have for setting the whole network up in a known starting state.

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/227

Taking into account https://github.com/bitcoin-dev-project/warnet/pull/303 as well and using hostnames instead of ipv4 addresses where applicable.

I think this may have already been done but in the graph file now `bitcoin_config` uses command line argument syntax (i.e. `-key=value ` not bitcoin.conf style `key=value\n`)

Pushing up commit by commit to burn CI minutes


src/warnet/tank.py
```diff
@@ -123,6 +135,18 @@ def status(self) -> RunningStatus:
     def exec(self, cmd: str):
         return self.warnet.container_interface.exec_run(self.index, ServiceType.BITCOIN, cmd=cmd)
 
+    def get_bitcoin_conf(self, nodes: list[str]) -> str:
```
willcl-ark, line 139: In 7f4e28460dd0ff75de3d74eb5816df6f57b3f276

If you re-touch could rename this to `gen_bitcoin_options` or something now?

docs/lightning.md
```diff
@@ -17,17 +17,18 @@ Example:
...
 ## Adding LN channels to graph
 
-LN channels are represented in the graphml file as edges with an extra data element
-with key `"channel"` and a value in satoshis representing channel capacity. Note
-that this data element is the only difference between LN channels and regular bitcoin
-p2p connections. The graph will be considered a `MultiDiGraph` and contain two different
-kinds of edges.
+LN channels are represented in the graphml file as edges with extra data elements
+that correspond to arguments to the lnd `openchannel` and `updatechanpolicy` RPC
+commands. The keys are with either `"source-policy"` (arguments added to `openchannel`)
```
willcl-ark: This is a neat way of doing it, nice!

willcl-ark (commented):
Approach ACK so far.

Like the direction this is going. Ripping out our half-config half-args setup feels right. And adding the channel properties directly to the edges also feels correct.

How much more is there to come?

scripts/graphdocs.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
```
willcl-ark: In "support channel policies in LN edges":

Now that we have two script to generate these docs, might be worth adding a `just update-docs` helper script (in a followup or w/e)

src/schema/graph_schema.json
```diff
@@ -2,19 +2,40 @@
...
-      "tc_netem": {"type": "string"},
-      "exporter": {"type": "boolean", "default": false},
-      "collect_logs": {"type": "boolean", "default": false},
-      "build_args": {"type": "string", "default": ""},
-      "ln": {"type": "string"},
-      "ln-image": {"type": "string"},
-      "ln-cb-image": {"type": "string"}
+      "version": {
+        "type": "string",
+        "comment": "Bitcoin Core version with an available Warnet tank image on Dockerhub. May also be a GitHub repository with format user/repository:branch to build from source code"},
```
willcl-ark: Really like these comments! Didn't know that was an option but makes sense that it should be one.

src/schema/node_schema.json
```diff
@@ -0,0 +1,25 @@
+{
```
willcl-ark: In "LN: pass ln node config arguments in from graphml file " 4ad3016e6b5cb569fa9d05f75feca889f52014a7:

We're adding a _node_schema.json_ file back in in this commit after such a nice tidy up in previous commit?

willcl-ark (commented):
looking good my side. a few rebase fixups would be nice, but it does end in the right place overall :P

scripts/graphdocs.py
pinheadmz: Considered this but seems unlikely that both docs would need to be updated in the same commit?

src/schema/node_schema.json
pinheadmz: Yeah this is a rebase mistake with a fix up commit later. Really I just wanted to rebase on the march docs PR, let's merge that first anyway

willcl-ark:
You may want to cherry-pick #319 instead of "ci: force docker compose 2.24" (if it works)

PR #316, opened 2024-03-17
Update docs for march 2024

pinheadmz:
Currently includes basic cleanup of the markdown files and introduction page. Also updated a few RPC command help texts, revised `scripts/apidocs.py` and ran it.

TODO:
- [x] Add `edge_schema.json` and validate edges with it (important for LN channels)
- [x] Add a new script like `apidocs.py` that parses the schema files and automatically documents in `graph.md` including an example of all the `<key attr.name=...` headers that are needed in each graphml file
- [ ] Add another script that parses the justfile and automatically documents those, maybe in a new `kubernetes.md` file
- [ ] Finally, make sure our installation instructions are still accurate


pinheadmz:
marking as ready for now, will check installation and k8s docs in a future PR

PR #323, opened 2024-03-26
remove frontend

pinheadmz:
🧹


willcl-ark:
Wanted to do this ages ago. Only had one question I never bothered to find the answer to; will this destroy warnet.dev?

pinheadmz:
Hm I think it has something to do with Vercel, bc I don't see a CNAME or like a static github pages connection. @josibake owns the domain I think?

willcl-ark:
Shall we just try it and see?

ACK from me...

pinheadmz:
🤷

willcl-ark:
Still up for me™

pinheadmz:
Phew! I was definitely worried this breaking change would prevent our customers from finding the link to this repo. With pink tanks.

PR #324, opened 2024-03-26
add sim-ln service

pinheadmz:
This is a WIP, pushing commits slowly to bang on CI

Strategy:
- [x] Move the backend/compose/services directory up to src/
- [x] Add graphml options to add those services only when needed
- [x] refactor the BaseService to accomodate both backends
- [x] add sim-ln service that can be optionally added to any warnet with either backend
- [x] export network data into the simln container so it can do its thing

TODO:
 - [ ] DOCS (follow up PR)

### DIFF
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -110,6 +110,13 @@ class BackendInterface(ABC):
         raise NotImplementedError("This method should be overridden by child class")
 
     @abstractmethod
+    def get_lnnode_hostname(self, index: int) -> str:
+        """
+        Get the hostname assigned to a lnnode attached to a tank from the backend
+        """
+        raise NotImplementedError("This method should be overridden by child class")
+
+    @abstractmethod
     def wait_for_healthy_tanks(self, warnet, timeout=60) -> bool:
         """
         Wait for healthy status on all bitcoind nodes
--- a/src/backends/backend_interface.py
+++ b/src/backends/backend_interface.py
@@ -115,11 +115,3 @@ class BackendInterface(ABC):
         Wait for healthy status on all bitcoind nodes
         """
         raise NotImplementedError("This method should be overridden by child class")
-
-
-    @abstractmethod
-    def service_from_json(self, json: dict) -> dict:
-        """
-        Create a single container from a JSON object of settings
-        """
-        raise NotImplementedError("This method should be overridden by child class")
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -356,9 +356,12 @@ class ComposeBackend(BackendInterface):
                 "networks": [tank.network_name],
             }
 
+    def get_lnnode_hostname(self, index: int) -> str:
+        return self.get_container_name(index, ServiceType.LIGHTNING)
+
     def add_lnd_service(self, tank, compose):
         services = compose["services"]
-        ln_container_name = self.get_container_name(tank.index, ServiceType.LIGHTNING)
+        ln_container_name = self.get_lnnode_hostname(tank.index)
         ln_cb_container_name = self.get_container_name(tank.index, ServiceType.CIRCUITBREAKER)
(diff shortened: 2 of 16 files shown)


src/backends/backend_interface.py
```diff
@@ -109,9 +109,23 @@ def get_tank_ipv4(self, index: int) -> str:
         """
         raise NotImplementedError("This method should be overridden by child class")
 
+    @abstractmethod
+    def get_lnnode_hostname(self, index: int) -> str:
```
willcl-ark, line 113: A nice (future) improvement could be to combine this with the `get_tank_ipv4` function above, and have it return the hostname for tanks or lnnodes

scripts/graphdocs.py
```diff
@@ -16,38 +16,33 @@
 doc += "### GraphML file format and headers\n"
 doc += "```xml\n"
 doc += '<?xml version="1.0" encoding="UTF-8"?><graphml xmlns="http://graphml.graphdrawing.org/xmlns">\n'
-for name, details in graph_schema["node"]["properties"].items():
```
willcl-ark: Seeing as you've put so much work into this, we could (in the future) add a CI job which ran the script and checked to docs were up-to-date

src/backends/backend_interface.py
```diff
@@ -115,3 +115,11 @@ def wait_for_healthy_tanks(self, warnet, timeout=60) -> bool:
         Wait for healthy status on all bitcoind nodes
         """
         raise NotImplementedError("This method should be overridden by child class")
+
+
+    @abstractmethod
+    def service_from_json(self, json: object):
+        """
+        Create a single container from an object of settings
```
willcl-ark: In backends: add method to deploy generic service JSON 

This seems odd to label it "object", isn't it coming from a dict?

src/backends/compose/compose_backend.py
```diff
@@ -506,3 +506,22 @@ def wait_for_healthy_tanks(self, warnet, timeout=60) -> bool:
             raise Exception(f"Tanks did not reach healthy status in {timeout} seconds")
 
         return healthy
+
+    def service_from_json(self, obj) -> object:
```
willcl-ark: In backends: add method to deploy generic service JSON 

Also here it seems weird to return a generic object, why not dict again?
Also also, its fine as it's internal I'd guess, but there's no code handling if the service doesn't exist.

src/templates/rpc/Dockerfile_rpc
```diff
@@ -11,7 +11,7 @@ WORKDIR /root/warnet
 
 # Copy the source directory contents into the container
 COPY . /root/warnet
-RUN pip install .
+RUN pip install -e .
```
willcl-ark: I don't think we want this in the RPC Dockerfile here, can't we just restore the `MANIFEST.in` file and use that?

It seems useful to test that proper install works...

willcl-ark (commented):
General ACK.

Left a few comments scattered through the (numerous) PRs, nothing blocking really expect perhaps using `pip install -e .` in the RPC dockerfile?

Some nice cleanups interspersed with the new features, which I think is OK in this repo when making a change of this size, as there's little point in breaking this into multiple dependant PRs and waiting on all of them.

src/backends/backend_interface.py
pinheadmz: damn me and my years of javascript!

pinheadmz: addressed in https://github.com/bitcoin-dev-project/warnet/pull/333

src/backends/compose/compose_backend.py
pinheadmz: nit addressed in #333 

if the service doesnt exist, there would be nothing to pass to this function anyway so thats really the caller's problem

src/templates/rpc/Dockerfile_rpc
pinheadmz: will delete this

PR #325, opened 2024-03-27
re-enable fork observer on compose

willcl-ark:
Uses a new graph attribute to selectively enable fork observer on compose only (for now).


pinheadmz:
this may conflict with #324 I will push something soon and we can see

willcl-ark:
I'll go after you, happy to rebase

PR #327, opened 2024-03-29
Rust cli

willcl-ark:
- [x] Commands implemented except `graph validate` are working. Converting an xml graphml file to json seems to be a bit of a pain for validation
- [x] added a --rust_cli matrix to tests on rpc tests, which will (lightly) test the new tool
- [x] pretty print warnet table repr in `network start` command
- [ ] `quickxml-to-serde` pulls in a different xml lib than we use `xmltree`, could possibly use the same as them?
- [x] implement the `image` commands
- [ ] write some tests
- [ ] @pinheadmz slightly scary-looking `graph import-json` not implemented yet.

Notes:

- Graphs are generated with some kind of randomness in the tagging, which looks weird, but works fine.


willcl-ark:
I think if it made sense, we could perhaps strip out deps on jsonrpsee and tokio, and use a more rdimentary JSONRPC client, possibly something more like this: https://github.com/apoelstra/rust-jsonrpc

willcl-ark:
Closing for now

PR #331, opened 2024-04-03
Use built image in test

willcl-ark:
Also restructure CI into single workflow. This makes it much easier to depend on other workflows, and also use workflow artifacts between jobs.

Also, add another matrix to the workflow to make it DRY-er

Finally, re-add required MANIFEST.in

cc @pinheadmz


willcl-ark:
An example of the new CI workflow can be seen here: https://github.com/willcl-ark/warnet/actions/runs/8546726444

.github/workflows/test.yml
```diff
@@ -1,13 +1,67 @@
...
+      - uses: actions/checkout@v4
+      - uses: chartboost/ruff-action@v1
+  deploy:
+    needs: [ruff]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
pinheadmz: what happens on merge to main?

.github/workflows/test.yml
```diff
@@ -1,13 +1,67 @@
...
+      - name: Build and push
+        uses: docker/build-push-action@v5
+        with:
+          file: src/templates/rpc/Dockerfile_rpc
+          platforms: linux/amd64,linux/arm64
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=registry,ref=${{ secrets.DOCKERHUB_USERNAME }}/${{ secrets.DOCKERHUB_RPC_REPO }}:latest
```
pinheadmz, line 54: whats this do exactly? i just see cache and latest, wondering if that could pull an unexpected version or something

.github/workflows/test.yml
```diff
@@ -1,13 +1,67 @@
...
+          cache-to: type=inline
+      - name: Create tag name file
+        run: echo "${{ steps.meta.outputs.tags }}" > tagname.txt
```
pinheadmz, line 57: this has to be a file? it cant just be exported like the other variables `{{ stepa.meta.output.tags }}` etc?

.github/workflows/test.yml
willcl-ark: Ah my if must have gotten rebased out, nice catch

.github/workflows/test.yml
willcl-ark: https://docs.docker.com/build/ci/github-actions/cache/

Faster builds

.github/workflows/test.yml
willcl-ark: This is an artifact of when the jobs were in different workflows. It does not need a file now

pinheadmz:
> An example of the new CI workflow can be seen here: https://github.com/willcl-ark/warnet/actions/runs/8546726444

Looks awesome and i believe it is working as expected

[built image:](https://github.com/willcl-ark/warnet/actions/runs/8546726444/job/23417649314#step:7:727)

```
#18 pushing manifest for docker.io/***/***:latest@sha256:c8f4f4abd7dc92de48527396e8479dc4d8039c77626c5ebd6a4c24be27ea50ab
```

[image downloaded in test](https://github.com/willcl-ark/warnet/actions/runs/8546726444/job/23417718921#step:7:12):

```
Using tagname: will8clark/warnet-rpc:latest
```

Which I can see [on dockerhub](https://hub.docker.com/r/will8clark/warnet-rpc/tags):

```
sha256:c8f4f4abd7dc92de48527396e8479dc4d8039c77626c5ebd6a4c24be27ea50ab
```

My only concern is using the tagname instead of, for example

```
will8clarkwarnet-rpc@sha256:c8f4f4abd7dc92de48527396e8479dc4d8039c77626c5ebd6a4c24be27ea50ab
```

just in case there is some weird thing with force pushing to a branch multiple times or a cancelled workflow or something like that

willcl-ark:
I think this can go in if CI passes. I cannot pass env var between jobs despite an hours worth of attempts like [this](https://github.com/willcl-ark/warnet/commit/7f332994b37a81ef786d3fa6ed20272113ca2c8c).

willcl-ark:
OK a new example with (most of) @pinheadmz comments addressed: https://github.com/willcl-ark/warnet/actions/runs/8560281420

.github/workflows/test.yml
```diff
@@ -60,12 +60,13 @@ jobs:
...
           name: image-tags-${{ github.sha }}
           path: tagname.txt
-  scenarios:
+  test:
     needs: [deploy]
     runs-on: ubuntu-latest
     strategy:
       matrix:
         backend: [compose, k8s]
+        test: [scenarios_test.py, rpc_test.py, graph_test.py, ln_test.py]
```
pinheadmz, line 69: Ohhhhh.... duh yeah right nice

PR #335, opened 2024-04-09
graph: use pub_key over id in import-json

carlaKC:
LND graph files label this field pub_key, but we were previously generating sub-graphs with a library that output id instead. Updating this to perfectly fit lnd's describegraph output.

This commit also updates to use latest reduced versions of the graph in tests (LN 10/100) which have pub_key rather than ID.


pinheadmz:
ACK

PR #338, opened 2024-04-11
add: addrman-observer support

0xB10C:
Adds an [addrman-observer](https://github.com/0xb10c/addrman-observer) service and tests. While the addrman-observer front-end is still WIP, it should be usable for warnet.

On the frontend, you should be able to add `?url=<tank_id>` to load the addrman of that tank. The same thing should also work with the `fetch from URL` and tank names.

### DIFF
--- a/src/warnet/warnet.py
+++ b/src/warnet/warnet.py
@@ -12,7 +12,7 @@ import networkx
 import yaml
 from backends import ComposeBackend, KubernetesBackend
 from templates import TEMPLATES
-from warnet.services import FO_CONF_NAME, GRAFANA_PROVISIONING, PROM_CONF_NAME
+from warnet.services import AO_CONF_NAME, FO_CONF_NAME, GRAFANA_PROVISIONING, PROM_CONF_NAME
 from warnet.tank import Tank
 from warnet.utils import gen_config_dir, load_schema, validate_graph_schema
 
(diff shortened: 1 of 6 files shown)


0xB10C:
I just saw https://github.com/bitcoin-dev-project/warnet/pull/325, which changes the way the fork-observer config is written. Should this be used here too?

0xB10C:
Using `http://localhost:23005/?url=0` with `test/data/services.graphml` works, but the addrman is empty as there is only a single node in that graph.

src/warnet/warnet.py
```diff
@@ -181,8 +181,10 @@ def generate_deployment(self):
         self.container_interface.generate_deployment_file(self)
         if "forkobserver" in self.services:
             self.write_fork_observer_config()
+        if "addrmanobserver" in self.services:
+            self.write_addrman_observer_config()
         if "grafana" in self.services:
-            self.write_grafna_config()
+            self.write_grafana_config()
```
pinheadmz, line 187: thank you! ;-)

pinheadmz:
Had to zoom waaaaay in but there they are!

Screenshot 2024-06-04 at 2 43 17 PM

pinheadmz:
Ok this is great thanks

willcl-ark:
🥳

PR #343, opened 2024-04-18
k8s: name services with network

pinheadmz:
service names like "simln" were colliding when attempting to run multiple named networks in a cluster. 

Note we cant currently run multiple networks in docker because they will overlap on 100.x.x.x


pinheadmz:
## Running 3 LN networks in k8s

Screen Shot 2024-04-18 at 1 38 54 PM

m3dwards:
Tested with two networks, merging.

PR #347, opened 2024-04-23
Add NotFoundError to catch missing logging resources

m3dwards:
There are two ways this can fail, this PR now catches the second!

Most of the changes seem to be because the file has drifted away from being Ruff formatted.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -14,7 +14,7 @@ from kubernetes.client.models.v1_pod import V1Pod
 from kubernetes.client.models.v1_service import V1Service
 from kubernetes.client.rest import ApiException
 from kubernetes.dynamic import DynamicClient
-from kubernetes.dynamic.exceptions import ResourceNotFoundError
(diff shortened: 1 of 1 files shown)


src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -447,7 +463,7 @@ def remove_prometheus_service_monitors(self, tanks):
                     name=f"warnet-tank-{tank.index:06d}",
                     namespace=MAIN_NAMESPACE,
                 )
-            except ResourceNotFoundError:
+            except (ResourceNotFoundError, NotFoundError):
```
willcl-ark, line 466: Perhaps we should write the errors to the log for debugging? Or is that what we;re trying to cover up here :D

willcl-ark (approved):
LGTM!

src/backends/kubernetes/kubernetes_backend.py
m3dwards: They are swallowed because they are expected to happen when the logging stack isn't installed. It's not really an error just a lazy way to delete them.

PR #349, opened 2024-04-24
use `minikube start --mount` to mount filesystem

mplsgrant:
# Issue 

I can't mount my host's directory to minikube's `/mnt/src` directory.

# Cause

Our [current `minikube mount` approach](https://github.com/bitcoin-dev-project/warnet/blob/76fd33f14cd178591a6ef81e09555b0758c32a67/justfile#L19) requires the use of the 9p filesystem but  [debian cloud](https://lists.debian.org/debian-kernel/2022/06/msg00109.html) as well as [cent os and red hat](https://github.com/kubernetes/minikube/issues/8202) all have issues supporting 9p.

# Solution

Place `minikube start --mount --mount-string="$PWD:/mnt/src"` inside warnet's `justfile start` command.
 
Note: This fix requires that we run `minikube delete` on each invocation of `just start`. This is because docker has an issue with re-mounting filesystems, and the `minikube delete` approach to sidestepping this limitation [comes recommended](https://github.com/kubernetes/minikube/issues/11084).

# Rationale

* The current justfile already relies on minikube.
* It simplifies our mounting logic (no more pid management). 
* It passes CI.

### DIFF
--- a/justfile
+++ b/justfile
@@ -15,14 +15,11 @@ start:
     #!/usr/bin/env bash
     set -euxo pipefail
 
-    # Mount local source dir
-    minikube mount $PWD:/mnt/src > /tmp/minikube_mount.log 2>&1 &
+    if ! minikube delete; then
+        echo "Detected a fresh minikube environment."
+    fi
 
-    # Capture the PID of the minikube mount command
-    MINIKUBE_MOUNT_PID=$!
-
-    # Save the PID to a file for later use
-    echo $MINIKUBE_MOUNT_PID > /tmp/minikube_mount.pid
+    minikube start --mount --mount-string="$PWD:/mnt/src"
 
     # Setup k8s
     kubectl apply -f src/templates/rpc/namespace.yaml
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -15,14 +15,11 @@ start:
     #!/usr/bin/env bash
     set -euxo pipefail
 
-    # Mount local source dir
-    minikube mount $PWD:/mnt/src > /tmp/minikube_mount.log 2>&1 &
-
-    # Capture the PID of the minikube mount command
-    MINIKUBE_MOUNT_PID=$!
+    if ! minikube delete; then
+        echo "Detectd a fresh minikube environment."
```
willcl-ark:
```suggestion
        echo "Detected a fresh minikube environment."
```

willcl-ark (approved):
ACK.

This seems like a nice simplification, nice work and thanks!

willcl-ark:
Sorry I re-triggered CI to see if it will pass. I think it will if you rebase on main now (🤞🏼 )

PR #350, opened 2024-04-25
sort import to make ruff happy

mplsgrant:
# Issue

Current CI is failing because ruff is unhappy that some of our imports are unsorted.

# Solution

Perform a `ruff --fix` to sort the imports.

### DIFF
--- a/src/backends/kubernetes/kubernetes_backend.py
+++ b/src/backends/kubernetes/kubernetes_backend.py
@@ -14,7 +14,7 @@ from kubernetes.client.models.v1_pod import V1Pod
 from kubernetes.client.models.v1_service import V1Service
 from kubernetes.client.rest import ApiException
 from kubernetes.dynamic import DynamicClient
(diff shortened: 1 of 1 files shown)


willcl-ark:
Thanks for this. The change is correct, but I feel like it's highlighting that something in our CI is off; for example the PR I merged was [passing](https://github.com/bitcoin-dev-project/warnet/actions/runs/8803725239/job/24162507726) and this one is failing, which is the inverse of what should be happening...

Just trying to figure out why.

willcl-ark:
I'm wondering now if we need [these lines](https://github.com/bitcoin-dev-project/warnet/blob/1ffe97091a760dd4a0641c8af7b9bf41593b35d6/.github/workflows/test.yml#L21-L22) in the ruff checkout step too.

Alternatively we could extract Ruff into its own workflow and have it run on `pull_request`, as we don't need `pull_request_target` for it.

willcl-ark:
Going to merge this for now, and fix properly later.

PR #352, opened 2024-04-25
Possibly fix ruff workflow action not checking our correct hash

willcl-ark:
Seems like using "on: pull_request_target" makes the checkout action
miss unless this line is included.


willcl-ark:
Well, that does not appear to work either :(

willcl-ark:
Merging as this appears to work when in main branch in my fork

PR #355, opened 2024-05-09
Add CLN

m3dwards:
Adds CLN as a lightning node option

### DIFF
--- a/test/data/ln.graphml
+++ b/test/data/ln.graphml
@@ -39,21 +39,32 @@
     </node>
     <node id="3">
         <data key="version">27.0</data>
+        <data key="bitcoin_config">-uacomment=w2</data>
+        <data key="ln">cln</data>
+        <data key="ln_config">--cltv-delta=33</data>
+    </node>
+    <node id="4">
+        <data key="version">27.0</data>
         <data key="bitcoin_config">-uacomment=w3</data>
     </node>
     <edge id="1" source="0" target="1"></edge>
     <edge id="2" source="1" target="2"></edge>
     <edge id="3" source="2" target="3"></edge>
     <edge id="4" source="2" target="0"></edge>
-    <edge id="4" source="3" target="0"></edge>
+    <edge id="4" source="3" target="4"></edge>
+    <edge id="5" source="4" target="0"></edge>
     <!-- LN channels -->
-    <edge id="5" source="0" target="1">
+    <edge id="6" source="0" target="1">
         <data key="channel_open">--local_amt=100000</data>
         <data key="target_policy">--base_fee_msat=2200 --fee_rate_ppm=13 --time_lock_delta=20</data>
     </edge>
-    <edge id="6" source="1" target="2">
+    <edge id="7" source="1" target="2">
         <data key="channel_open">--local_amt=100000 --push_amt=50000</data>
         <data key="source_policy">--base_fee_msat=5500 --fee_rate_ppm=3 --time_lock_delta=40</data>
+    </edge>
+    <edge id="8" source="3" target="2">
+        <data key="channel_open">amount=100000 push_msat=50000000</data>
+        <data key="source_policy"> feebase=5500 feeppm=3</data>
     </edge>
   </graph>
 </graphml>
--- a/test/graph_test.py
+++ b/test/graph_test.py
@@ -7,7 +7,7 @@ import uuid
 from pathlib import Path
 
 from test_base import TestBase
-from warnet.utils import DEFAULT_TAG, channel_match
+from warnet.utils import DEFAULT_TAG
 
 graph_file_path = Path(os.path.dirname(__file__)) / "data" / "services.graphml"
 json_file_path = Path(os.path.dirname(__file__)) / "data" / "LN_10.json"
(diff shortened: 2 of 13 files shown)


willcl-ark (commented):
Thanks for this PR, just taken a brief surface-level skim.

It seems to me that we have ended up with `lnnode.py` just including a tonne of `if...else` statements to support the two implementations. If we ever decide to add LDK this will only get worse. Perhaps we should, before merging this, create an `LNNode` base class, and subclass `LndNode` and `CLNNode` as we did previously for the two backends?

That way we keep them clean, and make adding new nodes much easier in the future?

It would also let us abstract away things like: https://github.com/bitcoin-dev-project/warnet/pull/355/files#diff-0bfb3fbf6aa1bb068dffe115137cfec50240a3e8713463149f47d8e1354ceef1R100-R107 too...?

test/graph_test.py
```diff
@@ -68,20 +68,42 @@
...
+    )
+
+
+def policy_match(pol1, pol2):
+    return (
+        max(int(pol1["time_lock_delta"]), 18) == max(int(pol2["time_lock_delta"]), 18)
+        and max(int(pol1["min_htlc"]), 1) == max(int(pol2["min_htlc"]), 1)
+        and pol1["fee_base_msat"] == pol2["fee_base_msat"]
+        and pol1["fee_rate_milli_msat"] == pol2["fee_rate_milli_msat"]
+    )
```
pinheadmz, line 90: Aren't these defined in utils.py? Do they need to be stubbed here?

test/ln_test.py
```diff
@@ -97,7 +100,7 @@ def check_invoices(index):
 
 print("\nEngaging simln")
 activity = [{"source": "ln-0", "destination": node2pub, "interval_secs": 1, "amount_msat": 2000}]
-base.warcli(f"network export --exclude=[1] --activity={json.dumps(activity).replace(' ', '')}")
+base.warcli(f"network export --exclude=[1,3] --activity={json.dumps(activity).replace(' ', '')}")
```
pinheadmz: The `exclude` option is intended for the attackathon to make the target node forwarding-only: https://github.com/carlaKC/attackathon/issues/19

simln should support cln ok, is there something else Im missing?

test/data/ln.graphml
```diff
@@ -38,22 +38,33 @@
         <data key="ln_config">--bitcoin.timelockdelta=33</data>
     </node>
     <node id="3">
+        <data key="version">27.0</data>
+        <data key="bitcoin_config">-uacomment=w2</data>
+        <data key="ln">cln</data>
+        <data key="ln_config">--bitcoin.timelockdelta=33</data>
```
pinheadmz: this is an arg passed to the ln container -- does CLN use the same syntax as lnd for this?

pinheadmz (commented):
Reviewed and tested, really cool to see cln in there. I was able to get invoices and make payments in the 3 node graph. I agree with Will though, I think the implementations should be subclasses of `LNNode`. If that sounds like too much work for now we can address it in a follow-up. Functionally this PR works great for now.

I also wonder how we should handle `warcli lncli... ` commands. Like, we should we abstract the LN commands such that all lnd commands work for cln and ldk as well?

test/graph_test.py
m3dwards: I really hoped you wouldn't notice this but you are too smart for me.

The difference is that in in utils.py it now works with LNChannel objects which are returned from helper functions from LNNode. LNChannel is an abstraction over channels objects as returned by LND or CLN.

Why I couldn't use that in the test is that the test framework only uses the RPC which in this test is returning specific JSON from LND. I couldn't think of a quick solution and so added these helper functions here to keep the test passing. Ideally I think more and more things should move to use the abstractions and the CLI should provide functions for common lightning functions.

test/ln_test.py
m3dwards: Could you perhaps help me get the test passing without `,3`?

test/data/ln.graphml
m3dwards: Good spot, no it is not.

m3dwards:
> Perhaps we should, before merging this, create an `LNNode` base class, and subclass `LndNode` and `CLNNode` as we did previously for the two backends

Agreed. Much nicer.

m3dwards:
@willcl-ark @pinheadmz switched to lightning subclasses.

test/graph_test.py
```diff
@@ -68,20 +68,42 @@
...
+
+
 print("Ensuring warnet LN channel policies match imported JSON description")
 with open(json_file_path) as file:
     actual = json.loads(base.warcli("lncli 0 describegraph"))["edges"]
+    # json.dump(actual, file.writable(), indent=2)
```
willcl-ark: Accidental inclusion?

willcl-ark (approved):
Looks and seems to work great! nice work.

I left a single comment here, and also prepared 3 additional commits which you may be interested in taking here, or in a followup?

https://github.com/bitcoin-dev-project/warnet/compare/main...willcl-ark:warnet:add-cln-review

If you'd prefer followup, we can merge when @pinheadmz question about the `--exclude` option is resolved IMO

m3dwards:
> 3 additional commits which you may be interested in taking here

Taken

test/ln_test.py
m3dwards: Should be sorted now

pinheadmz:
#408

PR #363, opened 2024-05-14
remove libbitcoinconsensus

josibake:
file does not exist anymore, which leads to an error when using warcli build

### DIFF
--- a/src/templates/Dockerfile
+++ b/src/templates/Dockerfile
@@ -48,8 +48,8 @@ RUN set -ex \
     && make install \
     && strip ${BITCOIN_PREFIX}/bin/bitcoin-cli \
     && strip ${BITCOIN_PREFIX}/bin/bitcoind \
-    && strip ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.a \
-    && strip ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.so.0.0.0 \
+    && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.a \
+    && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.so.0.0.0 \
     && rm ${BITCOIN_PREFIX}/bin/bitcoin-tx \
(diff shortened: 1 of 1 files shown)


josibake:
Note: it occurred to me this file might still be present for older releases, but seems fine to keep it in there vs writing some conditional logic based on the version for remove it?

pinheadmz:
Is this about the IsReachable patch or just cleanup to keep the image small?

willcl-ark:
It will be better to simply `rm -f` (don't fail if don't exist) them, no?

```bash
    && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.a \
    && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.so.0.0.0 \
```

josibake:
> Is this about the IsReachable patch or just cleanup to keep the image small?

Keeping the image small

willcl-ark:
> It will be better to simply `rm -f` (don't fail if don't exist) them, no?
> 
> ```shell
>     && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.a \
>     && rm -f ${BITCOIN_PREFIX}/lib/libbitcoinconsensus.so.0.0.0 \
> ```

I'd prefer this approach personally, as I don't think skipping the `strip` call will actually remove the libs? They will still be copied across (but be larger in size!)

josibake:
> I'd prefer this approach personally, as I don't think skipping the strip call will actually remove the libs?

Do you think its worth changing all of the calls to `rm` , instead of `strip`?

willcl-ark:
No?

rm deletes it, strip removes debug symbols

josibake:
@willcl-ark updated to use `rm -f` instead of strip

josibake:
CI failing due to https://github.com/docker/docker-py/issues/3256

willcl-ark:
> CI failing due to [docker/docker-py#3256](https://github.com/docker/docker-py/issues/3256)

Well thats silly!

willcl-ark:
One more rebase on current `main` in here?

PR #365, opened 2024-05-22
Fix broken requirments, use `uv` to manage

willcl-ark:
Requirements should go in `requirements.in`, and then if updated, be dumped by `uv` (or other tools) using

```bash
uv pip compile --exclude-newer 2024-01-01 requirements.in -o requirements.txt
```

or equivalent commands.


willcl-ark:
Presumably this will fail CI, but let's find out!

PR #366, opened 2024-05-23
fix requirements

willcl-ark:
Ignoring everything else, I think fixing requirements like this will fix the issue with urllib3 (from #363) for now


willcl-ark:
Ok wtf. The only difference I can see now (vs master branch) is that this uses `pip` v24.0?

Also, with these requirements it works on my machine™️ 

I don't know what to do about this right now...

PR #368, opened 2024-05-24
reduce the goods by one

mplsgrant:
# Issue
Sometimes the `ln_init` routine fails with an error about insufficient funds.

# Cause
The `ln_init.py` process will fail due to insufficient funds if the number of coins it generates is evenly divisible relative to the number of recipients. This leaves the wallet without any money for fees.

# Solution
Set aside one bitcoin before dividing up the goods.


pinheadmz:
brilliant thanks!

PR #369, opened 2024-05-25
add quick_start

mplsgrant:
# Issue
New users have a hard time getting going with Warnet.

# Cause
Warnet has lots of dependencies that stand in the way of starting up warnet.

# Solution
Create a `quickstart` script that checks for dependencies and points the user in the right direction for meeting those dependencies. The `quickstart` should also fire up a basic 12 node default warnet.

- [x] minikube
- [x] kubectl
- [x] docker
- [x] docker permissions
- [x] just
- [x] python3
- [x] venv
- [x] BPF

### DIFF
--- a/quick_start.sh
+++ b/quick_start.sh
@@ -1,6 +1,19 @@
 #!/bin/bash
 set -euo pipefail
 
+
+is_cygwin_etal() {
+    uname -s | grep -qE "CYGWIN|MINGW|MSYS"
+}
+is_wsl() {
+    grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null
+}
+if is_cygwin_etal || is_wsl; then
+    echo "Quick start does not support Windows"
+    exit 1
+fi
+
+
 # Colors and styles
 RESET='\033[0m'
 BOLD='\033[1m'
@@ -69,8 +82,8 @@ docker_path=$(command -v docker || true)
 if [ -n "$docker_path" ]; then
     print_partial_message " ⭐️ Found " "docker" ": $docker_path" "$BOLD"
 else
-    print_partial_message " 💥 Could not find " "docker" ". Please follow this link to install it..." "$BOLD"
-    print_message "" "   https://minikube.sigs.k8s.io/docs/drivers/docker/" "$BOLD"
+    print_partial_message " 💥 Could not find " "docker" ". Please follow this link to install Docker Engine..." "$BOLD"
+    print_message "" "   https://docs.docker.com/engine/install/" "$BOLD"
     exit 127
 fi
 
@@ -128,9 +141,16 @@ print_message "" "" ""
 print_message "" "    Let's try to spin up a python virtual environment..." ""
 print_message "" "" ""
 
-python3 -m venv .venv
+if [ -d ".venv" ]; then
+    print_message "" "    It looks like a virtual environment already exists!" ""
+else
+    print_message "" "    Creating a new virtual environment..." ""
+    python3 -m venv .venv
+fi
+
 source .venv/bin/activate
 
+print_message "" "" ""
 print_partial_message " ⭐️ " "venv" ": The python virtual environment looks good" "$BOLD"
 print_message "" "" ""
 print_message "" "    Let's install warnet into that virtual environment..." ""
@@ -156,4 +176,4 @@ done
 print_message "" "🥳" ""
 print_message "" "Run the following command to enter into the python virtual environment..." ""
 print_message "" "    source .venv/bin/activate" "$BOLD"
-print_partial_message "   After that, you can run " "warcli" " to start running Warnet commands." "$BOLD"
+print_partial_message "   After that, you can run " "warcli help" " to start running Warnet commands." "$BOLD"
--- a/quick_start.sh
+++ b/quick_start.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+set -euo pipefail
+
+# Colors and styles
+RESET='\033[0m'
+BOLD='\033[1m'
+
+# Use colors if we can and have the color space
+if command -v tput &> /dev/null; then
+    ncolors=$(tput colors)
+    if [ -n "$ncolors" ] && [ "$ncolors" -ge 8 ]; then
+        RESET=$(tput sgr0)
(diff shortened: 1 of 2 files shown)


willcl-ark:
A few shellcheck suggestions:

```fish
will@ubuntu in ~/src/warnet on  2024-05-ez-startup [$?] : 🐍 (warnet)
₿ shellcheck quick_start.sh

In quick_start.sh line 42:
current_git=$(basename `git rev-parse --show-toplevel` || true)
                       ^-----------------------------^ SC2046 (warning): Quote this to prevent word splitting.
                       ^-----------------------------^ SC2006 (style): Use $(...) notation instead of legacy backticks `...`.

Did you mean:
current_git=$(basename $(git rev-parse --show-toplevel) || true)


In quick_start.sh line 113:
bpf_status=$(grep CONFIG_BPF /boot/config-$(uname -r) || true)
                                          ^---------^ SC2046 (warning): Quote this to prevent word splitting.


In quick_start.sh line 132:
source .venv/bin/activate
       ^----------------^ SC1091 (info): Not following: .venv/bin/activate was not specified as input (see shellcheck -x).

For more information:
  https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
  https://www.shellcheck.net/wiki/SC1091 -- Not following: .venv/bin/activate...
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

willcl-ark:
This is lush, love it:

image

mplsgrant:
Looks like the build test is failing. https://github.com/bitcoin-dev-project/warnet/actions/runs/9306448844/job/25616111319#step:4:2456

Will look into this.
 
Update: appears to be some kind of issue on  a 'remote providers' end:

````
build-test
Received request to deprovision: The request was cancelled by the remote provider.

````

Update 2: It passes on re-run. Maybe some kind of hang-up with a remote server (docker hub's?). I created an issue (#374) to keep an eye on it.

**Comment by mplsgrant** [comment]
> A few shellcheck suggestions:

@willcl-ark Solved. Thank you.

File: quick_start.sh [hunk 1]
```diff
@@ -0,0 +1,159 @@
...
+    print_message "" "   https://kubernetes.io/docs/tasks/tools/" "$BOLD"
+    exit 127
+fi
+
+docker_path=$(command -v docker || true)
+if [ -n "$docker_path" ]; then
+    print_partial_message " ⭐️ Found " "docker" ": $docker_path" "$BOLD"
+else
+    print_partial_message " 💥 Could not find " "docker" ". Please follow this link to install it..." "$BOLD"
+    print_message "" "   https://minikube.sigs.k8s.io/docs/drivers/docker/" "$BOLD"
```
willcl-ark: I don't particuarly mind linking to this page as it has a link to docker docs, but this could also just directly lilnk to https://docs.docker.com/engine/install/ ?

Also, should we recommend installing "Docker Enginer" and _not_ "Docker Desktop"? (probably not, I'm prob just a CLI snob...)

quick_start.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
+set -euo pipefail
+
```
willcl-ark, line 3: We could fail right away if OS==Windows? Might not be worth the difficulty though as a) we don't expect windows users, and b) I don't know how WSL reports e.g. the `$OSTYPE` variable...

Just thinking out loud really

willcl-ark:
```bash
if uname -s | grep -qE "CYGWIN|MINGW|MSYS"; then
    echo "Quick start does not support Windows"
    exit 1
fi
```

quick_start.sh
```diff
@@ -0,0 +1,159 @@
...
+else
+    print_partial_message " 💥 Could not find " "BPF" ". Please figure out how to enable Berkeley Packet Filters in your kernel." "$BOLD"
+    exit 1
+fi
+
+print_message "" "" ""
+print_message "" "    Let's try to spin up a python virtual environment..." ""
+print_message "" "" ""
+
+python3 -m venv .venv
```
willcl-ark: Could test for an existing venv here perhaps? that said, I don't think running this script actually _overwrote_ my existing venv; more testing needed...

quick_start.sh
```diff
@@ -0,0 +1,159 @@
...
+sleep 1
+warcli network start
+sleep 1
+while warcli network connected | grep -q "False"; do
+    sleep 2
+done
+print_message "" "🥳" ""
+print_message "" "Run the following command to enter into the python virtual environment..." ""
+print_message "" "    source .venv/bin/activate" "$BOLD"
+print_partial_message "   After that, you can run " "warcli" " to start running Warnet commands." "$BOLD"
```
willcl-ark:
```suggestion
print_partial_message "   After that, you can run " "warcli help" " to start running Warnet commands." "$BOLD"
```

willcl-ark (approved):
Looks great to me, nice work! Tested on Linux where everything was already installed, and macos where it failed midway. Didn't test each step, but code looks good to me.

Left a few comments with optional changes -- feel free to ignore them (all), or make changes you agree with. Let me know when you're done and I will hit that merge button!

quick_start.sh
mplsgrant: Added a few windows checks, but I have no easy way to test them. So, flying blind a little.

quick_start.sh
mplsgrant: Will do a check for .venv directory.

mplsgrant:
One thing I noticed is that `warcli network connected` gobbles up the output somehow. So, I can't do this..

````bash
while `warcli` network connected | grep -q "False"; do
    echo "Something reassuring for the user"
    sleep 2
done
````

mplsgrant:
> Let me know when you're done and I will hit that merge button!

@willcl-ark  I updated with your recommendations. 

What was up with macos failing midway?

willcl-ark:
> > Let me know when you're done and I will hit that merge button!
> 
> @willcl-ark I updated with your recommendations.

Sweet, LGTM!

> What was up with macos failing midway?

Oh, only that I don't have minikube on it, so got to see it "fail successfully"

Issue #36, opened 2023-08-15
extend test framework functions


pinheadmz:
anything pertaining to node must be extended into docker, examples:

`assert_debug_log()`
`restart_node()`
`add_nodes()`

also classes like `P2PInterface()` should be able to run locally but interact with deployed nodes

willcl-ark:
Is this still an issue?

pinheadmz:
yes a contributor trying to write a scenario using these or many other Test Framework methods is gonna have a bad time :-(

Nosh-thee-techy:
Hii, is this still an issue?

pinheadmz:
> Hii, is this still an issue?

Yes, do you understand what needs to be done and how it would benefit the warnet project and bitcoin more generally?

PR #371, opened 2024-05-27
add connect_nodes

mplsgrant:
# Issue
Running `connect_nodes` from the Test Framework fails.

# Cause
`connect_nodes` assumes bitcoind processes running via `popen`, and they don't expect to see dns-style addresses in place of ip addresses.

# Solution
Update the backend abstract class, and align the original `connect_nodes` function with the ip configurations we use in k8s.


src/warnet/test_framework_bridge.py
```diff
@@ -302,3 +310,69 @@ def parse_args(self):
...
+    def connect_nodes(self, a, b, *, peer_advertises_v2=None, wait_for_connect: bool = True):
+        """
+        Kwargs:
+            wait_for_connect: if True, block until the nodes are verified as connected. You might
+                want to disable this when using -stopatheight with one of the connected nodes,
+                since there will be a race between the actual connection and performing
+                the assertions before one node shuts down.
+        """
+        from_connection = self.nodes[a]
+        to_connection = self.nodes[b]
```
pinheadmz, line 322: the test framework in warnet also has `self.warnet` so you can get the tanks directly from there, and if there is `Tank.ipv4` or `Tank.get_external_ip()` working then the I think the rest of this will be a lot simpler.

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -561,6 +562,34 @@ def get_tank_ipv4(self, index: int) -> str:
         else:
             return None
 
+    def get_tank_dns_addr(self, index: int) -> str | None:
+        service_name = f"{self.network_name}-tank-{index:06d}-service"
```
pinheadmz: maybe use `self.get_service_name()` here? and below

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -561,6 +562,34 @@ def get_tank_ipv4(self, index: int) -> str:
...
+        try:
+            initial_subset = endpoints.subsets[0]
+        except IndexError:
+            self.log.info(f"{service_name}'s endpoint does not have an initial subset")
+            return None
+        try:
+            initial_address = initial_subset.addresses[0]
+        except IndexError:
+            self.log.info(f"{service_name}'s initial subset does not have an initial address")
+            return None
```
pinheadmz: I don't think try/except is the best choice here, I'd probably do something like

```python
if len(endpoints.subsets) < 1:
    raise Exception(...)
```

src/backends/kubernetes/kubernetes_backend.py
```diff
@@ -561,6 +562,34 @@ def get_tank_ipv4(self, index: int) -> str:
...
+            self.client.read_namespaced_service(name=service_name, namespace="warnet")
+        except ApiValueError as e:
+            self.log.info(ApiValueError(f"dns addr request for {service_name} raised {str(e)}"))
+            return None
+        return service_name
+
+    def get_tank_ip_addr(self, index: int) -> str | None:
+        service_name = f"{self.network_name}-tank-{index:06d}-service"
+        try:
+            endpoints = self.client.read_namespaced_endpoints(name=service_name, namespace="warnet")
```
pinheadmz: So this returns the IP address of the actual pods themselves which I don't think we need or even want anyone to need. If a pod crashes and K8s restarts it, it may have a different IP. The reason we use kubernetes services is to maintain a static IP for the bitcoin nodes even if they crash / restart.

src/scenarios/connect_dag.py
```diff
@@ -0,0 +1,115 @@
...
+        # | F  5 |  1 |   2 | D      | E, G    |
+        # | G  6 |  1 |   1 | F      | H       |
+        # | H  7 |  1 |   0 | G      | -       |
+        #
+        #          ╭──────> E                       ╭──────> 4
+        #          │        ∧                       │        ∧
+        #  A ─> C ─┤        │               0 ─> 2 ─┤        │
+        #       ∧  ╰─> D ─> F ─> G ─> H          ∧  ╰─> 3 ─> 5 ─> 6 ─> 7
+        #       │      ∧                         │      ∧
+        #  B ───┴──────╯                    1 ───┴──────╯
```
pinheadmz: great ascii art! I'm a little confused by having both letters and numbers for all the nodes ...?

src/warnet/tank.py
```diff
@@ -136,6 +136,13 @@ def status(self) -> RunningStatus:
     def exec(self, cmd: str):
         return self.warnet.container_interface.exec_run(self.index, ServiceType.BITCOIN, cmd=cmd)
 
+    def get_dns_addr(self) -> str:
+        dns_addr = self.warnet.container_interface.get_tank_dns_addr(self.index)
+        return dns_addr
+
+    def get_ip_addr(self) -> str:
+        ip_addr = self.warnet.container_interface.get_tank_ip_addr(self.index)
+        return ip_addr
```
pinheadmz, line 145: need a newline after this

src/backends/kubernetes/kubernetes_backend.py
pinheadmz: Update: ok I see you're using this to get the "from" IP in `connect_nodes()`. I think that makes sense then

pinheadmz (commented):
looking good! some comments / questions below

src/backends/kubernetes/kubernetes_backend.py
mplsgrant: I originally went with "dns" because I figured "service" was k8s specific, and I wanted to keep the language generic and still accurate. That said, I'm totally fine renaming it.

pinheadmz: sorry I meant, there already is a get service name function you can use instead of building the string yourself

src/scenarios/connect_dag.py
mplsgrant: Thanks!

The lettered version is there b/c I prefer seeing nodes labeled as letters. The numbered version is there b/c it aligns with how we describe nodes in the warnet naming/indexing scheme. I can remove the lettered version if it makes the documentation appear cluttered.

mplsgrant: I decided to keep the indices graph b/c it's way too useful for me when reading through the `connect_nodes` calls below it. I also kept the letter graph b/c a numbers-only table plus a numbers-only graph is unpleasant to read for me. I labeled indices graph as such so it appears less surprising.

pinheadmz:
this is good progress for #36

pinheadmz (approved):
ACK 5d834156ddab6656c68acf2754161046d31c647b

```
Hash: SHA256

ACK 5d834156ddab6656c68acf2754161046d31c647b
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmZiEoUACgkQ5+KYS2KJ
yTpV9g//U/3pTbZcECG9eixEhb10nRzhyDDYE7gbFN2dOdrqC9DbIBQUWODkIsBd
DkZxkxtFtObgO7jPU35torQ2FOMIgUAL8ZdnBG/5ANFrd5ef+fYmsoL+HFJ8b0H8
9dXpFkg1UzNnPGA/XIV2b/dg9ErRg+yxR7W3+3uf392mMPUn0DLwgL1x4HUsuzux
OYt736AATeWoQxMQjD/ReiCDfrTxwXJGbbhqcvb5sojdg1yBn4lsXemOWm8D8QPg
JHa9wOgDWAm4EGIDbXZTl0huN0xV5t9ogbUTRM+nT9cneUzDdvhBC1JiQTgqVfqw
KySGave/4EoVoIZtgYVBYUCZ3BD3bwA4nvbMkZOPvhet8T+9/EghxIimTcF2DyGz
NYeqHjGHD0Xb9QLY/vAYj35lyHZX4yi86tvxUSFa/iwEXmA1fi9BETzkqHXPjO9Z
hObAiVhMy7NuZs6UOqeKX02ixbMtTI0Q5dGs8tQyuSg2AKnpUZ9zl76uKNsh3QlF
nPiOXA/KLu64N4Q/W9vJA18rrGzfNbMDn58WzoFPMDFbbPsd+T2N/7/V5JGM1/Gp
X2YxUdlUTECiGacxnBOLrvATe+bHg2zpRtFwBjhjx73K1z9inC1nPsMMCFPuOCLB
wS9XH7yxIKydyfgXxbf6PWT5CIpEG9mbS/9M4eKOsM/aZtZsbXE=
=ghMN
-----END PGP SIGNATURE-----
```

pinheadmz's public key is [on keybase](https://keybase.io/pinheadmz/pgp_keys.asc?fingerprint=e61773cd6e01040e2f1bd78ce7e2984b6289c93a)
</details>

PR #372, opened 2024-05-27
add replacement cycling

mplsgrant:
# Draft
Want to make a CI test for this; waiting on #384 

# Issue
I can not run ariard's [replacement cycling test](https://github.com/bitcoin/bitcoin/commit/9ca92dbfccdc129a467da63f3398714ce06d7b5c#diff-d80e7c9dd2873bbb8e25c5161ca550dcea94da862229c475d9285ab0f07f0f23) in Warnet.

# Cause
A number of the test framework components available in Bitcoin Core have not been ported into Warnet.

# Solution
Add the replacement cycling test as a scenario.

# Wins
The test runs without meaningful augmentation. This is great because it points us in a direction of copy/pasting tests that run today in Core's functional test framework. We can then augment the tests to work in realistic environments with lightning nodes and other future tank types. 

# Depends on
#371

PR #373, opened 2024-05-27
add anticycle

mplsgrant:
# Draft
Want to make a CI test for this, waiting on #384 
# Issue
Can not run any defenses against the replacement cycling attack.

# Cause
We do not have any defense scenarios implemented, and we don't have zmq connected in a way that we can monitor the attack via zmq mempool data streams.

# Solution
Bring in the instagibbs [anticycle countermeasure](https://github.com/instagibbs/anticycle), and link up zmq appropriately so we can get the correct data for running the defense

# Depends on 
#370 
#371 
#372 
#384

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -12,3 +12,4 @@ requests<2.30
 rich
 tabulate
 PyYAML
+pyzmq
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,11 +15,17 @@ certifi==2023.11.17
 charset-normalizer==3.3.2
     # via requests
 click==8.1.7
-    # via flask
+    # via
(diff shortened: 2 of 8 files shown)


mplsgrant:
Updated anticycle by delinting and introducing some quality of life improvements. Also, merged most recent replacement_cycle changes into this PR. 

Anticycle should be run against the replacement_cycle scenario, which is why I bothered to include the replacement_cycling scenario in this PR.

```
warcli scenarios run anticycle
warcli scenarios run replacement_cycling

kubectl logs pod/rpc-0
```

mplsgrant:
Rebased on main

PR #375, opened 2024-06-03
CI housekeeping

willcl-ark:
Splits our single workflow into two. This means that we can use `on: pull_request`, which works much better™️  with other actions (like checkout).

It also means (I think) that PRs which themselves modify the workflow(s), should be exercised _in the PR_ (unconfirmed, but I think it should work like that), which grant has been running into a few times recently...

The rationale for `on: pull_request_target` was that the github token given to `on: pull_request` does not have enough permissions to read repo secrets and things, so it could not deploy. This is not so with `pull_request_target`.

By splitting deploy into `on: workflow`, the token its natively given has the appropriate permissions.

The image from the `build` job is shared with the `test` job in the same workflow as an artifact. It is not shared directly (e.g. as an artifact) with the `deploy` workflow, but the build steps use a shared cache (see `cache_from` and `cache_to`), which achieves basically the same result.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,60 @@
+name: deploy
+
+on:
+  workflow_run:
+    workflows: ["test"]
+    types:
+      - completed
+
+jobs:
+  deploy-to-dockerhub:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


willcl-ark:
I have two additional commits [here](https://github.com/willcl-ark/warnet/pull/11) which we can also look to add in the future if wanted

mplsgrant:
I tried this out on my branch. [It seemed to work](https://github.com/mplsgrant/warnet/actions/runs/9359090148), including against my new `get_service_ip` test.

willcl-ark:
Going to merge this now then, I think it's working too, and should be quite a big improvement to be _actually_ testing the PR branch code!

pinheadmz:
looks like the CI on this PR ran twice (`pull_request` and `pull_request_target`?) but dont see that on the merge to main

PR #376, opened 2024-06-05
templates: fix addrman patch

pinheadmz:
The current addrman patch results in an addrman that buckets ipv4 address by 32 bits instead of 16. The issue is that in `NetGroupManager::GetGroup()` we are dealing with [IPv4 mapped to IPv6](https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses) addresses and so the 32 bits the function actually processes are the null prefix in the embedding scheme. This means all ipv4 addresses in warnet always end up in the same bucket :-P

The solution in this PR ensures that we process the *last* 32 bits of the embedded ipv6 structure which is where the actual ipv4 address really lives.

The issue is illustrated really well in the debugger while stopped in `GetGroup()`:

```
(lldb) print address
(const CNetAddr &) 0x0000000170028e50: {
  m_addr = {
    _union = {
      direct = "d\U00000014\U0000001e("
      indirect_contents = (indirect = "", capacity = 0)
    }
    _size = 4
  }
  m_net = NET_IPV4
  m_scope_id = 0
}

(lldb) print address.GetLinkedIPv4()
(uint32_t) 1679040040

(lldb) print address.GetAddrBytes()
(std::vector<unsigned char>) size=16 {
  [0] = '\0'
  [1] = '\0'
  [2] = '\0'
  [3] = '\0'
  [4] = '\0'
  [5] = '\0'
  [6] = '\0'
  [7] = '\0'
  [8] = '\0'
  [9] = '\0'
  [10] = '\xff'
  [11] = '\xff'
  [12] = 'd'
  [13] = '\x14'
  [14] = '\x1e'
  [15] = '('
}

```


pinheadmz:
h/t to @0xB10C and #338 which brought this issue to light!

willcl-ark:
Dang, that's a nice bugfix fren!

Patch LGTM.

We'll want to rebuild the images after this one. I'll kick them all off in a bit on my machine, and then merge if they all complete?

pinheadmz:
Just realized I made a mistake as well will update....

src/templates/addrman.patch
```diff
@@ -1,17 +1,15 @@
...
      } else if (address.HasLinkedIPv4()) {
--        // IPv4 addresses (and mapped IPv4 addresses) use /16 groups
--        uint32_t ipv4 = address.GetLinkedIPv4();
--        vchRet.push_back((ipv4 >> 24) & 0xFF);
--        vchRet.push_back((ipv4 >> 16) & 0xFF);
--        return vchRet;
-+        nBits = 32;
+         // IPv4 addresses (and mapped IPv4 addresses) use /16 groups
+         uint32_t ipv4 = address.GetLinkedIPv4();
++        vchRet.push_back((ipv4 >> 32) & 0xFF);
```
pinheadmz: This will be one of the \0 bytes whoops...0

pinheadmz:
Fixed my mistake and added an addrman unit test for reviewers

pinheadmz:
added a test that pulls the new v27.0 image and checks addrman buckets. also had to bump lnd version: https://x.com/roasbeef/status/1781086945986404559

willcl-ark (approved):
ACK fa0e3836f820ffd47a5be9439404c35439145567

tested locally, and it works nicely

PR #379, opened 2024-06-07
clean up backend types and return values

mplsgrant:
### Issue
Getting type errors when running PyCharm's automated checker. Some related to #371.

### Cause
The inconsistent types and return values exist but do not trigger any tests.

### Solution
Update the types and return values as appropriate.

### DIFF
--- a/src/backends/compose/compose_backend.py
+++ b/src/backends/compose/compose_backend.py
@@ -109,7 +109,7 @@ class ComposeBackend(BackendInterface):
             case _:
                 raise Exception("Unsupported service type")
 
-    def get_container(self, tank_index: int, service: ServiceType) -> Container:
+    def get_container(self, tank_index: int, service: ServiceType) -> Container | None:
         container_name = self.get_container_name(tank_index, service)
         if len(self.client.containers.list(filters={"name": container_name})) == 0:
             return None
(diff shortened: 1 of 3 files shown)


src/backends/compose/compose_backend.py
```diff
@@ -177,18 +177,18 @@ def get_file(self, tank_index: int, service: ServiceType, file_path: str):
...
 
-    def get_tank_ip_addr(self, index: int) -> str:
-        self.get_tank_ipv4(index)
+    def get_tank_ip_addr(self, index: int) -> str | None:
+        return self.get_tank_ipv4(index)
 
-    def get_tank_dns_addr(self, index: int) -> str:
-        self.get_tank_ip_addr(index)
+    def get_tank_dns_addr(self, index: int) -> str | None:
+        return self.get_tank_ip_addr(index)
```
pinheadmz, line 191: crap how'd we miss this? how did this pass the tests?

pinheadmz (commented):
LGTM, should we add PyCharm to our lint tests?

src/backends/compose/compose_backend.py
mplsgrant: The tests never call that function!

The other thing I noticed was that the dag_connect test does not call `debug_assert_log`. I think that means that as long as the test runs, it passes. Need to stitch the tests up.

mplsgrant:
> LGTM, should we add PyCharm to our lint tests?

I think so. I only included the lints from the backend. There's quite a number of other lints to clean up that PyCharm warned about.

willcl-ark:
> I think so. I only included the lints from the backend. There's quite a number of other lints to clean up that PyCharm warned about.

I'd be OK with this, but how does it work? I don't have that IDE and don't think you can get it's internal linter seperately?

I've also been meaning to add `ruff format` to the lints, so we always have formatted code...

willcl-ark:
LGTM too, thanks!

mplsgrant:
> I've also been meaning to add `ruff format` to the lints, so we always have formatted code...

@willcl-ark That seems like a great idea. I had imagined that `ruff` would be as picky as clippy right out of the box. But, I guess not!

> I'd be OK with this, but how does it work?

PyCharm has a "Try Locally" button I can press that allegedly allows me to run something called [Qodana](https://www.jetbrains.com/qodana/). When I press the button it examines everything in my local warnet project and lints it.

PR #380, opened 2024-06-10
add helm to quick_start

mplsgrant:
### Issue
Logging does not work when using `quick_start.sh`

### Cause
`quick_start.sh` does not help the user install `helm`

### Solution
Add `helm` to `quick_start.sh`


willcl-ark:
Looks good to me!

PR #381, opened 2024-06-12
compose: warn on attempt to overwrite existing network

### LINKED ISSUE #134: "Pool overlaps with other one on this address space"
You'll see this error if you try to start a new network, i.e. maybe you ran a network named "warnet" previously but it has been shut down in some way that is NOT `docker-compose down`. Even though the containers may be stopped and deleted, the network may still exist and that occupies our subnet `100....`.

willcl-ark:
Closes: #134


pinheadmz:
close bc compose?

PR #384, opened 2024-06-24
Allow asserting againt TestBase log file

mplsgrant:
### Issue
I want to use an `assert_log` function against the output of rpc-0. Specifically, I want to assert the existence of some text that is produced by a scenario and logged by rpc-0. I don't believe that is possible today.

Also, when scenarios fail today, that failure [does not bubble up](https://github.com/mplsgrant/warnet/actions/runs/9654531220/job/26628947670#step:8:333) to the test framework.

### Cause
- The `server_thread` in `TestBase` prints the log output of rpc-0, but it does not store that information anywhere in a way that allows tests to access it.
- The current `server` in `TestBase` reads the entire history of `kubectl log -f pod/rpc-0` which causes assert statements in test cases to incorrectly check old/stale log entries.

### Solution
Allow `TestBase` to examine the output of the kubectl process log and create assertions against it. Limit that log output to the current run by preventing kubectl from logging old messages by using the `--since` flag.

### DIFF
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -36,14 +36,6 @@ class DAGConnectionTest(TestBase):
         self.log_unexpected_msgs = ["Test failed."]
         self.wait_for_all_scenarios()
         self.assert_log_msgs_passed()
-        self.stop_running_scenario()
-
-    def stop_running_scenario(self):
-        running_scenarios = self.rpc("scenarios_list_running")
-        if running_scenarios:
-            pid = running_scenarios[0]["pid"]
-            self.log.warning(f"Stopping scenario with PID {pid}")
-            self.warcli(f"scenarios stop {pid}", False)
 
 
 if __name__ == "__main__":
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -30,19 +30,19 @@ class DAGConnectionTest(TestBase):
 
     def run_connect_dag_scenario(self):
         self.log.info("Running connect_dag scenario")
-        self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
         self.log_expected_msgs = [
             "Successfully ran the connect_dag.py scenario using a temporary file"
         ]
         self.log_unexpected_msgs = ["Test failed."]
+        self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
         self.wait_for_all_scenarios()
         self.assert_log_msgs()
 
     def run_connect_dag_scenario_post_connection(self):
         self.log.info("Running connect_dag scenario")
-        self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
         self.log_expected_msgs = ["Successfully ran the connect_dag.py scenario"]
         self.log_unexpected_msgs = ["Test failed"]
+        self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
         self.wait_for_all_scenarios()
         self.assert_log_msgs()
 
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 import os
-import time
 from pathlib import Path
 
 from test_base import TestBase
@@ -13,7 +12,6 @@ class DAGConnectionTest(TestBase):
         self.graph_file_path = (
             Path(os.path.dirname(__file__)) / "data" / "ten_semi_unconnected.graphml"
         )
-        self.scenario_timeout = 180  # seconds
 
     def run_test(self):
         self.start_server()
@@ -32,23 +30,13 @@ class DAGConnectionTest(TestBase):
     def run_connect_dag_scenario(self):
         self.log.info("Running connect_dag scenario")
         self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
-
-        start_time = time.time()
-        while time.time() - start_time < self.scenario_timeout:
-            running_scenarios = self.rpc("scenarios_list_running")
-            if not running_scenarios:
-                self.log.info("Scenario completed successfully")
-                return
-
-            if len(running_scenarios) == 1 and not running_scenarios[0]["active"]:
-                self.log.info("Scenario completed successfully")
-                return
-
-            time.sleep(1)
-
-        self.log.error(f"Scenario did not complete within {self.scenario_timeout} seconds")
+        self.log_expected_msgs = [
+            "Successfully ran the connect_dag.py scenario using a temporary file"
+        ]
+        self.log_unexpected_msgs = ["Test failed."]
+        self.wait_for_all_scenarios()
+        self.assert_log_msgs_passed()
         self.stop_running_scenario()
-        raise AssertionError(f"Scenario timed out after {self.scenario_timeout} seconds")
 
     def stop_running_scenario(self):
         running_scenarios = self.rpc("scenarios_list_running")
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -36,7 +36,7 @@ class DAGConnectionTest(TestBase):
         ]
         self.log_unexpected_msgs = ["Test failed."]
         self.wait_for_all_scenarios()
-        self.assert_log_msgs_passed()
+        self.assert_log_msgs()
 
     def run_connect_dag_scenario_post_connection(self):
         self.log.info("Running connect_dag scenario")
@@ -44,7 +44,7 @@ class DAGConnectionTest(TestBase):
         self.log_expected_msgs = ["Successfully ran the connect_dag.py scenario"]
         self.log_unexpected_msgs = ["Test failed"]
         self.wait_for_all_scenarios()
-        self.assert_log_msgs_passed()
+        self.assert_log_msgs()
 
 
 if __name__ == "__main__":
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -18,6 +18,7 @@ class DAGConnectionTest(TestBase):
         try:
             self.setup_network()
             self.run_connect_dag_scenario()
+            self.run_connect_dag_scenario_post_connection()
         finally:
             self.stop_server()
 
@@ -34,6 +35,14 @@ class DAGConnectionTest(TestBase):
             "Successfully ran the connect_dag.py scenario using a temporary file"
         ]
         self.log_unexpected_msgs = ["Test failed."]
+        self.wait_for_all_scenarios()
+        self.assert_log_msgs_passed()
+
+    def run_connect_dag_scenario_post_connection(self):
+        self.log.info("Running connect_dag scenario")
+        self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
+        self.log_expected_msgs = ["Successfully ran the connect_dag.py scenario"]
+        self.log_unexpected_msgs = ["Test failed"]
         self.wait_for_all_scenarios()
         self.assert_log_msgs_passed()
 
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -112,7 +112,7 @@ class TestBase:
         # TODO: check for conflicting warnet process
         #       maybe also ensure that no conflicting docker networks exist
 
-        # For kubernetes we assume the server is started outside test base
+        # For kubernetes we assume the server is started outside test base,
         # but we can still read its log output
         self.log.info("Starting Warnet server")
         self.server = Popen(
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -64,11 +64,11 @@ class TestBase:
             self.server = None
 
     def _print_and_assert_msgs(self, message):
+        print(message)
         if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
             message, self.log_expected_msgs, self.log_unexpected_msgs
         ):
             self.log_msg_assertions_passed = True
-        print(message)
 
     def assert_log_msgs(self):
         assert (
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -114,7 +114,7 @@ class TestBase:
         # but we can still read its log output
         self.log.info("Starting Warnet server")
         self.server = Popen(
-            ["kubectl", "logs", "-f", "rpc-0"],
+            ["kubectl", "logs", "-f", "rpc-0", "--since=1s"],
             stdout=PIPE,
             stderr=PIPE,
             bufsize=1,
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -53,7 +53,7 @@ class TestBase:
             self.log_msg_assertions_passed = True
         print(message)
 
-    def assert_log_msgs_passed(self):
+    def assert_log_msgs(self):
         assert (
             self.log_msg_assertions_passed
         ), f"Log assertion failed. Expected message not found: {self.log_expected_msgs}"
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -46,19 +46,6 @@ class TestBase:
         self.log = logging.getLogger("test")
         self.log.info("Logging started")
 
-    def _print_and_assert_msgs(self, message):
-        if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
-            message, self.log_expected_msgs, self.log_unexpected_msgs
-        ):
-            self.log_msg_assertions_passed = True
-        print(message)
-
-    def assert_log_msgs(self):
-        assert (
-            self.log_msg_assertions_passed
-        ), f"Log assertion failed. Expected message not found: {self.log_expected_msgs}"
-        self.log_msg_assertions_passed = False
-
     def cleanup(self, signum=None, frame=None):
         if self.server is None:
             return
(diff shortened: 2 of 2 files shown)


pinheadmz:
Warnet RPC server itself logs to `~/.warnet/warnet.log` inside the rpc-0 pod during tests and that includes all scenario output. So I think if you want to assert a log message in a test, that would be the file to search. You mine as well include a log assertion in some existing test for this PR as well so we can see it working.

However, I see what you mean about reading old log messages. For that reason, it might make sense for us to move the warnet server log into the network-specific subfolder (there will be a new one for each test run e.g. `~/.warnet/warnet/warnet-test-rhyjpwjn/` etc). So like, we wouldn't even write any log file at all until `network start`, and then we start logging right after the folder is created in `server.py` `generate_deployment()`.

And finally yes the addition of `--since=1s` is smart, I don't think I realized each new test run printed the entire container's life history every time!

test/test_base.py
```diff
@@ -84,10 +84,22 @@ def start_server(self):
         # TODO: check for conflicting warnet process
         #       maybe also ensure that no conflicting docker networks exist
 
+        def write_and_print(line):
+            path = self.tmpdir / "tmp.log"
```
pinheadmz: no need to define this every time a log line is emitted. lets just define `self.testlog` when we instantiate the TestBase class, and open the file as a stream and write to it like a real log file. Also maybe call the file `testbase.log` instead of `tmp.log` ?

mplsgrant: We also have self.logfilepath which more or less is going unused.

pinheadmz: actually looks like that is going entirely unused, it looks like that is supposed to be the log file written by the rpc server, which I guess would have worked with docker because the server runs locally.

mplsgrant: I'm going to remove `self.logfilepath`.

pinheadmz:
@mplsgrant I think #385 is ready for merge, if youre done with it lets merge that then rebase this one

mplsgrant:
This has been rebased.

test/test_base.py
```diff
@@ -84,6 +84,18 @@ def start_server(self):
...
+            if path.exists():
+                with open(path, 'a') as file:
```
willcl-ark: Instead of this for every line to be logged, why not use a python logger configured to log to stdout and a file? Something like:

```python
import logging
import sys

log_file = "rpc-0.log"
# Create logger
logger = logging.getLogger('rpc-0-logger')
logger.setLevel(logging.DEBUG)

# Create file handler
file_handler = logging.FileHandler(log_file)
file_handler.setLevel(logging.DEBUG)

# Create console handler
console_handler = logging.StreamHandler(sys.stdout)
console_handler.setLevel(logging.DEBUG)

# Create formatter
# Dunno what format our other logs use, don't think it's needed here in any case as we should use log config from src/logging_config/config.json automagically
# formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')

# Set formatter for handlers
file_handler.setFormatter(formatter)
console_handler.setFormatter(formatter)

# Add handlers to logger
logger.addHandler(file_handler)
logger.addHandler(console_handler)

```

test/test_base.py
```diff
@@ -86,14 +87,13 @@ def start_server(self):
         #       maybe also ensure that no conflicting docker networks exist
 
         def write_and_print(line):
-            path = self.tmpdir / "tmp.log"
-            if path.exists():
-                with open(path, 'a') as file:
+            if self.testlog.exists():
```
willcl-ark: If we stick with non-python `logging` module logging, may as well backport these changes to the first commit?

willcl-ark (commented):
Concept ACK from me.

Just left a few comments/questions on first pass over

test/test_base.py
mplsgrant: I like this idea. 

Also, I turned the string "rpc-0" into an attribute and plugged it into the logger and into the call to kubectl.

test/test_base.py
mplsgrant: I think we should go with the logging approach.

mplsgrant:
Assuming we go with this logging approach, I want to go in an replace the existing print statements with calls to logger.

mplsgrant:
I switched the logger config to warnet's established config json file. 

Also, converted all the existing print statements into logger.info.

willcl-ark:
Great! I'd like to re-review and try to merge this before my giant refactor, I think... (also you reminded me that I should use our loggingconfig in my pr, probably)

willcl-ark:
I think in addition to the conflict with (newly reformatted) `dag_connection_test.py`, this may now conflict with the two final commits in https://github.com/bitcoin-dev-project/warnet/pull/400/commits which fixup the logging in test_base.py, I think in the same way (roughly) being done here.

I also slightly changed the entire log output format: https://github.com/bitcoin-dev-project/warnet/actions/runs/9846784937/job/27185339346#step:8:133 but I think that will actually not be an issue for your regex?

Perhaps you can drop your first three commits here, if that one gets merged first (or visa versa)? Not checked thoroughly yet...

mplsgrant:
@willcl-ark I just rebased on main, and I am willing to rebase on main again if you get #400 merged. I'm not sure that I want to break up this PR. 

> I also slightly changed the entire log output format: https://github.com/bitcoin-dev-project/warnet/actions/runs/9846784937/job/27185339346#step:8:133 but I think that will actually not be an issue for your regex?

I don't think this one will be an issue.

mplsgrant:
This is rebased/refactored and ready to go.

test/test_base.py
```diff
@@ -176,3 +194,26 @@ def get_scenario_return_code(self, scenario_name):
...
+    assert_equal(type(expected_msgs), list)
+    assert_equal(type(unexpected_msgs), list)
+
+    found = True
+    for unexpected_msg in unexpected_msgs:
+        if re.search(re.escape(unexpected_msg), log_message, flags=re.MULTILINE):
+            raise AssertionError(f"Unexpected message found in log: {unexpected_msg}")
+    for expected_msg in expected_msgs:
+        if re.search(re.escape(expected_msg), log_message, flags=re.MULTILINE) is None:
+            found = False
```
pinheadmz, line 217: I was able to modify the expected log line in the test but it still passed:

```diff
diff --git a/test/dag_connection_test.py b/test/dag_connection_test.py
index 6e87b79..1b5ee0f 100755
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -31,7 +31,7 @@ class DAGConnectionTest(TestBase):
         self.log.info("Running connect_dag scenario")
         self.warcli("scenarios run-file test/framework_tests/connect_dag.py")
         self.log_expected_msgs = [
-            "Successfully ran the connect_dag.py scenario using a temporary file"
+            "My spleen hurts"
         ]
         self.log_unexpected_msgs = ["Test failed."]
         self.wait_for_all_scenarios()

```

test/test_base.py
```diff
@@ -176,3 +194,26 @@ def get_scenario_return_code(self, scenario_name):
...
+def assert_log(log_message, expected_msgs, unexpected_msgs=None) -> bool:
+    if unexpected_msgs is None:
+        unexpected_msgs = []
+    assert_equal(type(expected_msgs), list)
+    assert_equal(type(unexpected_msgs), list)
+
+    found = True
+    for unexpected_msg in unexpected_msgs:
+        if re.search(re.escape(unexpected_msg), log_message, flags=re.MULTILINE):
+            raise AssertionError(f"Unexpected message found in log: {unexpected_msg}")
```
pinheadmz, line 214: I was able to sneak an unexpected message into the log without test failure:

```diff
diff --git a/test/dag_connection_test.py b/test/dag_connection_test.py
index 6e87b79..b5df653 100755
--- a/test/dag_connection_test.py
+++ b/test/dag_connection_test.py
@@ -33,7 +33,7 @@ class DAGConnectionTest(TestBase):
         self.log_expected_msgs = [
             "Successfully ran the connect_dag.py scenario using a temporary file"
         ]
-        self.log_unexpected_msgs = ["Test failed."]
+        self.log_unexpected_msgs = ["Successfully ran the connect_dag.py scenario using a temporary file"]
         self.wait_for_all_scenarios()
         self.assert_log_msgs_passed()
 
```

test/dag_connection_test.py
```diff
@@ -32,30 +30,12 @@ def setup_network(self):
...
-        raise AssertionError(f"Scenario timed out after {self.scenario_timeout} seconds")
-
-    def stop_running_scenario(self):
-        running_scenarios = self.rpc("scenarios_list_running")
-        if running_scenarios:
-            pid = running_scenarios[0]["pid"]
-            self.log.warning(f"Stopping scenario with PID {pid}")
-            self.warcli(f"scenarios stop {pid}", False)
+        self.log_expected_msgs = [
+            "Successfully ran the connect_dag.py scenario using a temporary file"
```
pinheadmz, line 34: Of course, now that https://github.com/bitcoin-dev-project/warnet/pull/406 is merged, we have an easier way to do this :-P

pinheadmz (commented):
I'm not convinced this is working... 🤔

test/test_base.py
mplsgrant: Thanks. Pushed a fix commit.

test/dag_connection_test.py
mplsgrant: True, but I think it will be helpful to be able to assert for specific things in our logs

mplsgrant:
I also re-added "since" to clear the messages.

test/test_base.py
```diff
@@ -42,6 +46,18 @@ def setup_logging(self):
...
+        if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
+            message, self.log_expected_msgs, self.log_unexpected_msgs
+        ):
+            self.log_msg_assertions_passed = True
+        print(message)
+
+    def assert_log_msgs_passed(self):
+        assert (
+            self.log_msg_assertions_passed
+        ), f"Log assertion failed: {self.log_expected_msgs}, {self.log_unexpected_msgs}"
```
pinheadmz: since unexpected messages get their own error and log message on L213 I think you can be more explicit here:

```suggestion
        ), f"Log assertion failed. Expected message not found: {self.log_expected_msgs}"
```

test/test_base.py
```diff
@@ -42,6 +46,18 @@ def setup_logging(self):
         self.log = logging.getLogger("test")
         self.log.info("Logging started")
 
+    def print_and_assert_msgs(self, message):
+        if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
+            message, self.log_expected_msgs, self.log_unexpected_msgs
+        ):
+            self.log_msg_assertions_passed = True
```
pinheadmz: Does this flag need to get reset ever? What if a test uses assert_log a few times with the same test_base instance?

mplsgrant: I imagined users setting `self.log_expected_msgs` and `self.log_unexpected_msgs` once per TestBase instance.

mplsgrant: We could reset the assertions status after checking. I updated with that concept. Need to test it.

test/test_base.py
```diff
@@ -59,6 +63,19 @@ def cleanup(self, signum=None, frame=None):
             self.server_thread.join()
             self.server = None
 
+    def _print_and_assert_msgs(self, message):
+        if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
+            message, self.log_expected_msgs, self.log_unexpected_msgs
+        ):
+            self.log_msg_assertions_passed = True
+        print(message)
```
pinheadmz: I think we should print the message first so if `assert_log` throws the unexpected message is still there in the printout.

test/test_base.py
```diff
@@ -59,6 +63,19 @@ def cleanup(self, signum=None, frame=None):
...
+
+    def assert_log_msgs(self):
+        assert (
+            self.log_msg_assertions_passed
+        ), f"Log assertion failed. Expected message not found: {self.log_expected_msgs}"
+        self.log_msg_assertions_passed = False
```
pinheadmz, line 77: I ran a test with an unexpected message which aborted early, as expected. Funny thing though, because it exited early the test never got to the part where it prints the expected message. So even though its true the expected message was not found in the log, the more relevant error is the unexpected message which occurred first. I'm not sure if there's anything we really need to change about this, unless you have an idea:

```
2024-07-15 13:55:25 | INFO    | scenario | ConnectDag PRNG seed is: 131260415370612
Exception in thread Thread-1 (output_reader):
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.12/3.12.3/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/matthewzipkin/Desktop/work/warnet/test/test_base.py", line 103, in output_reader
    func(line)
  File "/Users/matthewzipkin/Desktop/work/warnet/test/test_base.py", line 67, in _print_and_assert_msgs
    if (self.log_expected_msgs or self.log_unexpected_msgs) and assert_log(
                                                                ^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/test/test_base.py", line 214, in assert_log
    raise AssertionError(f"Unexpected message found in log: {unexpected_msg}")
AssertionError: Unexpected message found in log: Tank
2024-07-15 09:55:29 | DEBUG   | test     | Executing RPC method: scenarios_list_running
2024-07-15 09:55:29 | INFO    | test     | Stopping network
2024-07-15 09:55:29 | DEBUG   | test     | Executing warcli command: network down
2024-07-15 09:55:30 | DEBUG   | test     | Waiting for predicate with timeout 60s and interval 1s
2024-07-15 09:55:30 | DEBUG   | test     | Executing RPC method: network_status
2024-07-15 09:55:30 | INFO    | test     | Waiting for all tanks to reach 'stopped': {'total': 10, 'stopped': 9, 'running': 1}
2024-07-15 09:55:31 | DEBUG   | test     | Executing RPC method: network_status
2024-07-15 09:55:32 | INFO    | test     | Waiting for all tanks to reach 'stopped': {'total': 10, 'stopped': 10}
  File "/Users/matthewzipkin/Desktop/work/warnet/test/dag_connection_test.py", line 52, in <module>
    test.run_test()
  File "/Users/matthewzipkin/Desktop/work/warnet/test/dag_connection_test.py", line 20, in run_test
    self.run_connect_dag_scenario()
  File "/Users/matthewzipkin/Desktop/work/warnet/test/dag_connection_test.py", line 39, in run_connect_dag_scenario
    self.assert_log_msgs()
  File "/Users/matthewzipkin/Desktop/work/warnet/test/test_base.py", line 75, in assert_log_msgs
    self.log_msg_assertions_passed
AssertionError: Log assertion failed. Expected message not found: ['Successfully ran the connect_dag.py scenario using a temporary file']
```

pinheadmz (approved):
ACK 96c7978c1e978abb3e2bd79bc7209a7cdf587350

I think this is good to go. I have 2 suggestions below you can take or leave, or apply to follow-ups...

```
Hash: SHA256

ACK 96c7978c1e978abb3e2bd79bc7209a7cdf587350
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmaVLLwACgkQ5+KYS2KJ
yTr1BhAArvSVNdvm92D/yStxqtkgy05PeQJVxI7X/oeleb/W2BjHq966TckyXKmA
XtnlYi0NirHqAZtldODwjPkXhDI/fYOjt0lr+nfkOtHXM6Oy0T7M7M/2fxNkxtBa
N/mCutSIH07H0s99W/RvdVPL20VINJESzxViR4VIo6LwUu7JuY2v6kNqebpyIJQ5
7XlbFCMNn/tPayqwiFhh09KoTaL7NNtpx+GeKBajGwYVFVNGkaVaahCTR8h2dKCf
SpWzR8PWWe1HhUd5SuslsLs/creHDIJv9NpGWyBq4Ne39meNQiffZe8w/HqIKF3w
RexTbSIfaNwc3HR8t2UhI1kHVXUY9IIPjXpQ71evDxlla4sgUdbSt6ohsFSPZ4oh
W4Xl6QXOz0Xe90lnoIbX2q76OEaEqNjF3d++OYtGpKfh0h2gLPjYZmaxNJZMNHFK
jFEi1gv6faW97c2ErwEiBkW2DpNFSppZeIUkcVAsAP51S2E2Ntq50zj66vkSb3IM
coWPhLuIAZ0lgEpEzyTn5SEQYcNqFuyxgKrY5AxhpMkMa5ul5nCshfPIY1DpLzC3
Jm+ZPbUniHEhQQAJ0GPku58pbKSpzynhf46cdjbLk2NdYT6+rQ113mT5KmDmaQLb
KcRNfbwInpdXFobzlDlySejCZ2QmBFaXi2U2Unnsm26/EtUVNmw=
=1aXP
-----END PGP SIGNATURE-----
```

pinheadmz's public key is [on keybase](https://keybase.io/pinheadmz/pgp_keys.asc?fingerprint=e61773cd6e01040e2f1bd78ce7e2984b6289c93a)
</details>

test/test_base.py
mplsgrant: It looks like the unexpected message AssertionError gets buried in the log, and the less meaningful message not found AssertionError gets highlighted by virtue of being the last message.

To fix this, I would probably use an enum named "Found": Nothing, Expected, Unexpected, and UnexpectedExpected. Then lift the assertion call out of the assert_log helper function and plug it into the TestBase's asset_log_msgs function. Then I would have the helper assert_log functions return the enum.

pinheadmz:
re-ACK 4098f6eb97e223b462d0c348af593211073884db

thanks Grant nice work

PR #385, opened 2024-06-25
`connect_nodes` - account for test networks

mplsgrant:
### Issue
Sometimes `connect_nodes` produces an error:

```shell
2024-06-24T23:39:30+0000 - INFO -     tank_index = int(peer['addr'].split('-')[2])  # NETWORK-tank-INDEX-service
2024-06-24T23:39:30+0000 - INFO -                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-24T23:39:30+0000 - INFO - ValueError: invalid literal for int() with base 10: 'vqyr11r'
```

### Cause
`connect_nodes` does not handle test networks properly. Specifically, it handles this dns format:

`NETWORK-tank-INDEX-service`

But not the "test" format:

`NETWORK-test-TEST-tank-INDEX-service`

### Solution
Modify `connect_nodes` to handle the "test" format, and update the test coverage to account for this.

### DIFF
--- a/src/scenarios/connect_dag.py
+++ b/src/scenarios/connect_dag.py
@@ -58,6 +58,12 @@ class ConnectDag(WarnetTestFramework):
         self.connect_nodes(5, 4)
         self.connect_nodes(5, 6)
         self.connect_nodes(6, 7)
+
+        # Nodes 8 & 9 shall come pre-connected. Attempt to connect them anyway to test the handling
+        # of dns node addresses
+        self.connect_nodes(8, 9)
+        self.connect_nodes(9, 8)
+
         self.sync_all()
 
         zero_peers = self.nodes[0].getpeerinfo()
--- a/src/scenarios/connect_dag.py
+++ b/src/scenarios/connect_dag.py
@@ -116,8 +116,7 @@ class ConnectDag(WarnetTestFramework):
         if connection_type == ConnectionType.DNS:
             assert any(d.get("addr") ==
                        self.warnet.tanks[connectee_index].get_dns_addr() for d in connector), \
-                (f"Could not find {self.options.network_name}-"
-                 f"tank-00000{connectee_index}-service")
+                f"Could not find {self.options.network_name}-tank-00000{connectee_index}-service"
         elif connection_type == ConnectionType.IP:
             assert any(d.get("addr").split(":")[0] ==
                        self.warnet.tanks[connectee_index].get_ip_addr() for d in connector), \
--- a/src/scenarios/connect_dag.py
+++ b/src/scenarios/connect_dag.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+from enum import Enum, auto, unique
 from time import sleep
 
 from warnet.test_framework_bridge import WarnetTestFramework
@@ -10,10 +11,16 @@ def cli_help():
     return "Connect a complete DAG from a set of unconnected nodes"
 
 
+@unique
+class ConnectionType(Enum):
+    IP = auto()
+    DNS = auto()
+
+
 class ConnectDag(WarnetTestFramework):
     def set_test_params(self):
         # This is just a minimum
-        self.num_nodes = 8
+        self.num_nodes = 10
 
     def add_options(self, parser):
         parser.add_argument(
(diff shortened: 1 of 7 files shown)


src/warnet/test_framework_bridge.py
```diff
@@ -337,8 +337,17 @@ def get_peer_ip(peer):
...
                 return ip_addr
-            except ValueError:  # or we encounter a service name
-                tank_index = int(peer['addr'].split('-')[2])  # NETWORK-tank-INDEX-service
+            except ValueError as err:  # or we encounter a service name
+                service_name_parts = peer['addr'].split('-')
+
+                if len(service_name_parts) == 4:
+                    tank_index = int(service_name_parts[2])  # NETWORK-tank-INDEX-service
+                elif len(service_name_parts) == 6:
+                    tank_index = int(service_name_parts[4]) # NETWORK-test-TEST-tank-INDEX-service
```
pinheadmz: `...-test` is, I believe part of the NETWORK name. If all you need here is the index number though, why not just use:

```python
tank_index = int(peer['addr'].split('-')[-2]) # ...-tank-{INDEX}-service
```

mplsgrant:
Updated with simplified tank index extraction logic.

I also took a moment to pull clarify the connect dag scenario by using a function to make assertions. This makes it easier to add new assertions such as for nodes 8 and 9.

src/scenarios/connect_dag.py
```diff
@@ -68,50 +81,50 @@ def run_test(self):
...
+        self.assert_connection(nine_peers, 8, ConnectionType.IP)
 
         self.log.info(f"Successfully ran the {os.path.basename(__file__)} scenario.")
 
+    def assert_connection(self, connector, connectee_index, connection_type: ConnectionType):
+        if connection_type == ConnectionType.DNS:
+            assert any(d.get("addr").split(":")[0] ==
+                       self.warnet.tanks[connectee_index].get_dns_addr() for d in connector), \
+                (f"Could not find {self.options.network_name}-"
+                 f"tank-00000{connectee_index}-service")
```
pinheadmz: nit: could just be one line for readability

pinheadmz (changes requested):
Looking at this again, I think the entire `connect_dag` scenario should just be a test, or can be copied into `dag_connection_test`. It is written very specifically for one graph file and is full of assert statements. Its not like the other scenarios that are written more generically and can apply to a broad range of graphs.

mplsgrant:
@pinheadmz 

> I think the entire `connect_dag` scenario should just be a test, or can be copied into `dag_connection_test`. 

I like this idea, but I don't think we have easy access to `WarnetTestFramework` from `TestBase`.

pinheadmz:
> I like this idea, but I don't think we have easy access to `WarnetTestFramework` from `TestBase`.

oh right i forgot we were going back and forth about using a scenario just for testing, hmmmmmm. It still feels anti-pattern to have the asserts in a scenario instead of a test, but then it also will be silly to split up the `connect_nodes` calls and the asserts into two separate files.

One way this might make more sense, and im not married to this but lets just jam on it a sec:

- move the asserts into the test, leaving the connect_dag scenario very simple, just makes connections
- move the connect_dag file into test/data
- in the test, call `warcli scenarios run_file <path>`

this makes a bit more sense to me because first of all i dont think we cover that particular rpc in any test, and second of all it demonstrates using a "custom" scenario that is written for a specific graph and not intended for general applications

mplsgrant:
I like your thinking about splitting `connect_dag.py` away from scenarios and into some other area that we have designated for testing. I think calling `warcli scenario` on a specific file in the testing area is probably a good way to do that.

In terms of specifics:

> move the connect_dag file into test/data

I'm fine with test/data. However, I think it would be even better to make a folder called "warnet_test_framework_files" or "tank_tests" or similar. That way, this area would be where we place test files that specifically use the test framework/tank features and are not simply files containing plain data.

>move the asserts into the test, leaving the connect_dag scenario very simple, just makes connections

I like keeping the `assert_connection` in connect_dag.py because `assert_connection` is tied really closely to the Tanks. Being tied to the Tanks means to me that we are not testing bitcoin nodes as such, but we are actually testing the interplay between bitcoin nodes and their respective tanks.

*edit*: Just to clarify, I also don't think we can move the asserts into the TestBase test b/c of the need to work with Tanks directly.

> in the test, call `warcli scenarios run_file <path>`

This sounds like it will work and I like it.

mplsgrant:
@pinheadmz I moved the scenario file and incorporated run-file as discussed.

test/framework_tests/connect_dag.py
```diff
@@ -110,7 +110,7 @@ def run_test(self):
         self.assert_connection(eight_peers, 9, ConnectionType.DNS)
         self.assert_connection(nine_peers, 8, ConnectionType.IP)
 
-        self.log.info(f"Successfully ran the {os.path.basename(__file__)} scenario.")
+        self.log.info(f"Successfully ran the scenario file: {os.path.basename(__file__)} ")
```
pinheadmz: This outputs something a little funny:

```
2024-07-01T18:29:51+0000 - INFO - Successfully ran the scenario file: tmpama_so18.py
```

...because we send the scenario file to the server as a base64 string, then the server saves it as a temporary file, which is what is executed.

I'm ok with keeping this as long as it matches your intent.

pinheadmz (approved):
👍  OKIE DOKE!

test/framework_tests/connect_dag.py
mplsgrant: I saw that and think we should keep it. Having the actual file name may be helpful later with debugging.

mplsgrant: Actually, I'll reword it. If it comes across as a bit funny, I think it may be worth it to rework it a bit.

mplsgrant:
@pinheadmz I updated the wording on that final log entry.

LGTM

PR #395, opened 2024-07-03
Use threads to parallelize tx_flood scenario

pinheadmz:
This PR also improves the random attributes of transactions generated, and improves `miner_std` as well.

`warcli scenarios run miner_std --interval=1 --allnodes --mature`
Will generate 101 blocks on each tank (so something is immediately spendable) and then generate 1 block per tank after that, all at 1 second intervals

`warcli scenarios run tx_flood`
Spawns a separate thread in `rpc-0` for every tank, which tries to spend a transaction every second. An address book of all address types from all tanks is maintained in the scenario and random amounts of bitcoin are sent to a random number of addresses from that table. Each tx spends up to half the wallet's available balance although that could probably be reduced so the wallet can send more transactions between blocks. Will play with the values or maybe add some options.

Going to test on 200 node cluster before switching out of draft.

### DIFF
--- a/src/scenarios/miner_std.py
+++ b/src/scenarios/miner_std.py
@@ -47,6 +47,7 @@ class MinerStd(WarnetTestFramework):
         while not self.warnet.network_connected():
             self.log.info("Waiting for complete network connection...")
             sleep(5)
+        self.log.info("Network connected. Starting miners.")
 
         max_miners = 1
         if self.options.allnodes:
@@ -60,9 +61,13 @@ class MinerStd(WarnetTestFramework):
                 if miner.mature:
                     num = 101
                     miner.mature = False
-                self.generatetoaddress(miner.node, num, miner.addr)
-                height = miner.node.getblockcount()
-                self.log.info(f"generated {num} block(s) from node {miner.node.index}. New chain height: {height}")
+                try:
+                    self.generatetoaddress(miner.node, num, miner.addr, sync_fun=self.no_op)
+                    height = miner.node.getblockcount()
+                    self.log.info(f"generated {num} block(s) from node {miner.node.index}. New chain height: {height}")
+                except Exception as e:
+                    self.log.error(f"node {miner.node.index} error: {e}")
+
             sleep(self.options.interval)
 
 
--- a/src/scenarios/miner_std.py
+++ b/src/scenarios/miner_std.py
@@ -7,13 +7,20 @@ from warnet.test_framework_bridge import WarnetTestFramework
 
 
 def cli_help():
-    return "Generate blocks over time. Options: [--allnodes | --interval=<number>]"
+    return "Generate blocks over time. Options: [--allnodes | --interval=<number> | --mature ]"
 
+class Miner:
+    def __init__(self, node, mature):
+        self.node = node
+        self.wallet = ensure_miner(self.node)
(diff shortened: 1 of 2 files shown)


src/scenarios/miner_std.py
```diff
@@ -7,13 +7,20 @@
...
 
 def cli_help():
-    return "Generate blocks over time. Options: [--allnodes | --interval=<number>]"
+    return "Generate blocks over time. Options: [--allnodes | --interval=<number> | --mature ]"
 
+class Miner:
+    def __init__(self, node, mature):
+        self.node = node
+        self.wallet = ensure_miner(self.node)
+        self.addr = self.wallet.getnewaddress()
```
willcl-ark, line 16: Really glad to finally see this abstracted away and only called once. been meaning to do that for ages!

src/scenarios/tx_flood.py
```diff
@@ -7,24 +10,61 @@
...
+    def orders(self, node):
+        wallet = ensure_miner(node)
+        for address_type in ["legacy", "p2sh-segwit", "bech32", "bech32m"]:
+            self.addrs.append(wallet.getnewaddress(address_type=address_type))
+        while True:
+            sleep(self.options.interval)
+            try:
+                bal = wallet.getbalance()
+                if bal < 1:
+                    continue
```
willcl-ark, line 37: Can we mine a block here or top-up somehow?

src/scenarios/tx_flood.py
```diff
@@ -7,24 +10,61 @@
```
willcl-ark: Perhaps you could remove 

```
BLOCKS = 100
TXS = 100
```

While you're in here? (unused?)

src/scenarios/tx_flood.py
```diff
@@ -7,24 +10,61 @@
...
+                bal = wallet.getbalance()
+                if bal < 1:
+                    continue
+                amounts = {}
+                num_out = randrange(1, len(self.nodes) // 2)
+                for _ in range(num_out):
+                    sats = int(float((bal / 20) / num_out) * 1e8)
+                    amounts[choice(self.addrs)] = randrange(sats // 4, sats) / 1e8
+                wallet.sendmany(dummy="", amounts=amounts)
+                self.log.error(f"node {node.index} sent tx with {num_out} outputs")
```
willcl-ark:
```suggestion
                self.log.info(f"node {node.index} sent tx with {num_out} outputs")
```

willcl-ark (approved):
LGTM.

Just left a few Qs, but would also be happy to merge in this state

src/scenarios/tx_flood.py
pinheadmz: I'm leaving the block mining to other scenarios for composability. Given the randomness of this scenario, i thought it best just to skip low balance wallets. Keep in mind that even without new blocks, other nodes will be sending you bitcoin as well.

src/scenarios/tx_flood.py
pinheadmz: nice thanks will do

src/scenarios/tx_flood.py
pinheadmz: whoopsie doodle

willcl-ark:
ACK

Issue #390, opened 2024-06-28
Create an automated base-level network activity scenario


willcl-ark:
This may first require something like https://github.com/bitcoin-dev-project/warnet/issues/272 so that activity is outsourced locally to the nodes, rather then relying on roundtrips from `rpc-0`.

- may want to ensure full blocks and/or mempools of nodes

willcl-ark:
Linked to https://github.com/bitcoin-dev-project/warnet/issues/7

m3dwards:
Related discussion: https://github.com/bitcoin-dev-project/warnet/discussions/392

josibake:
Relevant for Tabconf, if only a conceptual discussion/documentation on how we expect network activity to be generated for the war-games.

Issue #272, opened 2024-02-09
Make Commanders run any executable?


willcl-ark:
I think we should run a second process on the bitcoin nodes, which will simply watch a directory for executables, and run them when new ones are detected.

This watches `/tmp/watchdog` and runs whatever it finds there (#dangerzone).

![image](https://github.com/bitcoin-dev-project/warnet/assets/6606587/4f503b0f-5437-449c-8ae2-f0f9677de3f8)

This is of course fun, but also we can use it to generate random network activity without requiring the RPC server to co-ordinate _everything_ using synchronous JSONRPC calls over http (zzzzzz)...

Someone already made a nice python lib to do this, so the code ends up pretty [simple](https://github.com/willcl-ark/warnet/blob/9cbc88e5d8cfedbe8c417c056c4e2f815891c03c/scripts/file-executor.py).

I know it _sounds_ like a terrible idea, but I think it might actually be quite useful! The only issue is that we _may_ want to add a gpg sig file or something if we had this during an attack day, as we wouldn't last long with folks dropping in auto-executed files...

Thoughts?

josibake:
So the idea here is you could scp' a script to one or more tanks and they would automatically start doing whatever the script tells the container to do until the file is later removed?

Seems like it *could* be useful, but somewhat goes against the idea of reproducibility and doesn't seem to fit with using the python test framework for driving network wide scenarios. Perhaps we table it for now, unless you have a compelling usecase?

willcl-ark:
ping @m3dwards @josibake @mplsgrant @pinheadmz 

Found it

willcl-ark:
Going to close this for now, as with the `Commander` layout I no longer feel the RPC server is a bottleneck (though I still do like this concept of running "anything", and this it's perhaps how a Commander should operate)

josibake:
I think in general we will move the commander to be able to dockerize any script the user wants. If we also want a script or process running in the same pod as a bitcoind node, I feel pretty strongly it should be injected via the helm chart. we talked about this notion of "profiles" before and that feels like the better place to start deciding what a bitcoind node looks like, including what little scripts we want running in the same container.

willcl-ark:
I'll leave this open and rename the issue then

m3dwards:
I prefer the idea of running scripts through commanders as they can be viewed and killed like any other scenario and nothing stopping many being launched all with a specific job.

willcl-ark:
[...]

PR #396, opened 2024-07-03
WIP: add remote executor and tank roles

willcl-ark:
Some real commits and some example commits to-be-removed.

Addresses the idea in https://github.com/bitcoin-dev-project/warnet/issues/272

This installs a "file watcher" inside the tank docker image, which will watch `/tmp/exe/` for executable files and run them as soon as they are added.

Add the ability to assign a "role" to a tank on the graph. This will (stupidly) attempt to copy a script from `scripts/{role}.sh` into the tank `/tmp/exe/` during warnet creation. This happens after the file watcher is started, (and likely bitcoind is initialised) so it will immediately execute.

A simple miner script, which mines blocks every 0 - 20 seconds, is applied as the role to tank `0`.

Currently this uses a custom [image](https://hub.docker.com/layers/bitcoindevproject/bitcoin/27.1-exe/images/sha256-7bba6af27eff756b5d96310d3e60ba3ed2d5a026f7a53569a09d2c044ec974fa?context=repo) (x86_64 linux only, I think) for the tank, with the new executor program installed, but this could obviously be rebuilt into the main images if we move forward with it.

This should let us (and hackers?) make the nodes do pretty wild things, unconstrained by `rpc-0`'s load/RPC timings.

Could also be used as a start to achieving https://github.com/bitcoin-dev-project/warnet/issues/390

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,11 +7,11 @@ on:
       - main
 
 jobs:
-  ruff:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: chartboost/ruff-action@v1
+  # ruff:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v4
+  #     - uses: chartboost/ruff-action@v1
   ruff-format:
     runs-on: ubuntu-latest
     steps:
@@ -20,7 +20,8 @@ jobs:
         with:
           args: 'format --check'
   build-image:
-    needs: [ruff, ruff-format]
+    # needs: [ruff, ruff-format]
+    needs: [ruff-format]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/resources/graphs/default.graphml
+++ b/resources/graphs/default.graphml
@@ -15,6 +15,7 @@
   <key id="channel_open"    attr.name="channel_open"     attr.type="string"   for="edge" />
   <key id="source_policy"   attr.name="source_policy"    attr.type="string"   for="edge" />
   <key id="target_policy"   attr.name="target_policy"    attr.type="string"   for="edge" />
+  <key id="role"            attr.name="role"             attr.type="string"   for="node" />
   <graph edgedefault="directed">
     <node id="0">
         <data key="version">27.1-exe</data>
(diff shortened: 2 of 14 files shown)


src/backend/kubernetes_backend.py
```diff
@@ -860,3 +860,103 @@ def write_service_config(self, source_path: str, service_name: str, destination_
             _preload_content=False,
         )
         self.log.info(f"Finished unpacking config data for {service_name} to {destination_path}")
+
+    def copy_file_to_pod(self, pod_name, src_path, dest_path):
```
willcl-ark: I hate this function.

Also, I noticed testing this out that a tonne of things are way easier via `kubectl exec -it warnet-tank-<index>`. This is what I was missing since we dropped compose; hop into the container with a single command.

It made me think though, I think a tonne of our functionality could be re-written to use the k8s client object directly, and end up a lot neater...

pinheadmz: You could also look in to using the sidecar image I made for simln: https://hub.docker.com/r/pinheadmz/sidecar. One benefit of this is that it is paired with a shared volume so if the container restarts, the data is still there.

pinheadmz (commented):
I'm not super-duper in love with this approach, especially if using threads in scenarios continues to work out OK. I also wonder if, when a user needs something like this, can they just create a one-off image with a new entrypoint file that runs some extra stuff?

src/backend/kubernetes_backend.py
willcl-ark: Interesting... will investigate

willcl-ark:
> I'm not super-duper in love with this approach, especially if using threads in scenarios continues to work out OK. I also wonder if, when a user needs something like this, can they just create a one-off image with a new entrypoint file that runs some extra stuff?

That seems to me to be pretty much directly equivalent, although more difficult for end-users? With this approach all they need to do is add a graph property and drop an executable in a dir. Re-creating an entire image with a new entrypoint.sh feels more difficult than this, to me.

I do agree however that, if we can essentially count on "unlimited threads", then a threaded, centralised co-ordinator may well be superior for bitcoind/lnd/cln orchestration.

That said, this approach would let you insert arbitrary bash scripts which could include doing systems things to nodes like:

- take pod networking on/offline
- shutting down bitcoind and restarting it later
- corrupt/remove files in the datadir
- etc.

...very easily. None of which are currently possible today.

Issue #389, opened 2024-06-28
One-click install/spin up based on commit hash


willcl-ark:
Users should be able to quick-start by simply providing a commit hash of a branch they want to test.

May require automated base-level activity scenario [link] ( Issue#? )

willcl-ark:
Linked to https://github.com/bitcoin-dev-project/warnet/issues/389

willcl-ark:
Started towards this in #397

pinheadmz:
> Linked to #389

this *IS* 389 🤔

pinheadmz:
closing as dupe of https://github.com/bitcoin-dev-project/warnet/issues/708

Issue #181, opened 2023-10-31
Support building custom PRs with k8s


josibake:
In https://github.com/bitcoin-dev-project/warnet/pull/178, custom PRs are not supported, unless they have been built and pushed to bitcoindevproject/bitcoin-core. Your everyday user will not have access to this docker hub account, so we need a way for them to build images and make them available to the cluster.

A few ideas:

* Build locally and push to a "local registry" running in the cluster
* Build inside the cluster (I was able to do this with Kaniko) and push to  a local registry
* Build the images as tar files and put them in a volume?
* Something else?

This could be handled all by the RPC server, and then nothing changes for the tanks: they still expect an image to be available in some sort of registry

willcl-ark:
We can now build images and push to any repo using `warcli image build` [command](https://github.com/bitcoin-dev-project/warnet/blob/main/docs/warcli.md#image)

I don't think it's unreasonable to expect a user to have access to a (free) dockerhub account? In addition, I think the above suggestions, are difficult to implement.

@josibake can you close this if you agree with me?

pinheadmz:
@willcl-ark closed by #397 ?

willcl-ark:
I would like to reopen this (but not for TABConf).

I think we should guide users through building a custom image, and uploading to hub.docker.com. This can be done quite nicely now using inquirer and checking for env vars `DOCKERHUB_USERNAME` and `DOCKERHUB_PASS` (or similar).

IMO a new user should be able to pick this up and, either as part of quickstart or `image build` have this application handle/hand-hold them through the entire process.

PR #397, opened 2024-07-05
image: build by commit hash, not branch

willcl-ark:
Build an image from any `repo + <commit_hash>`.

I think this is more flexible and more useful.

It's also a step towards https://github.com/bitcoin-dev-project/warnet/issues/389 and #181

### DIFF
--- a/src/templates/Dockerfile
+++ b/src/templates/Dockerfile
@@ -1,7 +1,7 @@
 # Setup deps stage
-FROM alpine as deps
+FROM alpine AS deps
 ARG REPO
-ARG BRANCH
+ARG COMMIT_SHA
 ARG BUILD_ARGS
 
 RUN --mount=type=cache,target=/var/cache/apk \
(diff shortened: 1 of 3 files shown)


src/templates/Dockerfile
```diff
@@ -27,14 +27,16 @@ COPY src/templates/addrman.patch /tmp/
...
 ENV BITCOIN_PREFIX=/opt/bitcoin
 WORKDIR /build
 
 RUN set -ex \
     && cd /build \
-    && git clone --depth 1 --branch "${BRANCH}" "https://github.com/${REPO}" \
+    && git clone --depth 1 "https://github.com/${REPO}" \
     && cd bitcoin \
+    && git fetch --depth 1 origin "$COMMIT_SHA" \
+    && git checkout "$COMMIT_SHA" \
```
pinheadmz, line 39: TIL you can do this: fetch a commit from another branch even though you cloned depth 1

pinheadmz (approved):
ACK

reviewed and tested. image is building locally right now but it checked out the right HEAD so I think we're good here

src/templates/Dockerfile
willcl-ark: `fetch` is _from the remote_, so really the `clone` is just needed to setup `git` with the remote, then you can `fetch` any commit_sha (at depth 1) after that.

It's prob even possible to do in a one-liner, but I don't got time for that

PR #400, opened 2024-07-08
Refactor tests

willcl-ark:
This has been bothering me for a long while... Fix it by refactoring all the tests (but not `WarnetTestFramework`, yet) to give them more homogenous structure.

There are a few "tidy-ups" in here too, sorry, but it's but mainly just breaking things down into a function-per-test, much like the bitcoin core functional test suite.

Makes it easier to add new tests, and re-use parts of other tests.

Also improve test logging using `logging` module where possible.

### DIFF
--- a/src/logging_config/config.json
+++ b/src/logging_config/config.json
@@ -3,12 +3,12 @@
   "disable_existing_loggers": false,
   "formatters": {
     "simple": {
-      "format": " %(asctime)s - %(levelname)s - %(message)s",
-      "datefmt": "%Y-%m-%dT%H:%M:%S%z"
+      "format": "%(asctime)s │ %(levelname)-7s │ %(name)-17s │ %(message)s",
+      "datefmt": "%Y-%m-%d %H:%M:%S"
     },
     "detailed": {
-      "format": " %(asctime)s - %(levelname)s - [%(module)s|L%(lineno)d] - %(message)s",
-      "datefmt": "%Y-%m-%dT%H:%M:%S%z"
+      "format": "%(asctime)s │ %(levelname)-7s │ [%(module)21s:%(lineno)4d] │ %(message)s",
+      "datefmt": "%Y-%m-%d %H:%M:%S"
     }
   },
   "filters": {
--- a/src/warnet/test_framework_bridge.py
+++ b/src/warnet/test_framework_bridge.py
@@ -48,16 +48,14 @@ class WarnetTestFramework(BitcoinTestFramework):
 
     def setup(self):
         signal.signal(signal.SIGTERM, self.handle_sigterm)
+
+        # Must setuup warnet first to avoid double formatting
+        self.warnet = Warnet.from_network(self.options.network)
         # hacked from _start_logging()
         # Scenarios will log plain messages to stdout only, which will can redirected by warnet
-        self.log = logging.getLogger()
+        self.log = logging.getLogger("WarnetTestFramework")
         self.log.setLevel(logging.INFO)  # set this to DEBUG to see ALL RPC CALLS
-        ch = logging.StreamHandler(sys.stdout)
-        formatter = logging.Formatter(fmt="%(message)s")
-        ch.setFormatter(formatter)
-        self.log.addHandler(ch)
 
-        self.warnet = Warnet.from_network(self.options.network)
         for i, tank in enumerate(self.warnet.tanks):
             ip = tank.ipv4
             self.log.info(f"Adding TestNode {i} from tank {tank.index} with IP {ip}")
--- a/test/ln_test.py
+++ b/test/ln_test.py
@@ -7,99 +7,118 @@ from pathlib import Path
 from test_base import TestBase
 from warnet.services import ServiceType
 
-graph_file_path = Path(os.path.dirname(__file__)) / "data" / "ln.graphml"
 
-base = TestBase()
-base.start_server()
+class LNTest(TestBase):
+    def __init__(self):
+        super().__init__()
+        self.graph_file_path = Path(os.path.dirname(__file__)) / "data" / "ln.graphml"
 
+    def run_test(self):
+        self.start_server()
+        try:
+            self.setup_network()
+            self.run_ln_init_scenario()
+            self.test_channel_policies()
+            self.test_ln_payment_0_to_2()
+            self.test_ln_payment_2_to_0()
+            self.test_simln()
+        finally:
+            self.stop_server()
 
-def get_cb_forwards(index):
-    cmd = "wget -q -O - 127.0.0.1:9235/api/forwarding_history"
-    res = base.wait_for_rpc(
-        "exec_run", [index, ServiceType.CIRCUITBREAKER.value, cmd, base.network_name]
-    )
-    return json.loads(res)
+    def setup_network(self):
+        self.log.info("Setting up network")
+        self.log.info(self.warcli(f"network start {self.graph_file_path}"))
+        self.wait_for_all_tanks_status(target="running")
+        self.wait_for_all_edges()
 
+    def get_cb_forwards(self, index):
+        cmd = "wget -q -O - 127.0.0.1:9235/api/forwarding_history"
+        res = self.wait_for_rpc(
+            "exec_run", [index, ServiceType.CIRCUITBREAKER.value, cmd, self.network_name]
+        )
+        return json.loads(res)
 
-print(base.warcli(f"network start {graph_file_path}"))
-base.wait_for_all_tanks_status(target="running")
-base.wait_for_all_edges()
+    def run_ln_init_scenario(self):
+        self.log.info("Running LN Init scenario")
+        self.warcli("rpc 0 getblockcount")
+        self.warcli("scenarios run ln_init")
+        self.wait_for_all_scenarios()
 
-print("\nRunning LN Init scenario")
-base.warcli("rpc 0 getblockcount")
-base.warcli("scenarios run ln_init")
-base.wait_for_all_scenarios()
+    def test_channel_policies(self):
+        self.log.info("Ensuring node-level channel policy settings")
+        node2pub, node2host = json.loads(self.warcli("lncli 2 getinfo"))["uris"][0].split("@")
+        chan_id = json.loads(self.warcli("lncli 2 listchannels"))["channels"][0]["chan_id"]
+        chan = json.loads(self.warcli(f"lncli 2 getchaninfo {chan_id}"))
 
-node2pub, node2host = json.loads(base.warcli("lncli 2 getinfo"))["uris"][0].split("@")
+        # node_1 or node_2 is tank 2 with its non-default --bitcoin.timelockdelta=33
+        if chan["node1_policy"]["time_lock_delta"] != 33:
+            assert (
+                chan["node2_policy"]["time_lock_delta"] == 33
+            ), "Expected time_lock_delta to be 33"
 
-print("\nEnsuring node-level channel policy settings")
-chan_id = json.loads(base.warcli("lncli 2 listchannels"))["channels"][0]["chan_id"]
-chan = json.loads(base.warcli(f"lncli 2 getchaninfo {chan_id}"))
-# node_1 or node_2 is tank 2 with its non-default --bitcoin.timelockdelta=33
-if chan["node1_policy"]["time_lock_delta"] != 33:
-    assert chan["node2_policy"]["time_lock_delta"] == 33
+        self.log.info("Ensuring no circuit breaker forwards yet")
+        assert len(self.get_cb_forwards(1)["forwards"]) == 0, "Expected no circuit breaker forwards"
 
-print("\nEnsuring no circuit breaker forwards yet")
-assert len(get_cb_forwards(1)["forwards"]) == 0
+    def test_ln_payment_0_to_2(self):
+        self.log.info("Test LN payment from 0 -> 2")
+        inv = json.loads(self.warcli("lncli 2 addinvoice --amt=2000"))["payment_request"]
+        self.log.info(f"Got invoice from node 2: {inv}")
+        self.log.info("Paying invoice from node 0...")
+        self.log.info(self.warcli(f"lncli 0 payinvoice -f {inv}"))
 
-print("\nTest LN payment from 0 -> 2")
-inv = json.loads(base.warcli("lncli 2 addinvoice --amt=2000"))["payment_request"]
+        self.wait_for_predicate(self.check_invoice_settled)
 
-print(f"\nGot invoice from node 2: {inv}")
-print("\nPaying invoice from node 0...")
-print(base.warcli(f"lncli 0 payinvoice -f {inv}"))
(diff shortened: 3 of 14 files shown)


willcl-ark:
fpush to re-add some erroneously-dropped comments,

test/test_base.py
```diff
@@ -94,67 +111,53 @@ def start_server(self):
...
+        self.log.debug(f"Waiting for predicate with timeout {timeout}s and interval {interval}s")
+        while timeout > 0:
             if predicate():
-                break
+                return
             sleep(interval)
             timeout -= interval
-            if timeout < 0:
-                raise Exception("Timed out waiting for predicate Truth")
+        raise Exception("Timed out waiting for predicate Truth")
```
pinheadmz: 77f72dfcd5696c6c7ff5e4a7b80c899ca761e08e

while youre here you can also log `predicate.__name__` which i think we've all wanted from this error

test/rpc_test.py
```diff
@@ -6,48 +6,70 @@
...
+            self.test_address_manager()
+        finally:
+            self.stop_server()
 
-# Fund wallet
-base.warcli("rpc 1 createwallet miner")
-base.warcli("rpc 1 -generate 101")
+    def setup_network(self):
+        self.log.info("Setting up network")
+        print(self.warcli(f"network start {self.graph_file_path}"))
```
pinheadmz: 4d2fdd0fbea8740951e48ca26abe0b2876b2afcc

switch to log here?

test/test_base.py
pinheadmz: update: chatgpt says this should work if the function is a lambda, to print its source code:

```python
import inspect

inspect.getsource(predicate).strip()
```

neat!

test/scenarios_test.py
```diff
@@ -5,48 +5,71 @@
...
+    def run_and_check_scenario(self, scenario_name):
+        self.log.info(f"Running scenario: {scenario_name}")
+        self.warcli(f"scenarios run {scenario_name} --allnodes --interval=1")
+        self.wait_for_predicate(lambda: self.check_blocks(30))
+        self.stop_scenario()
+
+    def run_and_check_scenario_from_file(self, scenario_file):
+        self.log.info(f"Running scenario from file: {scenario_file}")
+        self.warcli(f"scenarios run-file {scenario_file} --allnodes --interval=1")
+        self.wait_for_predicate(lambda: self.check_blocks(32))
```
pinheadmz: e36df8ea8e1e1d16ec60373c1bf2be1601fbc1b5

This isnt great because there might still be a race from the last test not stopping in time. Maybe you can start the updated tx_scenario here instead, and wait for predicate can check for `getchaintxstats["tx_count"] > getblockcount` (or ensure there are 50 more txs than blocks i.e.  coinbase txs). Although I guess if you do that youll need to keep `miner_std` running as well so the txs confirm and get counted

test/scenarios_test.py
```diff
@@ -5,48 +5,71 @@
...
+    def setup_network(self):
+        self.log.info("Setting up network")
+        self.log.info(self.warcli(f"network start {self.graph_file_path}"))
+        self.wait_for_all_tanks_status(target="running")
+
+    def test_scenarios(self):
+        self.check_available_scenarios()
+        self.run_and_check_scenario("miner_std")
+        self.run_and_check_scenario_from_file("src/scenarios/miner_std.py")
+
```
pinheadmz, line 31: e36df8ea8e1e1d16ec60373c1bf2be1601fbc1b5

heh oh kay, OR you could just put these lines in `run_test` ?

test/v25_net_test.py
```diff
@@ -6,55 +6,66 @@
...
-onion_addr = None
-
-
-def wait_for_reachability():
-    try:
-        global onion_addr
-        info = json.loads(base.warcli("rpc 0 getnetworkinfo"))
-        for net in info["networks"]:
-            if net["name"] == "ipv4" and not net["reachable"]:
+class V25NetTest(TestBase):
```
pinheadmz: 0f439f1a5812cd033516fccb56d14b1ea7e4a103

lets rename this test class and file `OnionTest`, we haven't used bitcoin core v25 in a year :-/ see #306 

I appreciate you cleaning this one up even though we cant use it yet, thanks.

test/ln_test.py
```diff
@@ -7,99 +7,118 @@
...
+            self.test_channel_policies()
+            self.test_ln_payment_0_to_2()
+            self.test_ln_payment_2_to_0()
+            self.test_simln()
+        finally:
+            self.stop_server()
+
+    def setup_network(self):
+        self.log.info("Setting up network")
+        print(self.warcli(f"network start {self.graph_file_path}"))
```
pinheadmz: cdc1de058b6efe42b55b30ae63e25b6ad1ffa434

damn i haven't been checking them all but i think this should be a log message not print.

pinheadmz (commented):
fuck yeah bravo, muchas gracias this is great.

some suggestions below

test/test_base.py
```diff
@@ -13,71 +13,87 @@
...
+    def setup_logging(self):
+        import logging
+
+        self.log = logging.getLogger("TestFramework")
+        self.log.setLevel(logging.DEBUG)
+        fh = logging.FileHandler(self.logfilepath, encoding="utf-8")
+        fh.setLevel(logging.DEBUG)
+        ch = logging.StreamHandler()
+        ch.setLevel(logging.DEBUG)
+        formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
```
pinheadmz: something is a bit funky here, im seeing double timestamps in some but not all log messages:

```

2024-07-08 15:18:43,320 - TestFramework - DEBUG - Node 2 has 1 settled invoices
2024-07-08 15:18:48,326 - TestFramework - DEBUG - Executing warcli command: lncli 2 listinvoices
2024-07-08 15:18:48,711 - TestFramework - DEBUG - 2024-07-08T19:18:48+0000 - DEBUG - {'jsonrpc': '2.0', 'method': 'tank_lncli', 'params': {'network': 'warnet-test-wspv4wt2', 'node': 2, 'command': ['listinvoices']}, 'id': 1}
2024-07-08 15:18:48,714 - TestFramework - DEBUG - 2024-07-08T19:18:48+0000 - DEBUG - Running lncli cmd=lncli --network=regtest listinvoices on tank.index=2
2024-07-08 15:18:48,714 - TestFramework - DEBUG - 2024-07-08T19:18:48+0000 - DEBUG - Running exec_cmd=['/bin/sh', '-c', 'lncli --network=regtest listinvoices'] on tank_index=2
```

willcl-ark: Yeah, i need to update this to use our logging.config too: https://github.com/bitcoin-dev-project/warnet/pull/384#issuecomment-2214747548

test/test_base.py
```diff
@@ -94,67 +111,53 @@ def start_server(self):
...
-                    if circuitbreaker_status not in stats:
-                        stats[circuitbreaker_status] = 0
-                    stats[circuitbreaker_status] += 1
-            print(f"Waiting for all tanks to reach '{target}': {stats}")
-            # All tanks are running, proceed
+                for service in ["bitcoin", "lightning", "circuitbreaker"]:
+                    status = tank.get(f"{service}_status")
+                    if status:
+                        stats[status] = stats.get(status, 0) + 1
+            self.log.info(f"Waiting for all tanks to reach '{target}': {stats}")
```
pinheadmz: something else funky here, `total` shouldn't be the smallest number!

examples from local tests

```
 Waiting for all tanks to reach 'stopped': {'total': 4, 'stopped': 3, 'pending': 4, 'running': 2}

```





```
 Waiting for all tanks to reach 'stopped': {'total': 4, 'stopped': 9}
```

This test still stopped eventually because of
```
 ERROR - Error bringing network down: Timed out waiting for predicate Truth

```

test/ln_test.py
willcl-ark: fixed all, thanks

test/rpc_test.py
willcl-ark: done

test/scenarios_test.py
willcl-ark: Might add more files in the future though? (probably not, lol!)

Gunna leave though I think

test/scenarios_test.py
willcl-ark: Nice catch, tried to fixup with https://github.com/bitcoin-dev-project/warnet/pull/400/commits/61e33318926a1a64dde61cdfe1a862f7eea8a452#diff-3c6951d855400160cce6184d9af9c0e5a663af6ae1f55a232fd53c53d6971188R51

test/test_base.py
willcl-ark: sweet, taken

test/test_base.py
willcl-ark: damn wth. I was so happy with that too. Is that happening every time? What test did you see that on?

It doesn't seem to for me:

image


Will investigate more tests in a bit.

test/v25_net_test.py
willcl-ark: nice, done

willcl-ark (commented):
Thanks @pinheadmz 

Taken almost everything in [5e361f9f60580c9e12046670e9843ed44d556ce2](https://github.com/bitcoin-dev-project/warnet/compare/3b224abc9131a25a3ae46c18f470fc448e19de4b..5e361f9f60580c9e12046670e9843ed44d556ce2)

test/test_base.py
pinheadmz: try the ln_test because it has 4 tanks but a bunch of other pods for ln and cb

test/test_base.py
```diff
@@ -108,7 +106,7 @@ def start_server(self):
         )
 
         self.server_thread = threading.Thread(
-            target=self.output_reader, args=(self.server.stdout, self.log.debug)
+            target=self.output_reader, args=(self.server.stdout, print)
```
mplsgrant, line 109: Why bring print back from earlier in the PR? https://github.com/bitcoin-dev-project/warnet/pull/400/commits/ca382c7268b456a900ca3667bc64359f649c42ef#diff-1a293b58d544ac5f9d384afd0652dc7e9e17ae7564df252dd5a8fc39543ee0e6L99

In #384, the route I took was to create an instance method that takes one argument and does a call to logger: https://github.com/mplsgrant/warnet/blob/3e2d110d3bfbf73518a0af492832c57ae04372bd/test/test_base.py#L48

```python
    def passthrough_logger(self, message):
        self.logger.info(f"{self.pod_name} - {message}")
```

But maybe we can just leave this as is and change it in #384.

test/test_base.py
willcl-ark: fixed in e1373a91bd7937de1d0dd5f7d746301e76b2c5d8 and 78722778eef821311af8b13a0b6685bbd0454e7c and 34e3331599db82565ad803387e482ff6e5c652ca

test/test_base.py
willcl-ark: Fixed in https://github.com/bitcoin-dev-project/warnet/commit/e1373a91bd7937de1d0dd5f7d746301e76b2c5d8

test/test_base.py
willcl-ark: > Why bring print back from earlier in the PR? [ca382c7#diff-1a293b58d544ac5f9d384afd0652dc7e9e17ae7564df252dd5a8fc39543ee0e6L99](https://github.com/bitcoin-dev-project/warnet/commit/ca382c7268b456a900ca3667bc64359f649c42ef#diff-1a293b58d544ac5f9d384afd0652dc7e9e17ae7564df252dd5a8fc39543ee0e6L99)

The reason is this is being logged from outside already (well, here in test_base), and we are using the `output_reader` to capture all output from stdout and log it. So if we `log`, it prepends all the log info (time data etc.). We then capture this entire line via stdout and prepend a second layer of log info in `output_reader`, so we end up with double log details.
 

> But maybe we can just leave this as is and change it in #384.

Hopefully this :P

willcl-ark: The passtrhough logger does feel quite nice though... it would probably be nicer to get an actual (sub) logger sharing the same logfile/stdout stream though

test/scenarios_test.py
```diff
@@ -5,48 +5,72 @@
...
+        self.warcli(f"scenarios run {scenario_name} --allnodes --interval=1")
+        self.wait_for_predicate(lambda: self.check_blocks(30))
+        self.stop_scenario()
+
+    def run_and_check_scenario_from_file(self, scenario_file):
+        self.log.info(f"Running scenario from file: {scenario_file}")
+        self.warcli(f"scenarios run-file {scenario_file} --allnodes --interval=1")
+        start = int(self.warcli("rpc 0 getblockcount"))
+        self.wait_for_predicate(lambda: self.check_blocks(2, start=start))
+
```
m3dwards, line 51: Should there also be a `stop_scenario()` here?

m3dwards:
A problem I had with the `ln_test` was that as part of it's setup it called a scenario (`ln_init`) but there's no way to check if the scenario failed or not. Perhaps the list of running scenarios should keep a list of failed ones?

test/scenarios_test.py
willcl-ark: Yes, thanks. I think it does get cleaned up ok, but better to have it in :)

willcl-ark:
> A problem I had with the `ln_test` was that as part of it's setup it called a scenario (`ln_init`) but there's no way to check if the scenario failed or not. Perhaps the list of running scenarios should keep a list of failed ones?

They are marked as `active` with a bool, no? Can we check that?

m3dwards:
> They are marked as `active` with a bool, no? Can we check that?

Oh maybe, I didn't understand what not active meant.

m3dwards:
LGTM. Ship it!

test/scenarios_test.py
pinheadmz: mmmm ok Id still prefer loading a different scenario file entirely, maybe even a fragment we can keep in `test/data` or `test/framework_tests`

src/warnet/test_framework_bridge.py
```diff
@@ -48,16 +48,14 @@ def handle_sigterm(self, signum, frame):
 
     def setup(self):
         signal.signal(signal.SIGTERM, self.handle_sigterm)
+
+        # Must setuup warnet first to avoid double formatting
```
pinheadmz, line 52: setuuuuuuup!

src/logging_config/config.json
```diff
@@ -3,12 +3,12 @@
   "disable_existing_loggers": false,
   "formatters": {
     "simple": {
-      "format": " %(asctime)s - %(levelname)s - %(message)s",
-      "datefmt": "%Y-%m-%dT%H:%M:%S%z"
+      "format": "%(asctime)s │ %(levelname)-7s │ %(name)-17s │ %(message)s",
```
pinheadmz, line 6: would prefer to leave out `|` to save a bit more space, and limit `(name)` as well, i bet we can cap it at 8 chars and come up with short little nicknames for each module

PR #404, opened 2024-07-08
Enforce ruff format

willcl-ark:
- Enables `ruff format --check .` which checks formatting
	- this should only operate on changed files
- Format all files
- Include scenarios in formatting & reformat all scenarios
- Fixup lint issues in tx_flood changes (not binding to values in lambda)
- Add a justfile command to check formatting and linting `just lint`

After these changes, PRs will fail CI checks if the entire file is not formatted and lint-free, which can be done with:

```
ruff format .
ruff check .
```

If we do this, I could add a precommit hook perhaps which does this? Although IMO it's better to configure your editor to do it automagically if possible, otherwise you get into a mess between what is staged, suddenly getting reformatted, and the reformatting going into a subsequent commit (if you only stage partial changes).

Added a `just lint` helper command for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,8 +12,18 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
+  ruff-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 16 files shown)


pinheadmz:
>  precommit hook 

please no, i hate it when code i push is not exactly how it looked in my IDE 😭

PR #408, opened 2024-07-10
[WIP] Refactor graph test cln

willcl-ark:
adds refactor and debugging


willcl-ark:
can see what tests make of it; graph will fail in the ususal way

PR #410, opened 2024-07-16
ln node might only know about 1 side of channel

m3dwards:
Because gossip for both sides of a channel doesn't arrive at the same time, nodes might not always be able to build a full picture of a channel. For the sake of comparison, if two nodes both do not know about the policy of one side of a channel we consider them to be in sync with each other.

### DIFF
--- a/src/warnet/cln.py
+++ b/src/warnet/cln.py
@@ -5,7 +5,7 @@ from backend.kubernetes_backend import KubernetesBackend
 from warnet.services import ServiceType
 from warnet.utils import exponential_backoff, generate_ipv4_addr, handle_json
 
-from .lnchannel import LNChannel
+from .lnchannel import LNChannel, LNPolicy
 from .lnnode import LNNode
 from .status import RunningStatus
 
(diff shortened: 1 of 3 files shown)


src/warnet/lnchannel.py
```diff
@@ -100,22 +82,54 @@ def compare_attributes(attr1, attr2, min_value=0, attr_name=""):
...
+            return ch2.node1_policy is None
+
+        if self.node2_policy is None:
+            return ch2.node2_policy is None
+
+        if ch2.node1_policy is None:
+            return self.node1_policy is None
+
+        if ch2.node2_policy is None:
+            return self.node2_policy is None
```
pinheadmz: I think its too early to return here?
example, if `ch 1 policy 1 == None and ch 2 policy 1 == None` we'd return True before comparing policy 2?

src/warnet/lnchannel.py
```diff
@@ -1,52 +1,51 @@
...
 
 
+class LNPolicy:
+    def __init__(
+        self,
+        min_htlc: int = 0,
+        max_htlc: int = 0,
+        base_fee_msat: int = 0,
+        fee_rate_milli_msat: int = 0,
+        time_lock_delta: int = 0,
```
pinheadmz: Not sure if these should have default values at all, but probably doesnt matter because we replace all or nothing

src/warnet/cln.py
```diff
@@ -96,29 +96,46 @@ def get_graph_channels(self) -> list[LNChannel]:
...
+                fee_rate_milli_msat=node2["fee_per_millionth"],
+            )
+            if node2 is None
+            else None
+        )
+
         return LNChannel(
             node1_pub=node1["source"],
-            node2_pub=node2["source"],
+            node2_pub=node2["source"] if node2 else node1["destination"],
```
pinheadmz: could remove the `if` here since there will always be a `node1["destination"]` and we shouldnt need to worry that node2 might have a different source

src/warnet/lnchannel.py
m3dwards: Excellent catch. Reworked logic.

src/warnet/lnchannel.py
m3dwards: Just left default in for time_lock_delta as CLN doesn't give us that.

src/warnet/cln.py
m3dwards: Obvious now you said it.

m3dwards:
Ready for review @pinheadmz

pinheadmz (approved):
light-is-green-trap-is-clean

PR #419, opened 2024-07-30
`warcli cluster` refactors

willcl-ark:
* Move scripts from python files into dedicated scripts (callable from both justfile and cli)
* Refactor `warcli cluster` commands:
  * write a small config saving cluster_type and any minikube settings with `warcli cluster configure`
  * `warcli cluster deploy` now runs the same script as the justfile, deploying on whatever the connected cluster is
* update docs

### DIFF
--- a/resources/scripts/deploy.sh
+++ b/resources/scripts/deploy.sh
@@ -22,9 +22,12 @@ kubectl apply -f "$WAR_MANIFESTS/warnet-rpc-service.yaml"
 
 # Deploy rpc server
 if [ -n "${WAR_DEV+x}" ]; then # Dev mode selector
+    # Build image in local registry
+    docker build -t warnet/dev -f "$WAR_RPC/Dockerfile_dev" "$WAR_RPC" --load
     if [ "$(kubectl config current-context)" = "docker-desktop" ]; then
         sed "s?/mnt/src?$(pwd)?g" "$WAR_MANIFESTS/warnet-rpc-statefulset-dev.yaml" | kubectl apply -f -
-    else
+    else # assuming minikube
+        minikube image load warnet/dev
         kubectl apply -f "$WAR_MANIFESTS/warnet-rpc-statefulset-dev.yaml"
     fi
 else
--- a/resources/scripts/deploy.sh
+++ b/resources/scripts/deploy.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+
+if [ -z "${WAR_MANIFESTS+x}" ]; then
+    echo "WAR_MANIFESTS is unset. Please provide a path to warnet manifests."
+    exit 1
+fi
+
+# Function to check if warnet-rpc container is already running
+check_warnet_rpc() {
+    if kubectl get pods --all-namespaces | grep -q "bitcoindevproject/warnet-rpc"; then
+        echo "warnet-rpc pod found"
+        exit 1
+    fi
+}
+
+# Deploy base configurations
+kubectl apply -f "$WAR_MANIFESTS/namespace.yaml"
+kubectl apply -f "$WAR_MANIFESTS/rbac-config.yaml"
+kubectl apply -f "$WAR_MANIFESTS/warnet-rpc-service.yaml"
+
+# Deploy rpc server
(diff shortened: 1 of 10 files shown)


pinheadmz:
The k8s python lib we use might have a [create_from_yaml()](https://github.com/kubernetes-client/python/blob/master/kubernetes/utils/create_from_yaml.py) method which might replace the need for using python to execute `kubectl apply` in sub-processes. I say "might" because I'm having trouble finding documentation and haven't tried the method myself yet.

I would love to reduce the mix of shell and python as much as possible, ideally all the justfile commands can just be executed by warcli.

mplsgrant:
`warcli cluster` rocks.

`minikube` is wanting to use qemu seemingly b/c I use rootless docker. I can suggest a patch based on this: https://github.com/bitcoin-dev-project/warnet/pull/420/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R22

I am also getting an ErrImageNeverPull error in kubectl after doing a `cluster deploy`.

willcl-ark:
> `minikube` is wanting to use qemu seemingly b/c I use rootless docker. I can suggest a patch based on this: [#420 (files)](https://github.com/bitcoin-dev-project/warnet/pull/420/files#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1R22)

Added a commit with this fix, seems to work just the same for me but I hope fixes for you.
 
> I am also getting an ErrImageNeverPull error in kubectl after doing a `cluster deploy`.

Hmm, I don't know how you might be getting this. I guess it comes from statefulset-dev.yaml (where we spec to Never pull, as built image should be loaded in local cache). Are you able to provide repro steps/or configuration being used?

m3dwards:
I really like this. I think it could be improved with some slight changes to make it clear what's for prod and what's for local.

How about:

`warcli cluster setup-minikube`
`warcli cluster deploy --dev --hot-reload` (without --dev for production)

Without dev flag, just install the production image same tag as the source code
With dev flag, run `kubectl config current-context` to see if it's minikube or docker-desktop (or something else) to see how to do the file mounts.

willcl-ark:
OK @m3dwards @pinheadmz, I think this should be a fair bit closer to what you were imagining.

I did not add any `--hot` (reload) flags, as I think this can be done later. It look quite invasive to do, as we need to propagate the flag all the way into the dockerfile env... https://github.com/bitcoin-dev-project/warnet/blob/bf9eff11b2f5def4f8c727bc223901bab43435a4/resources/images/rpc/Dockerfile_dev#L29

I am sure it can be done, but not sure I have the stomach for it here 😋

willcl-ark:
@mplsgrant I kept the check for `$WAR_RPC` you removed in 8f49bd0779954eb76450c05e4aa949299d19f81e "remove unused rpc checks".

I am not actually sure why it seemed to work for me without it, but it should be needed to build the image: https://github.com/bitcoin-dev-project/warnet/pull/419/files#diff-444b1723cb5f02d3a7b6b4616e5e72f841e957f1c0973cc87fa4cb5cc06fdfb2R100

mplsgrant:
$WAR_RPC is set in [cluster.py](https://github.com/bitcoin-dev-project/warnet/pull/419/files#diff-b59cc9ca267a1649fe867c7f0c9e94bd2100433db5044a8d838b99b22514555cR9) and sent into the bash script as an env variable.

resources/scripts/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+# Deploy base configurations
+kubectl apply -f "$WAR_MANIFESTS/namespace.yaml"
+kubectl apply -f "$WAR_MANIFESTS/rbac-config.yaml"
+kubectl apply -f "$WAR_MANIFESTS/warnet-rpc-service.yaml"
+
+# Deploy rpc server
+if [ -n "${WAR_DEV+x}" ]; then # Dev mode selector
+    if [ "$(kubectl config current-context)" = "docker-desktop" ]; then
```
m3dwards, line 27: Should we build the dev container on this line?

m3dwards:
Tested both docker desktop and minikube and all appeared to work.

m3dwards (approved):

PR #421, opened 2024-08-01
make `pip install warnet` use prod

willcl-ark:
When using `warcli deploy` we will now use a production image from the current version, removing the requirement to mount the source directory into minikube, which is causing some issues for MacOS users.

* deploy tagged versions to hub.docker.com (only)
* Add a "version" to warnet. This let's us "sync" image version with python package version.
  * add CI test for this to ensure we update both in sync
* tidy up the CI: re-org the various workflows, and rename things where behaviour differs
* version bump to test it out?

Developers should continue to therefore use the justfile for local dev.

Note, this does naturally make testing these CLI commands slightly harder for deverlopers, but I think we can manage :)


.github/workflows/deploy.yml
```diff
@@ -17,44 +17,33 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build dev RPC image
         uses: docker/build-push-action@v5
         with:
           file: resources/images/rpc/Dockerfile_dev
           platforms: linux/amd64,linux/arm64
           context: src/warnet/templates/rpc
-          push: true
+          push: false
```
pinheadmz: 7257b9c9ccb7d82a00639de2a6a364f30eac61ee

Is the title of this commit missing the word "don't" ?

src/warnet/version.py
```diff
@@ -0,0 +1 @@
+VERSION = "0.9.11"
```
pinheadmz: e75b02d1a8bee542ba54a0ff10480e22e45f0b84

Seems like overkill to me to have the version string required in two places and compare them all the time. Is `pyproject.toml` always bundled with the rest of the code? Can't we just read the version directly from the toml file? We could add `warcli version` to the cli?

src/warnet/cli/cluster.py
```diff
@@ -95,7 +91,7 @@ def deploy():
     kubectl apply -f {MANIFEST_PATH}/namespace.yaml
     kubectl apply -f {MANIFEST_PATH}/rbac-config.yaml
     kubectl apply -f {MANIFEST_PATH}/warnet-rpc-service.yaml
-    kubectl apply -f {MANIFEST_PATH}/warnet-rpc-statefulset-dev.yaml
+    sed "s|bitcoindevproject/warnet-rpc:latest|bitcoindevproject/warnet-rpc:{VERSION}|" {MANIFEST_PATH}/warnet-rpc-statefulset.yaml | kubectl apply -f -
```
pinheadmz, line 94: 299122363f350ad87b115ca84936eece37d08bd3

making me wonder again if python-embedded-bash (plus external yaml files) can be simplified with just... python.

https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/AppsV1Api.md#create_namespaced_stateful_set

pinheadmz: or maybe `warnet-rpc-statefulset.yaml` should just always have its own version hard-coded? I'm trying to think if we'd ever need `latest` in prod?

pinheadmz:
doesnt apply to helm model

PR #422, opened 2024-08-01
Replacement Cycling with split

mplsgrant:
## Issue
Need to run Replacement Cycling and allow interactivity

## Solution
Split the Replacement Cycling demo into its constituent parts. Allow monitoring with zmq.

## Still needs
Scenario test

## Supersedes
https://github.com/bitcoin-dev-project/warnet/pull/372

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -12,3 +12,4 @@ requests<2.30
 rich
 tabulate
 PyYAML
+pyzmq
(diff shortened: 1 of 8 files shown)


instagibbs:
@mplsgrant curious how much has the scenario changed since I last looked?

mplsgrant:
@instagibbs The mechanics have stayed exactly the same. What's changed is splitting the steps apart so we can run them out of order.

josibake:
could be nice to see if this "just works" with the new commander launcher

mplsgrant:
@josibake I agree, but I'm going to leave this on the back burner for a bit, and then revisit.

Issue #353, opened 2024-05-02
warcli: network commands report errors when no network is running


carlaKC:
This is just aesthetic/doesn't impact usage, but it would be nice to have friendlier errors when you run network status/info with no network running. 

`warcli network info`
-> `-32000: Server error`

`warcli network status` 
-> `-32000: Error getting network status: [Errno 2] No such file or directory: '/root/.warnet/warnet/warnet/graph.graphml'`

pinheadmz:
This is not an issue since we switched to serverless + helm:

```

(.venv) --> warnet status
╭────────────── Warnet Overview ───────────────╮
│                                              │
│                Warnet Status                 │
│ ┏━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓ │
│ ┃ Component ┃ Name                ┃ Status ┃ │
│ ┡━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩ │
│ │ Scenario  │ No active scenarios │        │ │
│ └───────────┴─────────────────────┴────────┘ │
│                                              │
╰──────────────────────────────────────────────╯

Total Tanks: 0 | Active Scenarios: 0
Network connected
```

PR #426, opened 2024-08-10
streamline script outputs

mplsgrant:
### Issue
Users run into unfriendly messages from the cli interface such as in issue #353.

### Cause
When the user enters unexpected commands or parameters, there is no code to handle those cases.

### Solution
Add code to handle exceptions and introduce helpful error messages.

### DIFF
--- a/src/warnet/cli/graph.py
+++ b/src/warnet/cli/graph.py
@@ -3,6 +3,7 @@ import os
 import sys
 from io import BytesIO
 from pathlib import Path
+from xml.etree.ElementTree import ParseError
 
 import click
 import networkx as nx
@@ -128,6 +129,10 @@ def validate(graph: Path):
     """
     Validate a <graph file> against the schema.
     """
-    with open(graph) as f:
-        graph = nx.parse_graphml(f.read(), node_type=int, force_multigraph=True)
+    try:
+        with open(graph) as f:
+            graph = nx.parse_graphml(f.read(), node_type=int, force_multigraph=True)
+    except ParseError as e:
+        print(f"Error. {e}")
+        sys.exit(1)
     return validate_graph_schema(graph)
--- a/src/warnet/cli/graph.py
+++ b/src/warnet/cli/graph.py
@@ -1,4 +1,6 @@
 import json
+import os
+import sys
 from io import BytesIO
 from pathlib import Path
 
@@ -47,6 +49,10 @@ def import_json(infile: Path, outfile: Path, cb: str, ln_image: str):
     channels and their policies as well.
     Returns XML file as string with or without --outfile option.
     """
+    if not os.path.exists(infile):
+        print(f"Error. {infile} does not exist")
+        sys.exit(1)
+
     with open(infile) as f:
         json_graph = json.loads(f.read())
 
--- a/src/warnet/cli/graph.py
+++ b/src/warnet/cli/graph.py
@@ -2,6 +2,7 @@ import json
 import os
 import sys
 from io import BytesIO
+from json import JSONDecodeError
 from pathlib import Path
 from xml.etree.ElementTree import ParseError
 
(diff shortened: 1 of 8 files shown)


mplsgrant:
Let's pay attention to whether this gives 'doubling up'/debug output on the ""⏲️ This could take a minute or so." message.

src/warnet/cli/rpc.py
```diff
@@ -27,6 +27,8 @@ def rpc_call(rpc_method, params: dict[str, Any] | tuple[Any, ...] | None):
         logging.error(
             f"Error connecting to {url}. Is the server running and using the correct URL?"
         )
+        print("\nPlease consider running the following warnet command:")
+        print("  warcli cluster deploy\n")
```
pinheadmz: 32c8f1ceab03a815693093483041529cc3f40291

why logging.error above but print() here? also maybe instead of hard coding a command that might change, just something like "is there a cluster deployed?"

src/warnet/cli/network.py
```diff
@@ -108,7 +108,8 @@ def info(network: str):
     Get info about a warnet named [network]
     """
     result = rpc_call("network_info", {"network": network})
-    print_repr(result)
+    if result:
+        print_repr(result)
```
pinheadmz, line 113: 75185fb0eb26238e65ac78b33005daf0a9f16790

I think RPC calls should always return *something*

src/warnet/cli/graph.py
```diff
@@ -128,6 +129,12 @@ def validate(graph: Path):
...
     Validate a <graph file> against the schema.
     """
-    with open(graph) as f:
-        graph = nx.parse_graphml(f.read(), node_type=int, force_multigraph=True)
+    try:
+        with open(graph) as f:
+            graph = nx.parse_graphml(f.read(), node_type=int, force_multigraph=True)
+    except ParseError as e:
+        print(f"Error. {e}")
+        print("The graph file appears malformed. Pleases consider generating a graph file with:")
```
pinheadmz: 10ecff1afbb2ce2efe91044cd5c4c872219c68f1

nit: "Please"

src/warnet/cli/graph.py
```diff
@@ -128,6 +129,12 @@ def validate(graph: Path):
...
+        print("  warcli graph create")
```
pinheadmz: 10ecff1afbb2ce2efe91044cd5c4c872219c68f1

again i think hard coding suggested commands in error messages could lead to issues

pinheadmz (commented):
looks good, thanks good work! left a few comments

pinheadmz:
Tested the scripts and errors locally, *nice*

pinheadmz:
This is a weird one, though. From `warcli cluster teardown`:


```
W0820 10:27:24.114029   79378 main.go:291] Unable to resolve the current Docker CLI context "default": context "default": context not found: open /Users/matthewzipkin/.docker/contexts/meta/37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f/meta.json: no such file or directory
```

src/warnet/cli/rpc.py
mplsgrant: Removed command mention; uses `logging` now: https://github.com/mplsgrant/warnet/blob/a68a229af6e90e4111d1ef6461cc8670cc1cdce8/src/warnet/cli/rpc.py#L28

src/warnet/cli/network.py
mplsgrant: Added handling here: https://github.com/mplsgrant/warnet/blob/a68a229af6e90e4111d1ef6461cc8670cc1cdce8/src/warnet/cli/network.py#L114

src/warnet/cli/graph.py
mplsgrant: Removed here: https://github.com/mplsgrant/warnet/blob/a68a229af6e90e4111d1ef6461cc8670cc1cdce8/src/warnet/cli/graph.py#L146

mplsgrant:
> This is a weird one, though. From `warcli cluster teardown`:

@pinheadmz I'd like to punt on this and open an issue.

mplsgrant:
YAML fixes this.

PR #429, opened 2024-08-15
scenarios: cli: show options in help

willcl-ark:
We do have this, but available options get out of sync. Fix this by reading options directly from the Class.

TODO: improve error handling.

### DIFF
--- a/src/warnet/cli/scenarios.py
+++ b/src/warnet/cli/scenarios.py
@@ -5,7 +5,7 @@ import sys
 import click
 from rich import print
 from rich.console import Console
-from rich.table import Table
+from rich.table import Table, Text
 
 from .rpc import rpc_call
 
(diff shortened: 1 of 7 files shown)


willcl-ark:
Do this by adding a new method to test_framework_bridge to generate the new help.

Before:

image

After:

image

src/warnet/cli/scenarios.py
```diff
@@ -22,18 +22,31 @@ def available():
...
-        sys.exit(1)
 
-    # Create the table
-    table = Table(show_header=True, header_style="bold")
-    table.add_column("Name")
-    table.add_column("Description")
+    # main table
+    main_table = Table(show_header=True, header_style="bold magenta")
+    main_table.add_column("Name", style="cyan", no_wrap=True)
+    main_table.add_column("Description", style="green")
```
josibake, line 29: nit: specifying the colours and styles seems like it might cause issues on depending on who is using it? is there a way to just inherit some sane defaults from the user?

src/warnet/server.py
```diff
@@ -325,8 +326,22 @@ def scenarios_available(self) -> list[tuple]:
...
+                            scenario = obj()
+                            help_info = scenario.cli_help()
+                            scenario_list.append(
+                                {
+                                    "name": s.name,
+                                    "help_text": help_info["help_text"],
+                                    "options": help_info["options"],
+                                }
+                            )
+                            break  # Assuming one scenario class per module
```
josibake, line 344: is this something we can enforce?

josibake (commented):
ACK, nice! left some nits but can probably merge this in favour of the other PR where you are updating the help text manually.

Only concern I have is hard coding some style params, since I imagine this will look weird on certain setups.

src/warnet/server.py
willcl-ark: We can really only call a single TestFrameWork class per scenario, so I don't think it should be a big issue?

src/warnet/cli/scenarios.py
willcl-ark: mmmm, yeah, I could possible make it look more default (worse) 😛

willcl-ark: These colors come from our CLI lib though, so I think they should handle it ok

pinheadmz:
Couldn't get this to work, may need a different approach. Kept getting weird "unrecognized option --dev"

PR #430, opened 2024-08-15
various fixups

willcl-ark:
Some fixups from josie's feedback.

- tx_flood doc tidy will conflict with #429 and prob should be dropped in favour of that

### DIFF
--- a/docs/quickrun.md
+++ b/docs/quickrun.md
@@ -51,7 +51,7 @@ warcli cluster deploy
 
 ...to deploy warnet to your cluster.
 
-`warcli deploy` also automatically configures port forwarding to the Server in the cluster.
+`warcli cluster deploy` also automatically configures port forwarding to the Server in the cluster.
 
 ## Stopping
 
--- a/src/warnet/cli/cluster.py
+++ b/src/warnet/cli/cluster.py
@@ -22,8 +22,7 @@ def cluster():
     """Start, configure and stop a warnet k8s cluster\n
     \b
     Supports chaining, e.g:
-      warcli cluster deploy
-      warcli cluster teardown
+      warcli cluster deploy-logging connect-logging
     """
     pass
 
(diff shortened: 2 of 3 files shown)


src/warnet/cli/cluster.py
```diff
@@ -20,7 +20,8 @@ def cluster():
     """Start, configure and stop a warnet k8s cluster\n
     \b
     Supports chaining, e.g:
-      warcli cluster deploy
+      warcli cluster setup-minikube deploy
+      warcli cluster deploy-logging connect-logging
```
josibake, line 25: nice!

src/warnet/scenarios/tx_flood.py
```diff
@@ -8,7 +8,7 @@
 
 
 def cli_help():
-    return "Make a big transaction mess"
+    return "Make a big transaction mess. Options: [--interval ]"
```
josibake: looks like this is a manual process? would be nice if we could automatically pipe through all of the commands, otherwise I suspect we will keep running into this.

josibake (approved):
ACK, lgtm

josibake:
Can take this out of draft?

src/warnet/scenarios/tx_flood.py
willcl-ark: See #429

josibake: Seems better to drop this change from this PR and merge #429 ?  I left a quick review on 429 but my comments are non-blocking

pinheadmz: oops i did just hard code the option on main, couldnt get #429  to work

docs/quickrun.md
```diff
@@ -51,7 +51,7 @@ warcli cluster deploy
 
 ...to deploy warnet to your cluster.
 
-`warcli deploy` also automatically configures port forwarding to the Server in the cluster.
+`warcli cluster deploy` also automatically configures port forwarding to the Server in the cluster.
```
pinheadmz, line 54: may conflict with Jonas' #454

PR #440, opened 2024-08-15
fix deploy CI job

willcl-ark:
I think this should run on PRs now, but only push on merge to main...

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,4 @@
 name: deploy
-
 on:
   workflow_run:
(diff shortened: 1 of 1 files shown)


mplsgrant:
LGTM  🖒

[This appears to build](https://github.com/mplsgrant/warnet/actions/runs/10409555980) on my end with the right context location.

~~Oops. Looks like I did have some kind of [build issue](https://github.com/mplsgrant/warnet/actions/runs/10409200310/workflow) (however, that might be related to my *existing* CI process)~~

Issue #441, opened 2024-08-16
bug: image building on M3 macOS failing


josibake:
Steps taken:

1. Create a venv, `python3 -m venv .warnet`
2. Activate and install warnet, `pip install warnet`
3. Install docker, `brew install --cask docker`
4. Run the image build command, `warcli image build --repo=bitcoin/bitcoin --commit-sha=99ecb9a630e68f7fa5a449065020292790e3e7fb --registry=josibake/warnet --tag master --action=push`

Image starts to build and eventually fails with the following output:

```
148.8   CXX      wallet/libbitcoin_wallet_a-walletutil.o
160.3 g++: fatal error: Killed signal terminated program cc1plus
160.3 compilation terminated.
160.4 make[2]: *** [Makefile:10298: libbitcoin_node_a-init.o] Error 1
160.4 make[2]: *** Waiting for unfinished jobs....
183.6 g++: fatal error: Killed signal terminated program cc1plus
183.6 compilation terminated.
183.6 make[2]: *** [Makefile:10438: libbitcoin_node_a-net_processing.o] Error 1
228.4 make[2]: Leaving directory '/build/bitcoin/src'
228.4 make[1]: *** [Makefile:20311: all-recursive] Error 1
228.4 make[1]: Leaving directory '/build/bitcoin/src'
228.4 make: *** [Makefile:749: all-recursive] Error 1
------
Dockerfile:34
--------------------
  33 |
  34 | >>> RUN set -ex \
  35 | >>>     && cd /build \
  36 | >>>     && git clone --depth 1 "https://github.com/${REPO}" \
  37 | >>>     && cd bitcoin \
  38 | >>>     && git fetch --depth 1 origin "$COMMIT_SHA" \
  39 | >>>     && git checkout "$COMMIT_SHA" \
  40 | >>>     && git apply /tmp/isroutable.patch \
  41 | >>>     && git apply /tmp/addrman.patch \
  42 | >>>     && sed -i s:sys/fcntl.h:fcntl.h: src/compat/compat.h \
  43 | >>>     && ./autogen.sh \
  44 | >>>     && ./configure \
  45 | >>>     LDFLAGS=-L`ls -d /opt/db*`/lib/ \
  46 | >>>     CPPFLAGS=-I`ls -d /opt/db*`/include/ \
  47 | >>>     --prefix=${BITCOIN_PREFIX} \
  48 | >>>     ${BUILD_ARGS} \
[...]

PR #444, opened 2024-08-16
bugfix: disable dependency tracking

josibake:
While building on MacOS, the build would randomly fail sometimes with the following error: 

```
49.41 config.status: executing depfiles commands
53.72 config.status: error: in '/build/bitcoin':
53.72 config.status: error: Something went wrong bootstrapping makefile fragments
53.72     for automatic dependency tracking.  If GNU make was not used, consider
53.72     re-running the configure script with MAKE="gmake" (or whatever is
53.72     necessary).  You can also try re-running configure with the
53.72     '--disable-dependency-tracking' option to at least be able to build
53.72     the package (albeit without support for automatic dependency tracking).
53.72 See 'config.log' for more details
```

This PR adds the `--disable-dependency-tracking` option suggested in the error message. Once I added this, I've built several more times without running into the previous error message and I'm fairly confident adding this build option doesn't have any side effects for our use case.

Related to #441

### DIFF
--- a/src/warnet/cli/image_build.py
+++ b/src/warnet/cli/image_build.py
@@ -24,7 +24,7 @@ def build_image(
     action: str,
 ):
     if not build_args:
-        build_args = '"--disable-tests --without-gui --disable-bench --disable-fuzz-binary --enable-suppress-external-warnings "'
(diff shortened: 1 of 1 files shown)


pinheadmz:
LGTM, testing locally...

pinheadmz:
goog.

PR #445, opened 2024-08-16
add kubectl and helm checks to setup-minikube

mplsgrant:
### Issue
When users run `setup-minikube` it does not verify they have kubectl or helm, but in our workflow we need those tools. 

### Solution
Allow the `setup-minikube` command to check for helm and kubectl because minikube will want to use kubectl right away, and helm is needed for logging later on. Prompt the user to install them with helpful links.

### DIFF
--- a/resources/scripts/setup_minikube.sh
+++ b/resources/scripts/setup_minikube.sh
@@ -114,6 +114,8 @@ fi
 # Start minikube with the constructed command
 eval "$MINIKUBE_CMD"
 
+echo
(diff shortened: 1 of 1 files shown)


willcl-ark:
Looks good to me -- detected missing all bins (as expected).

Perhaps one thing we could discuss this week:

- Can we develop one (or two) main "supported workflow paths" which we will strive to support, document and have work well. For everything else user can be on their own...


  1. User: (Only `pip install`, `warcli` commands, only via minikube)
  2. Developer: (clone repo, any k8s backend)

PR #456, opened 2024-08-19
set resource requests and limits on tank object

josibake:
tell k8s up front the min and max CPU, memory we want for a tank. can be overriden from a graph file if a tank needs more/less.

did some testing, and this allows k8s to do autoscaling when trying to start a network. this means if a cluster does not have enough nodes and you try to start a network with 50 tanks, k8s will keep the pods in pending until it adds more nodes to the cluster. this allows for much more graceful cluster management.

annoyingly, if the pod is already scheduled and the workload of the pod increases, k8s can't do cluster autoscaling anymore, but there might be another way to tackle this via vertical scaling. as of today, the best way to handle this would be to set a much higher `limits`, which should help? (not tested).

## testing

tested on digital ocean cluster with autoscaling, has not been tested on Minikube or docker desktop, but _should_ work since requests and limits are not k8s specific config options.

### DIFF
--- a/src/warnet/utils.py
+++ b/src/warnet/utils.py
@@ -421,15 +421,8 @@ def create_cycle_graph(n: int, version: str, bitcoin_conf: str | None, random_ve
                 conf_contents = dump_bitcoin_conf(conf_dict, for_graph=True)
 
     # populate our custom fields
-    for i, node in enumerate(graph.nodes()):
-        if random_version:
-            graph.nodes[node]["version"] = random.choice(WEIGHTED_TAGS)
-        else:
-            # One node demoing the image tag
-            if i == 1:
-                graph.nodes[node]["image"] = f"bitcoindevproject/bitcoin:{version}"
-            else:
-                graph.nodes[node]["version"] = version
+    for node in graph.nodes():
+        graph.nodes[node]["image"] = f"bitcoindevproject/bitcoin:{version}"
         graph.nodes[node]["bitcoin_config"] = conf_contents
         graph.nodes[node]["tc_netem"] = ""
         graph.nodes[node]["build_args"] = ""
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,8 @@ jobs:
       - uses: azure/setup-helm@v4.2.0
       - uses: medyagh/setup-minikube@master
         with:
+          cpus: max
+          memory: 10000m
           mount-path: ${{ github.workspace }}:/mnt/src
       - uses: actions/download-artifact@v4
         with:
--- a/src/warnet/backend/kubernetes_backend.py
+++ b/src/warnet/backend/kubernetes_backend.py
@@ -384,33 +384,62 @@ class KubernetesBackend:
         bitcoind_options = tank.get_bitcoin_conf(peers)
         container_env = [client.V1EnvVar(name="BITCOIN_ARGS", value=bitcoind_options)]
 
-        bitcoind_container = client.V1Container(
-            name=container_name,
-            image=container_image,
-            env=container_env,
-            liveness_probe=client.V1Probe(
-                failure_threshold=3,
-                initial_delay_seconds=5,
-                period_seconds=5,
-                timeout_seconds=1,
-                _exec=client.V1ExecAction(command=["pidof", "bitcoind"]),
-            ),
-            readiness_probe=client.V1Probe(
-                failure_threshold=1,
-                initial_delay_seconds=0,
-                period_seconds=1,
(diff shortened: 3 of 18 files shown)


m3dwards:
Ran on Docker Desktop's Kubernetes and could see:

```shell
"resources": {
    "limits": {
        "cpu": "1",
        "memory": "1500Mi"
    },
    "requests": {
        "cpu": "500m",
        "memory": "500Mi"
    }
},
```

m3dwards:
CI failed, perhaps because minikube doesn't think it has enough resources to start the tanks

willcl-ark:
Added a commit to
- fix schema
- validate generated graphs during creation
- apply default resource limits to all created graphs (@josibake do we want this?)

willcl-ark:
An alternative using user-friendly "profiles": https://github.com/willcl-ark/warnet/pull/24

src/warnet/utils.py
```diff
@@ -380,7 +380,9 @@ def set_execute_permission(file_path):
     os.chmod(file_path, current_permissions | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
 
-def create_cycle_graph(n: int, version: str, bitcoin_conf: str | None, random_version: bool):
+def create_cycle_graph(
+    n: int, version: str, bitcoin_conf: str | None, random_version: bool, resources: dict | None
```
pinheadmz: `resources` isnt being used in the function?

pinheadmz (commented):
Can't we just set lower resource usage in the CI so the tests can run as written without any changes?

pinheadmz:
Error on docker desktop, `warcli network start` (so, default.graphml)

```
2024-08-20 12:04:09 | DEBUG   | k8s      | Deploying pods
2024-08-20 12:04:09 | DEBUG   | k8s      | Creating bitcoind container for tank 0
2024-08-20 12:04:09 | ERROR   | server   | Unhandled exception starting warnet: 'NoneType' object is not subscriptable
  File "/root/warnet/src/warnet/server.py", line 469, in thread_start
    wn.warnet_up()
  File "/root/warnet/src/warnet/warnet.py", line 176, in warnet_up
    self.container_interface.up(self)
  File "/root/warnet/src/warnet/backend/kubernetes_backend.py", line 57, in up
    self.deploy_pods(warnet)
  File "/root/warnet/src/warnet/backend/kubernetes_backend.py", line 682, in deploy_pods
    bitcoind_container = self.create_bitcoind_container(tank)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/warnet/src/warnet/backend/kubernetes_backend.py", line 410, in create_bitcoind_container
    requests=tank.resources["requests"],
             ~~~~~~~~~~~~~~^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable


```

pinheadmz:
I *think* the bug Im getting is in the default set in schema.json

pinheadmz:
example fix:

```diff
diff --git a/src/warnet/graph_schema.json b/src/warnet/graph_schema.json
index aa1518f..4b178a4 100644
--- a/src/warnet/graph_schema.json
+++ b/src/warnet/graph_schema.json
@@ -69,25 +69,13 @@
         "comment": "A string of arguments for the lightning network node in command-line format, e.g. '--protocol.wumbo-channels --bitcoin.timelockdelta=80'"
       },
       "resources": {
-            "type": "object",
-            "comment": "Kubernetes resource requests and limits for the node",
-            "properties": {
-              "requests": {
-                "type": "object",
-                "properties": {
-                  "cpu": {"type": "string", "default": "500m"},
-                  "memory": {"type": "string", "default": "500Mi"}
-                }
-              },
-              "limits": {
-                "type": "object",
-                "properties": {
-                  "cpu": {"type": "string", "default":"1000m"},
-                  "memory": {"type": "string", "default":"1500Mi"}
-                }
-              }
-            }
+        "type": "object",
+        "comment": "Kubernetes resource requests and limits for the node",
+        "default": {
+          "requests": {"cpu": "100m", "memory": "250Mi"},
+          "limits": {"cpu": "250m", "memory": "500Mi"}
         }
+      }
     },
     "additionalProperties": false,
     "oneOf": [
```

josibake:
Closing as this is now obsolete. We can now configure everything in one place using the network.yaml/node-defaults.yaml and namespaces.yaml/namespace-defaults.yaml helm files.

Only thing currently missing is good documentation and examples on how to use the yaml configuration files which I think we will need before tabconf, opened a backlog issue for that here: https://github.com/bitcoin-dev-project/warnet/issues/496

PR #460, opened 2024-08-19
added 100 nodes 15ps graphfile for tx relay

m3dwards:


pinheadmz:
tested graph on 100 vCPU cluster, works.

PR #461, opened 2024-08-19
tune gcc to be less resource hungry

### LINKED ISSUE #441: bug: image building on M3 macOS failing

josibake:
fix OOM on docker desktop when building custom images. takes the recommendations from https://github.com/bitcoin/bitcoin/tree/master/doc/build-unix.md

fixes #441

### DIFF
--- a/resources/images/bitcoin/Dockerfile
+++ b/resources/images/bitcoin/Dockerfile
@@ -43,7 +43,7 @@ RUN set -ex \
     && ./autogen.sh \
     && ./configure \
     LDFLAGS=-L`ls -d /opt/db*`/lib/ \
-    CPPFLAGS=-I`ls -d /opt/db*`/include/ \
+    CPPFLAGS="-g0 -I`ls -d /opt/db*`/include/ --param ggc-min-expand=1 --param ggc-min-heapsize=32768" \
     --prefix=${BITCOIN_PREFIX} \
     ${BUILD_ARGS} \
     && make -j$(nproc) \


josibake:
cc @willcl-ark since you've done most of the image building stuff. any objections to merging this?

willcl-ark:
I don't understand why this should be necessary.

Does this happen on your machine with building the same commit hash outside of docker? It should do presumably (but I would be surprised if it does). Did you start docker with less than 1.5GB of RAM allocated? If so, my view would be that that is what should change, not compilation settings here. I don't think we should take many steps to support e.g. building custom bitcoin core warnet images on a raspi (because we just don't need to support that).

The garbage collection options slow down building a fair bit (and really shouldn't be needed on modern hardware) so would prefer not to modify `ggc-min-expand` and `ggc-min-heapsize`, but would be very happy to not build debug symbols (that we later strip).

willcl-ark:
udpate: did some more testing and managed to get the same failure on my macbook.

Happy to merge as it may save a future user some debugging in the future (and I'm using raw `docker buildx` commands to build our own images anyhow)

josibake:
> Happy to merge as it may save a future user some debugging

fwiw, I hit the failure immediately when I first used warnet to build an image on my Mac. this left me with the impression that warnet was just broken and led me down a rabbit hole trying to troubleshoot the docker images/my docker setup/etc. overall this cost me about 3 hours of debugging. I suspect it would be even worse for someone less familiar with docker/warnet and leave them with the impression the feature is just broken.


more of a meta point as to why prefer this approach:

the only other approach I see to problems like this is document everything, but people don't read the docs. in general, I think we should strive to make warnet work "out of the box" without requiring the user to make resource changes (as much as possible!). for everything else, we should try to make warnet fail in super obvious ways.

Issue #458, opened 2024-08-19
better configuration management


josibake:
today, any change I make requires re-building the rpc server image, pushing or loading it into a cluster, etc. if we move more of the configuration out of the rpc server and make it more the responsibility of the CLI , we could change config options without needing to redeploy the RPC server.

A few examples:

- tank resources, cpu and memory
- registry to pull from (e.g., bitcoindevproject)

mplsgrant:
Another config item:

* set minikube vs docker-desktop

Possible rules for the config:
1. Any item set in the config file cannot be changed.
2. Any item not sent in the config file maybe set later, but once set, it cannot be changed.

mplsgrant:
I think YAML solves this. Also, prompting the user for minikube vs docker-desktop in quickstart.

PR #463, opened 2024-08-19
Add a config file 

mplsgrant:
### Issue
Need a way to hold warnet config elements.
See issue: [#458](https://github.com/bitcoin-dev-project/warnet/issues/458)

### Solution
Use a yaml file to hold values. 

Rules for the config:

    Any item set in the config file cannot be changed.
    Any item not sent in the config file maybe set later, but once set, it cannot be changed.

### DIFF
--- a/resources/configs/warcli.yaml
+++ b/resources/configs/warcli.yaml
@@ -1,0 +1,1 @@
+backend:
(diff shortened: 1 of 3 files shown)


mplsgrant:
YAML fixes this

Issue #464, opened 2024-08-19
Investigate K8s resource management / rate limiting


willcl-ark:
Avoid taking down whole network when attacking, instead rate-limit attackers?

josibake:
per my comment https://github.com/bitcoin-dev-project/warnet/pull/465#issuecomment-2309653056, perhaps using resource quotas at the namespace level would work? cc @m3dwards

josibake:
Adding this as a milestone for Tabconf. We will need to either a) be sure this is not an issue or b) have some idea of how to limit resources in a namespace to prevent users from borking the cluster.

m3dwards:
~We can give commanders (and tanks) resource limits which I think would cover us?~

No of course that won't work as you could just spin up many commanders. Happy to look into namespace resource limits.

willcl-ark:
I still don't understand particuarly what other thing we are looking for here that's not #465?

Just some other native k8s way of doing it? Or are you guys trying to only limit cpu/memory resources (which we already have, no?)

josibake:
Right, I think this issue is a bit confusingly named. So starting with the problem: we won't want people to bring down the network by doing expensive attacks. To me, the most obvious way to do this is limit the resources any single node has access to (via our current methods of `requests/limits`). You can't really blast out tons of network calls if your CPU is being throttled. Secondly, we can place resource limits on the namespace, such as "blue team can only have up to X total CPU and memory in this namespace."

This feels more straightforward to me than trying to implement networking policies on the namespaces and more reflective of the real world. I am very uncomfortable with any solution that involves fiddling with network traffic as it's much harder, imo, to reason about and can have weird side effects.

@willcl-ark perhaps you can provide a better justification for why you think network rate limiting is the right approach here and why namespace limits wouldn't solve the problem?

willcl-ark:
My understanding was that k8s already natively handles CPU and memory resources, everything is vCPU (and I guess "vRAM" too), so we already can't exhaust a "real" cluster. The story may be different on a simulated single-machine cluster. Additionally we **could** artificially resource contstrain nodes too, if desired.

Therefore I was trying to address someone saturating network bandwidth, as AFAIK this is unconstrained.

My worry if we resource (CPU/memory) throttle nodes, one could lock them up by simply sending them so much work to do that they will never finish it. This would not be the spirit of the "crash" we are attempting to have folks demonstrate.

If we alternatively throttle bandwidth, it is the attacker who will find themselves constrained, not the node.

[...]

PR #465, opened 2024-08-20
add global intra-pod rate-limiter

willcl-ark:
Rate-limit traffic between pods.

Folks doing DoS attacks might otherwise [take|bog] down entire network.

### DIFF
--- a/resources/manifests/istio-global-rate-limit.yaml
+++ b/resources/manifests/istio-global-rate-limit.yaml
@@ -1,0 +1,62 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: global-ratelimit
+  namespace: istio-system
+spec:
+  configPatches:
+    - applyTo: NETWORK_FILTER
+      match:
+        context: ANY
+        listener:
+          filterChain:
+            filter:
+              name: "envoy.filters.network.tcp_proxy"
+      patch:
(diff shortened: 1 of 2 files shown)


willcl-ark:
I think we might want something like this, not sure if it can be done in a more lightweight way

adamjonas:
ref #464

josibake:
Another way to think about this would be defining `ResourceQuotas` directly on the namespace. I haven't researched this too much, but seems cleaner to me to say something like "This namespace can have at most X pods, X total CPU and X total RAM." within that, you're allowed to do whatever you want.

So yeah, you could still do a ton of network stuff, but only as much as your CPU allows because as soon as your pods start asking for more CPU than the namespace is allowed, you get CPU throttled. The advantages of this approach is that we _already_ need to set resource quotas on namespaces to get war games properly working, so seems simpler to reuse the same construction.

josibake:
Suggest we close this for now or put in draft until we've done some more investigation on #464 . Also, doesn't seem strictly necessary for Tabconf?

willcl-ark:
> Suggest we close this for now or put in draft until we've done some more investigation on #464 . Also, doesn't seem strictly necessary for Tabconf?

We added the tabconf label to #464 though? IIUC `resources` is not the same as (network) rate-limiting containers.

willcl-ark:
Closing for now

PR #468, opened 2024-08-20
Add stderr to scenario output

m3dwards:
If a scenario fails on startup or otherwise prints to stderr we can't see the output in the RPC logs

How to test: Write a scenario that has a coding error in it, run the scenario. Before PR there will be no output in the logs, after PR you will see the problem.


pinheadmz:
Yes 👍

PR #480, opened 2024-08-22
refactor out from cli

willcl-ark:


pinheadmz:
i think the commander dockerfile may need to be updated as well

PR #486, opened 2024-08-23
regtestold for 0.16

### LINKED ISSUE #484: bitcoincore/templates/pod.yaml: remove "[regtest]" from conf
older versions of bitcoin core (e.g. 0.16) wont parse the conf file

pinheadmz:
closes https://github.com/bitcoin-dev-project/warnet/issues/484

### DIFF
--- a/resources/charts/bitcoincore/templates/configmap.yaml
+++ b/resources/charts/bitcoincore/templates/configmap.yaml
@@ -9,6 +9,9 @@ data:
     {{- if eq .Values.chain "regtest" }}
       {{- tpl .Values.regtestConfig . | nindent 4 }}
     {{- end }}
+    {{- if eq .Values.chain "regtestold" }}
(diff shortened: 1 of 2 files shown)


pinheadmz:
Actually can do a lot better than this just adding a default false flag "oldconfig" and maybe auto setting that if we can parse the version

josibake:
Another idea would be to manage the backwards compatibility issues in the Helm chart. Helm already has conceptual building blocks for this (e.g. being able to compare semvers in the template language), so we our users are always using the most up to date way of doing things and then in the templates we back port that when creating the chart on a tag by tag basis.

Proposing this because this is likely to come up again so having a more general way to handle it seems nice?

pinheadmz:
https://docs.helm.sh/docs/chart_template_guide/function_list/#semantic-version-functions

SWEET. closing this, will do it properly in new PR

PR #491, opened 2024-08-25
Don't restart bitcoin pods by default

### LINKED ISSUE #438: Turn off auto-restart
During a battle on warnet, we want the nodes that go down to stay down. 

How do we turn off auto-restart for those nodes?

Or, do we monitor "uptime" in grafana and look for drop-to-0?

willcl-ark:
Fixes #438

before:
image

after:
image

ping @m3dwards

### DIFF
--- a/resources/charts/bitcoincore/templates/pod.yaml
+++ b/resources/charts/bitcoincore/templates/pod.yaml
@@ -14,6 +14,7 @@ metadata:
   annotations:
     init_peers: "{{ .Values.connect | len }}"
 spec:
+  restartPolicy: "{{ .Values.restartPolicy }}"
   {{- with .Values.imagePullSecrets }}
   imagePullSecrets:
     {{- toYaml . | nindent 4 }}
--- a/resources/charts/bitcoincore/values.yaml
+++ b/resources/charts/bitcoincore/values.yaml
@@ -3,6 +3,8 @@
 # Declare variables to be passed into your templates.
 namespace: warnet
 
+restartPolicy: Never
+
 image:
   repository: bitcoindevproject/bitcoin
   pullPolicy: IfNotPresent


pinheadmz:
ACK tested by adding `config: illegaloption` to cause initerror, tank didnt restart.

Also `rpc stop` resulted in `completed` status with no restart, not sure if thats changed

resources/charts/bitcoincore/templates/pod.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
   annotations:
     init_peers: "{{ .Values.connect | len }}"
 spec:
+  restartPolicy: Never
```
josibake: I _think_ the preferred approach is to always add these to values.yaml ? i.e., values.yaml contains all the things we expect the user to change , and provides the defaults. the actual resource .yaml files and .tpl files are then strictly used to _apply_ the user provided / default values to the template. Could be wrong on this so might want to get a second opinion from @m3dwards / the helm docs.

willcl-ark: Yes, just on looks, this appears out of place in this file, and looks like it would fit in Values.yml

willcl-ark: OK perhaps I have it right now in 8e8d7d1510e3464ea5ab00420f68208a1a68ee27

josibake: nice, looks much better!

josibake:
code review ACK https://github.com/bitcoin-dev-project/warnet/pull/491/commits/8e8d7d1510e3464ea5ab00420f68208a1a68ee27

willcl-ark:
If @m3dwards could confirm this is good, we can merge this too. :)

m3dwards (approved):

m3dwards:
LGTM

PR #492, opened 2024-08-25
Don't error on bringing down a pod which is not found

willcl-ark:
Currently `warcli down` can produce a lot of (incorrect) errors, fix these by ignoring not-found pods

image


josibake:
post merge ACK, nice!

PR #494, opened 2024-08-25
format bitcoin.conf based on semantic versioning

### LINKED ISSUE #484: bitcoincore/templates/pod.yaml: remove "[regtest]" from conf

pinheadmz:
Closes #484 

This only really affects Bitcoin Core versions older than 0.17 but we have a use case for 0.16 and support it...

Using helm's built in [semver](https://docs.helm.sh/docs/chart_template_guide/function_list/#semver) parsing abilities we check if the image tag represents a version older than 0.17 when [conf sections were added](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.17.0.md#configuration-sections-for-testnet-and-regtest).

The method added to `_helpers.tpl` and `configMap.yaml` are a little hacky and if we support other networks in the future like signet there will be a little tech debt to pay off then.

This PR also tests a network composed of all our supported bitcoin core versions and ensures that all tanks parse their `config` properly and respond as expected to RPCs.


willcl-ark (approved):
Looks great!

Also appears not to mind about my mishmash of v and non-v docker hub tags with trim prefix (although I will tidy those up v soon...)

resources/charts/bitcoincore/templates/configmap.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
 data:
   bitcoin.conf: |
     {{- if eq .Values.chain "regtest" }}
+    regtest=1
+
+    {{ template "bitcoincore.check_semver.regtest" . }}
```
josibake, line 12: we could define a _helper function or an entire _bitcoinconf.tpl just for managing all of the backwards compat, deprecated options, etc logic. so then you wouldn't have any of that logic in configmap.yaml, it would just call your helper function. this way we can reuse all of the config creation logic wherever we need it (altho it seems today we only need it in the config map).

resources/charts/bitcoincore/templates/_helpers.tpl
```diff
@@ -55,3 +55,15 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+
+{{/*
+Add network section heading in bitcoin.conf after v0.17.0
+*/}}
+{{- define "bitcoincore.check_semver.regtest" -}}
+{{- $tag := .Values.image.tag | trimPrefix "v" -}}
```
josibake, line 64: we should pick a standard for our tags, either always with v or always without (probs best to follow whatever core is currently doing?), and then re-tag all our images in bitcoindevproject. cc @willcl-ark

josibake (commented):
post merge ACK, left a few comments but all future suggestions. thanks for taking this approach @pinheadmz , looks cleeeeean.

resources/charts/bitcoincore/templates/_helpers.tpl
willcl-ark: Yeah it’s bank holiday here today so no work, but I’ll retag all images without v later/tomorrow

josibake: sounds good, we should also start documenting these conventions _somewhere_, (tho not sure where since this more just for us)

willcl-ark: Yes, I also have custom patches for all the versions! Will stick those in resources too

willcl-ark: Opened #498

PR #497, opened 2024-08-26
add `logs` to warcli

mplsgrant:
The idea is to lift kubectl's `logs` command up into warcli. Since we will have many hard-to-type
pod names, we can use inquirer to check with the user about which pod they would like to query.

Also, I know that we do have the `warcli bitcoin debug-log` command which is similar in how it is implemented, but it is focused on tanks, whereas this `warcli logs` command intuitively works on scenarios as well.

### DIFF
--- a/src/warnet/main.py
+++ b/src/warnet/main.py
@@ -1,3 +1,4 @@
+import json
 import os
 import random
 import subprocess
@@ -6,7 +7,12 @@ from importlib.resources import files
 from pathlib import Path
 
 import click
+import inquirer
 import yaml
+from inquirer.themes import GreenPassion
+
+from warnet.k8s import get_default_namespace
+from warnet.process import run_command, stream_command
 
 from .admin import admin
 from .bitcoin import bitcoin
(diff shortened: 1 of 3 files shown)


willcl-ark:
FYI if you are using `uv` then you could run

```bash
uv lock
```

After adding a new dep to pyproject.toml, which will update the `uv.lock` file. (Or you can use `uv add inquirer` to do it all in one).

If you're not no worries, I can do it after merge.

willcl-ark:
Very smart!

image

Would it be possible to restore support for `--follow` mode as we had here: https://github.com/bitcoin-dev-project/warnet/blob/01195c2fa497cd95b7a9bf41cf87a8bfe8d577c2/src/warnet/cli/network.py#L164-L174

josibake:
Concept ACK, haven't reviewed but I generally agree with this approach of moving things more into the top level of the cli and reusing the same commands for different types of things. As an example, I changed `warcli namespace deploy/warcli network deploy` to `warcli deploy <thing you want to deploy>`. This feels like a similar improvement , where we would need to do something like `warcli bitcoin logs` and `warcli scenario logs` or something, but seems better to just do `warcli logs <thing you want logs for>`

willcl-ark:
We could, instead of having specific helpers, wrap `kubectl exec` with a inquirer picker to a pod. This way user can do, anything in a container (bitcoin-cli, lightning-cli, list files, etc):

```bash
warnet exec
[?] choose a pod
  tank-000
> tank-001
bitcoin-cli -getinfo

```

(but I would still advocate for this log wrapper to remain as a standalone, I actually think calling `bitcoin-cli` rather than `warnet rpc <index> [args]` is easier for most)

mplsgrant:
@willcl-ark Added `--follow`

josibake:
> We could, instead of having specific helpers, wrap kubectl exec with a inquirer picker to a pod. This way user can do, anything in a container (bitcoin-cli, lightning-cli, list files, etc)

I think this is the correct approach long term, but if you're in agreement we would still want a top level logs command then I think this can be merged now?

src/warnet/main.py
```diff
@@ -225,6 +231,37 @@ def auth(kube_config: str) -> None:
         )
 
 
+@cli.command()
+@click.argument("pod_name", type=str, default="")
+@click.option("--follow", "-f", is_flag=True, default=False, help="Follow logs")
+def logs(pod_name: str, follow: bool):
+    """Show the logs of a pod"""
+    namespace = get_default_namespace()
+    pods = run_command(f"kubectl get pods -n {namespace} -o json")
```
josibake: if the user explicitly passes the name of the pod they would like logs for, seems better to directly try and get logs for that pod instead of allows asking the cluster for a list of pods. if we have 1,000 pods, this will make the log command slow every time, not just in the case where the user wants to use the selector menu.

I'd recommend if the user does not provide a pod name, it brings up the selector menu. if a user does provide a pod name, it tries to get the logs and if it fails tells the user the pod doesn't exist and prompts them to use the selector menu.

mplsgrant: @josibake I reworked the workflow based on your advice.

josibake:
code review ACK https://github.com/bitcoin-dev-project/warnet/pull/497/commits/576f67f9dffce2ce882dd8e64139ef63b7eea4b8

haven't tested but changes look good. if someone else gets a chance to test id say this looks RFM

willcl-ark:
> > We could, instead of having specific helpers, wrap kubectl exec with a inquirer picker to a pod. This way user can do, anything in a container (bitcoin-cli, lightning-cli, list files, etc)
> 
> I think this is the correct approach long term, but if you're in agreement we would still want a top level logs command then I think this can be merged now?

Yes I agree, was just pontificating on the wider approach in here

willcl-ark:
tACK 

image

PR #498, opened 2024-08-26
Build CVE images + remove `v*` tags

willcl-ark:


mplsgrant:
These images pull down and work just fine on my machine. They all seem to connect to each other, and miner_std works with the first node as a default 27.

resources/images/bitcoin/insecure/entrypoint_0.16.1.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+# Incorporate additional arguments for bitcoind if BITCOIN_ARGS is set.
+if [ -n "$BITCOIN_ARGS" ]; then
+  IFS=' ' read -ra ARG_ARRAY <<< "$BITCOIN_ARGS"
+  set -- "$@" "${ARG_ARRAY[@]}"
+fi
+
+echo
+exec "$@" -printtoconsol
```
mplsgrant: Is this supposed to be `-printtoconsole`?

Also, I don't think this file is used anywhere (yet?).

willcl-ark: Yes I believe both points need addressing, thanks!

@pinheadmz you said that you needed some special sauce to get 0.16.1 working, no? Or was it only the (now-fixed) config file stuff that needed fixing?

josibake: IIRC, this was fixed via his "old config" change? the main point is I don't think he needed to change anything about the docker image itself so I don't think this is a blocker for this PR

willcl-ark: Sweet

resources/images/bitcoin/insecure/addrman_v0.21.1.patch
```diff
@@ -0,0 +1,52 @@
...
++    AddrInfo info2 = AddrInfo(addr2, source);
++    AddrInfo info3 = AddrInfo(addr3, source);
++    AddrInfo info4 = AddrInfo(addr4, source);
++
++    // All addrs are in different buckets
++    BOOST_CHECK_EQUAL(info1.GetNewBucket(nKey1, EMPTY_NETGROUPMAN), 533);
++    BOOST_CHECK_EQUAL(info2.GetNewBucket(nKey1, EMPTY_NETGROUPMAN), 20);
++    BOOST_CHECK_EQUAL(info3.GetNewBucket(nKey1, EMPTY_NETGROUPMAN), 143);
++    BOOST_CHECK_EQUAL(info4.GetNewBucket(nKey1, EMPTY_NETGROUPMAN), 291);
++}
```
josibake, line 48: do we need this in the patch? feels like noise to me

resources/images/bitcoin/insecure/Dockerfile_v0.16.1
```diff
@@ -0,0 +1,113 @@
+# Setup deps stage
+FROM alpine:3.7 AS deps
+ENV REPO="bitcoin/bitcoin"
+ENV COMMIT_SHA="dac5d68fc6cf136e0d7b21b9ed4fa053d54e6059"
+ENV BUILD_ARGS="--disable-tests --without-gui --disable-bench --disable-fuzz-binary --enable-suppress-external-warnings"
```
josibake, line 5: unclear to me where we are tracking our "defaults." seems fragile to be repeating them all in each image, and also now there is a discrepancy between these images and the fixes I did for https://github.com/bitcoin-dev-project/warnet/issues/441

if you have any ideas on how to make "one source of truth" for our docker configs that can then be overridden by the user, it would be nice to handle that, either here or in a separate PR.

however, at the very least I think you should include the changes for disabling dependency tracking and making gcc less resource hungry in this PR.

josibake (commented):
perhaps there's a reason im not seeing, but it feels like a lot of these images are super similar and the differences between them aren't apparent/well documented. this feels like this is gonna cause trouble in the future when we look back at these images and have no idea what each one is doing.

sadly, I don't have a silver bullet to recommend. id be okay with moving forward with this approach so long as we have an issue to refactor this (or better documentation as to why it needs to be this way).

josibake:
ping @willcl-ark , would be nice to at least include the resource limits fixes from #441

resources/images/bitcoin/insecure/Dockerfile_v0.16.1
willcl-ark: We aren't tracking defaults as far as I know. Different builds need different things to get working in roughly the same way

willcl-ark:
> perhaps there's a reason im not seeing, but it feels like a lot of these images are super similar and the differences between them aren't apparent/well documented. this feels like this is gonna cause trouble in the future when we look back at these images and have no idea what each one is doing.
> 
> sadly, I don't have a silver bullet to recommend. id be okay with moving forward with this approach so long as we have an issue to refactor this (or better documentation as to why it needs to be this way).

Yeah the images are similar, but they are simply hardcoded working builds of old images. IMO would be a waste of time to try and extract their commonality into some base image, then get them all working from that, but I would review that PR if you want to open it!

> and have no idea what each one is doing

Which bits do you think someone would have no idea what they are doing? Each version has a dockerfile and pair of patches you build with. I also provided a command for each. For what the Dockerfile is doing, see inside the Dockerfile :)

willcl-ark:
> ping @willcl-ark , would be nice to at least include the resource limits fixes from #441

I'm not sure I will get around to rebuilding all of these. I also don't see why users would need to? They can just use our images, no?

josibake:
> I also don't see why users would need to

We are users in this sense, and if any of use need to rebuild these images for whatever reason in the future, I'd prefer they have those fixes, rather than hitting an obtuse Docker error where the fix is messing with the default resources Docker desktop starts with. Also, I don't see why you would need to rebuild all of the images just for a docker file change. Seems sufficient to add it and test that one of them works?

josibake:
> but I would review that PR if you want to open it!

Sure, as I mentioned this can be a follow-up after Tabconf. I've opened https://github.com/bitcoin-dev-project/warnet/issues/508 as a backlog item.

willcl-ark:
Added in 469f9786651302d98a23da115eb527e01d8b31c1 but this PR is now untested by me

resources/images/bitcoin/insecure/addrman_v0.21.1.patch
josibake: ping @willcl-ark

josibake:
Would prefer if we were not committing a test in the patch per https://github.com/bitcoin-dev-project/warnet/pull/498#discussion_r1732434981. Otherwise, looks RFM.

resources/images/bitcoin/insecure/addrman_v0.21.1.patch
pinheadmz: the test was added in #376 because the original addrman patch wasn't correct. i wanted a test to ensure the new patch worked. I guess we can take it out, but it is a handy way to test that our patch does what we think it does

resources/images/bitcoin/insecure/entrypoint_0.16.1.sh
pinheadmz: yes fixed in #494 but also if this "v" thing is fixed you can remove this line:

https://github.com/bitcoin-dev-project/warnet/blob/5fbcd63ebbc5b02ace4a4aa1730a74bf3efad768/resources/charts/bitcoincore/templates/_helpers.tpl#L64

pinheadmz:
`test/conf_test.py` deploys `test/data/bitcoin_conf` which is a test that (should) run all the images we support. I think you should remove the v's from the network.yaml for that test. Ideally there'd also be some additional way to ensure that our latest images are being used by that test, maybe even by image hash?

willcl-ark:
> `test/conf_test.py` deploys `test/data/bitcoin_conf` which is a test that (should) run all the images we support. I think you should remove the v's from the network.yaml for that test. Ideally there'd also be some additional way to ensure that our latest images are being used by that test, maybe even by image hash?

Yeah, I also plan to remove teh tags from hub.docker.com, which would cause us to have noticed this!

Edit: these are already removed in this PR?

resources/images/bitcoin/insecure/entrypoint_0.16.1.sh
willcl-ark: Done

resources/images/bitcoin/insecure/addrman_v0.21.1.patch
josibake: I don't see the value of having this committed to our repo considering:

1. The test is not being run in CI (afaict)
2. The code for 0.21.1 is not going to change
3. We aren't testing any of the other images

So this ends up just being dead code for this patch in particular. Maybe I'm missing something here, but it feels like we are committing a one-time debug script along with the patch. If we really think this is needed, let's create tests for the rest of the patches and run them in CI? Even better, lets have tests in warnet CI that test these images behave as expected, without relying on test code in the patch itself.

willcl-ark: I’ll happily review that PR

pinheadmz: > The code for 0.21.1 is not going to change

Oh yeah i didnt realize this was just applied to one version. It's ok with me if the unit test is only applied and run in CI. I just want the test somewhere in the repo so we have confidence our regtest addr gossip works

PR #499, opened 2024-08-26
rename warcli cmd to warnet

willcl-ark:
needs a package refresh to test

```bash
pip uninstall warnet
pip install -e .
```


josibake:
ACK from me for the following reasons:

* installing a tool and then needing to use a different command to access the tool feels weird
* users can alias the command to whatever they prefer, so feels like we should go with the simplest option here
* anecdotally , ive typed `warnet` on the CLI multiple times as a reflex and observed other users doing the same. this indicates to me its a more intuitive name
* @willcl-ark mentioned this also makes python packaging easier, I haven't verified but ill take his word for it

would be great to get @pinheadmz 's opinion before merging

josibake:
ping @pinheadmz , iirc you had an alternative suggestion for this PR?

pinheadmz:
post merge ACK 🍻

PR #501, opened 2024-08-27
fix `warcli down` namespace

mplsgrant:
### Issue
`warcli down` will throw an error such as

```` bash
$ helm uninstall tank-0000 
Error: uninstall: Release not loaded: tank-0000: release: not found 
````

### Solution
Adding the default namespace to the command seems to fix this.

### DIFF
--- a/src/warnet/control.py
+++ b/src/warnet/control.py
@@ -123,7 +123,7 @@ def down():
     tanks = get_mission("tank")
     with console.status("[yellow]Uninstalling tanks...[/yellow]"):
         for tank in tanks:
-            cmd = f"helm uninstall {tank.metadata.name}"
+            cmd = f"helm uninstall {tank.metadata.name} --namespace {get_default_namespace()}"
             if stream_command(cmd):
                 console.print(f"[green]Uninstalled tank: {tank.metadata.name}[/green]")
             else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/warnet/control.py [hunk 1]
```diff
@@ -123,7 +123,7 @@ def down():
     tanks = get_mission("tank")
     with console.status("[yellow]Uninstalling tanks...[/yellow]"):
         for tank in tanks:
-            cmd = f"helm uninstall {tank.metadata.name}"
+            cmd = f"helm uninstall {tank.metadata.name} --namespace {get_default_namespace()}"
```
josibake, line 126: I think this is fine for now, but leaving a comment for future self that we will need to rethink this if we are deploying commanders into a separate namespace.

josibake:
ACK https://github.com/bitcoin-dev-project/warnet/pull/501/commits/900920e0f0b4a46e599a7d877cd9e12da88491d4

src/warnet/control.py
mplsgrant: @josibake I'm going to merge then in order to get `warcli down` working.

As for commanders, perhaps we could concat get_default_namespace with '-commander', and place them in that namespace.

PR #506, opened 2024-09-02
Run any executable file on a commander

### LINKED ISSUE #272: Make Commanders run any executable?

willcl-ark:
Closes: #272

- Run any binary in a commander.
- Python files and other script-based languages will need a correct shebang to run.
- Other binaries should work natively.


josibake:
Needed for Tabconf?

willcl-ark:
Not if we want to force everyone to write attacks in only python.

Otherwise, I'd say yes?

src/warnet/control.py
```diff
@@ -214,25 +217,24 @@ def run(scenario_file: str, additional_args: tuple[str]):
                 "containers": [
                     {
                         "name": name,
-                        "image": "bitcoindevproject/warnet-commander:latest",
-                        "args": additional_args,
+                        "image": "bitcoindevproject/warnet-commander:dev",
```
willcl-ark: Before merging this line needs to be reverted (and a new image pushed)

pinheadmz:
could you give an example on how this would work? or *add a test* ?

willcl-ark:
Sure:

```bash
echo "#!/usr/bin/env sh
echo 'Hello, World!'" > echo.sh

chmod +x echo.sh

warcli run echo.sh

# observe logs for whatever the command podname happened to be on that run
```

Looks like:

image

I could consider a test I suppose.

josibake:
> I could consider a test I suppose.

Generally speaking, I think we should start requiring tests on PRs for any new user facing feature. The tests also serve as rudimentary documentation. If you can add a test or create a follow-up issue for adding one, I think we can merge this.

src/warnet/control.py
pinheadmz: this kind of thing is so awkward. i know its a LOT but i feel like ideally CI would build images and use them locally, then on merge to main the image gets pushed to dockerhub. But that means that in CI we'd want to "never" pull (use local image) and in prod we'd want to "always" pull (?) to make sure the latest is used?

willcl-ark:
Test added @josibake @pinheadmz 

Don't merge before letting me update the commander image on dockerhub (and reverting the image from dev to latest)

willcl-ark:
Damn, anyone got any clues why CI fails? Test passes for me locally...

I don't know what https://github.com/bitcoin-dev-project/warnet/actions/runs/10689084760/job/29630400523?pr=506#step:7:119 means particularly

src/warnet/control.py
josibake: +1, I think our CI/image pipeline could use some love. Don't think it should be a blocker here and not needed before Tabconf, so I created a backlog to track https://github.com/bitcoin-dev-project/warnet/issues/520

josibake:
> Damn, anyone got any clues why CI fails? Test passes for me locally...

Seems related to inquirer and CI running in a headless environment. Not sure why inquirer is being called in this PR tho?

willcl-ark:
Rebased now #523 is in

.github/workflows/test.yml
```diff
@@ -35,15 +35,20 @@ jobs:
...
           context: .
           tags: bitcoindevproject/warnet-commander:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          outputs: type=docker,dest=/tmp/commander.tar
+          load: true
+
+      - name: Export docker image with tags
+        run: |
+          # This workaround should not be needed...
```
willcl-ark, line 50: This indicates that whilst #523 was working, it was not working as intended.

With this change we now (for sure!) use the CI-built docker image in the CI run...

256dummy:


### Conflicts

* [#524](https://github.com/bitcoin-dev-project/warnet/pull/524) (Add create-network cli command by m3dwards)

.github/workflows/test.yml
josibake: a bit confused as to what's going on here. is #523 good or do we need a fix? can we drop the work around if everything is good?

resources/scenarios/commander.py
```diff
@@ -135,9 +135,13 @@ def parse_args(self):
...
-        if help_args.help:
-            help_parser.print_help()
-            sys.exit(0)
+        try:
+            # This fails inside of CI, and I don't know why particularly
+            if help_args.help:
+                help_parser.print_help()
+                sys.exit(0)
+        except Exception:
+            pass
```
josibake: this seems like a hack, would be better if we can figure out why CI is failing. happy to help troubleshoot this.

src/warnet/control.py
```diff
@@ -157,17 +157,21 @@ def get_active_network(namespace):
...
 
-    with open(scenario_path) as file:
-        scenario_text = file.read()
+    import base64
 
-    name = f"commander-{scenario_name.replace('_', '')}-{int(time.time())}"
+    with open(file_path, "rb") as f:
+        b_64_file = base64.b64encode(f.read())
+
+    pull_policy = "Never" if os.getenv("GITHUB_ACTIONS") else "IfNotPresent"
```
josibake, line 195: why can't you set `pullPolicy: Never` on the node-defaults.yaml for the test?

josibake: ah, I see, its because the commander is not deployed via helm 😢 (and node-defaults refers to just tanks, anyways).

I think we need to fix the commander first and have it be deployed via a helm chart, or at least move the k8s yaml out of the python code and decided how we want to configure it. Then, we can rebase this on that PR? 

I don't think having our python code be aware it's running in CI and then changing behaviour based on that is a good idea. Let me know if you have any interest in moving the commander to a helm chart, or I can take a crack at it once I've finished the signet stuff.

willcl-ark: Attempted in #550 👍🏼

resources/scenarios/commander.py
willcl-ark: This was fixed in https://github.com/bitcoin-dev-project/warnet/pull/552 after the failure was diagnosed fully.

I will rebase this shortly

josibake:
If this is rebased on #550 and #552 , I think it's g2g? would be nice to get this in

willcl-ark:
> If this is rebased on #550 and #552 , I think it's g2g? would be nice to get this in

Rebased. A good one to manually test I think

m3dwards:
Followed instructions and got:

```shell
  Normal   Scheduled  45s   default-scheduler  Successfully assigned default/commander-echo-1726148810 to docker-desktop
  Normal   Pulled     45s   kubelet            Container image "bitcoindevproject/warnet-commander:latest" already present on machine
  Normal   Created    45s   kubelet            Created container commander
  Warning  Failed     45s   kubelet            Error: failed to start container "commander": Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/entrypoint.sh": stat /entrypoint.sh: no such file or directory: unknown
```

Any ideas? I have run `docker pull bitcoindevproject/warnet-commander:latest`.

josibake:
> Any ideas? I have run docker pull bitcoindevproject/warnet-commander:latest.

@m3dwards I ran into a very similar error when working on the CVE images and it was due to bash not being installed in the container

resources/images/commander/Dockerfile
```diff
@@ -4,8 +4,12 @@ FROM python:3.12-slim
...
+RUN mkdir /wbin
+
 COPY resources/scenarios/commander.py /
 COPY src/test_framework /test_framework
+COPY resources/images/commander/entrypoint.sh /
+RUN chmod +x /entrypoint.sh
 
-# -u: force the stdout and stderr streams to be unbuffered
-ENTRYPOINT ["python", "-u", "/scenario.py"]
+ENTRYPOINT ["/bin/sh", "/entrypoint.sh"]
```
josibake, line 15: says /bin/sh here, but `#!/bin/bash ` in the entrypoint script? not sure , but this could be causing the error @m3dwards is seeing

m3dwards:
Am I right in thinking to test this I need to build a new commander image? That would make sense.

Yeah, that was it:

```shell
Executing command with args:
/command
Hello, World!
+ echo 'Command execution completed'
Command execution completed
```

pinheadmz:
We can discuss this in the chat or on the call today but I'm feeling "approach nack" on this. The commander image is streamlined for scenarios written in python. If users want to deploy something else, why do we need to force that into the commander? Can't we just deploy a plain alpine container that runs whatever the user wants?

.github/workflows/test.yml
m3dwards: I'm also intrigued

src/warnet/control.py
m3dwards: Now that 550 is in can this be revisited?

resources/charts/commander/templates/pod.yaml
```diff
@@ -12,21 +12,19 @@ spec:
     - name: {{ .Chart.Name }}
       image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
       imagePullPolicy: {{ .Values.image.pullPolicy }}
-      command: ["/bin/sh", "-c"]
+      command: ["/entrypoint.sh"]
```
m3dwards, line 15: Command is already specified in the dockerfile as `ENTRYPOINT`. Is this line required?

willcl-ark:
> Am I right in thinking to test this I need to build a new commander image? That would make sense.
> 
> Yeah, that was it:
> 
> ```shell
> Executing command with args:
> /command
> Hello, World!
> + echo 'Command execution completed'
> Command execution completed
> ```

Sorry, should have left instructions on local testing. CI does test it properly though as it builds the image and loads it

josibake:
Closing, I think the idea is good but we might want to implement it in a different way.

PR #507, opened 2024-09-02
add forkobserver

willcl-ark:


willcl-ark:
~~Need to complete the `templates/` stuff~~

resources/charts/fork-observer/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+Thank you for installing {{ include "bitcoincore.fullname" . }}.
```
josibake: this might fail? should reference `fork-observer.fullname` iirc

josibake:
Concept ACK, I think you can delete a lot of the stuff copied over from the bitcoincore charts, e.g., we don't need a prom monitor for fork-observer.

resources/charts/fork-observer/templates/NOTES.txt
willcl-ark: Yeah I didn't do the templates bit. Might ask @m3dwards very nicely for help here when he's back around 👅

resources/charts/fork-observer/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+Fork-observer is watching you
```
m3dwards, line 1: 😃 👀

resources/charts/fork-observer/values.yaml
```diff
@@ -0,0 +1,133 @@
...
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: chart-example.local
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
```
m3dwards: This probably also applies to the bitcoin-core chart but being as there isn't any ingress at the moment, probably shouldn't have anything referring to that in the values.yaml

m3dwards:
Should there be a "connect forkobserver" cli command to access the GUI?

resources/charts/fork-observer/values.yaml
```diff
@@ -0,0 +1,133 @@
...
+        </div>
+      </div>
+    """
+
+  [[networks]]
+  id = 0xDEADBE
+  name = "Warnet"
+  description = "A Warnet"
+  min_fork_height = 0
+  max_interesting_heights = 5
```
0xB10C: ~you probably don't need and want to set this~ It will only show you the most recent 5 forks. You might want to set it to 100 for now

resources/charts/fork-observer/values.yaml
```diff
@@ -0,0 +1,133 @@
...
+  # fork-observer base configuration file
+
+  # Database path of the key value store. Will be created if non-existing.
+  database_path = "db"
+
+  # path to the location of the static www files
+  www_path = "./www"
+
+  # Interval for checking for new blocks
+  query_interval = 10
```
0xB10C: How many nodes do you plan to run? Checking 100 nodes every 10 seconds might be a bit much, but checking 10 nodes every 10 seconds should be fine.

m3dwards:
Could do with some documentation?

resources/charts/bitcoincore/values.yaml
```diff
@@ -118,6 +118,9 @@ regtestConfig: |
   rpcport=18443
   rpcallowip=0.0.0.0/0
   rpcbind=0.0.0.0
+  rpcauth=forkobserver:ef3d61b7ffecd81ffaaa7ae43091543d$a10bc25b5ca8910ebbee9b1538d38242b09d0a165e220f9a47535eccc4c089a4
```
josibake: Probably don't want to set this as a default because I don't think we will be able to override it with a user supplied config, which means we will need a code change every time we want to change the password (ew). I'd recommend moving this rpcauth code black to the `node-defaults.yaml` file. node-defaults is always applied, and can be easily overridden with a user supplied config.

src/warnet/deploy.py
```diff
@@ -53,13 +54,60 @@ def deploy(directory):
...
+        # deploy_fork_observer(directory)
     elif (directory / NAMESPACES_FILE).exists():
         deploy_namespaces(directory)
     else:
         click.echo(
             "Error: Neither network.yaml nor namespaces.yaml found in the specified directory."
         )
 
+def deploy_fork_observer(directory: Path):
+    network_file_path = directory / NETWORK_FILE
```
josibake, line 68: note to future self: this works perfectly fine for a "network per namespace" deployment (i.e., how we would expect people to use warnet in a non-tabconf/wargames setting). however, for tabconf, we will have a network file per team that will deployed into each teams namespace but will want fork observer to be watching all nodes from all namespaces.

as a manual-ish hack, its super easy to just concatenate all of the network files and then use that for fork observers config, but we might also want a fork-observer specific config.yaml file, which would essentially just be a list of nodes with their names/hostnames. 

long-term, I think we will want some python code for managing all of these individual yaml configs to make sure they all line up correctly.

src/warnet/deploy.py
```diff
@@ -53,13 +54,60 @@ def deploy(directory):
...
+        node_name = node.get("name")
+        node_config = f"""
+[[networks.nodes]]
+id = {i}
+name = "{node_name}"
+description = "A node. Just A node."
+rpc_host = "{node_name}"
+rpc_port = 18443
+rpc_user = "forkobserver"
+rpc_password = "tabconf2024"
```
josibake, line 96: note to future self: we will want to move these hardcoded values out of the python code. today, if you update the tank configs to use a different port and forget to update the port value here, fork observer is borked.

even worse, the user can change the port via the yaml config (no code change required), but then to update fork observer would require a PR!

josibake (approved):
ACK, left some comments on parts that will need improvement going forward, but probably fine to merge for now (unless you're feeling motivated to come up with a fix in this PR 😄 ).

I'll also move some of this code out of network.py in a clean up PR. Ideally, I'd like to have a more general "deploy.py" that has all of the code for deploying helm charts.

src/warnet/deploy.py
josibake: created https://github.com/bitcoin-dev-project/warnet/issues/511 to track discussion on this

willcl-ark:
Thanks for review all, will try and address now :)

resources/charts/fork-observer/values.yaml
josibake: I'm thinking we will want to move these values out into a config.yaml for fork-observer (same as we do for networks/namespaces) and allow these values to be set by the user. For some use cases, someone might only want to see max 5, but for the Tabconf use case we will want more.

resources/charts/fork-observer/values.yaml
josibake: Same comment as https://github.com/bitcoin-dev-project/warnet/pull/507#discussion_r1741882126, this should be configurable by the user. It would also allow us to easily redeploy fork-observer with different intervals if we find its starting to slow down for a particular network.

resources/charts/fork-observer/values.yaml
willcl-ark: Done

resources/charts/fork-observer/values.yaml
willcl-ark: > you probably don't need and want to set this. It will only show you the most recent 5 forks

Done

>  I'm thinking we will want to move these values out into a config.yaml for fork-observer (same as we do for networks/namespaces) and allow these values to be set by the user. For some use cases, someone might only want to see max 5, but for the Tabconf use case we will want more.

I have done this halfway in that it's now its own config value, but I don't pass any user-suppied values into it yet.

resources/charts/bitcoincore/values.yaml
willcl-ark: Done in new generic `node-defaults.yaml`

resources/charts/fork-observer/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "fork-observer.fullname" . }}
+  labels:
+    {{- include "fork-observer.labels" . | nindent 4 }}
+data:
+  config.toml: |
+    {{- .Values.configQueryInterval | nindent 4 }}
```
josibake, line 9: very nice, I like this pattern: user configurable items that MUST be present get their own key in values.yaml (which allows us to set a default), baseConfig is what we need regardless of user input can be defines as a text string, and config can then be a catch all for everything else , where the user needs to pass in the actual string to concatenate to the config file. all of the logic for creating the final config file that will eventually end up in a configmap can live in the template function, and we can use helms semver features to make sure we always construct a correctly formatted config for whatever version of the thing we are deploying.

this should work regardless of the config file type, e.g. bitcoin.conf , config.toml, etc.

willcl-ark: Just hooking this up to the `quickstart` command, as why not...

josibake: Nice! might also be worth adding `max_interesting_heights` as a key values.yaml, as well?

willcl-ark: Could do ... could do ... 😭

josibake: Haha either we add it here (and then this looks RFM), or we can merge now and you can open a follow-up issue to fix the hardcoded value?

pinheadmz:
Code looks OK and branch runs but I dunno how to see the actual website (docs? port-forward command?)

Also would be nice to have test coverage for new features. I had a fork observer test here: https://github.com/bitcoin-dev-project/warnet/pull/333/commits/4552290c8a5bb40ddbfa0d43196a59bc10e069fa

ok figured out ` k port-forward fork-observer 2323`

pinheadmz:
I'm getting `RPC unreachable` errors in the web UI and this in the F-O log:

```
[2024-09-03T17:04:12Z ERROR fork_observer] Could not fetch chaintips from Node(id=0, name='tank-0001', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=0, name='tank-0001', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=4, name='tank-0005', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=3, name='tank-0004', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=5, name='tank-0006', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=2, name='tank-0003', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
[2024-09-03T17:04:32Z ERROR fork_observer] Could not fetch chaintips from Node(id=1, name='tank-0002', implementation='Bitcoin Core') on network 'Warnet' (id=14593470): BitcoinCoreRPC(JsonRpc(Transport(HttpErrorCode(401))))
```

Individual tank logs are reporting incorrect RPC password

m3dwards:
> I'm getting `RPC unreachable` errors in the web UI and this in the F-O log

me too

m3dwards:
Looks good to me

Screenshot 2024-09-03 at 21 38 20

m3dwards (approved):

src/warnet/deploy.py
pinheadmz: better yet, since we parse the `network.yaml` first in `deploy_network()` we can:

1. dynamically add `rpcauth=forkobserver` to each tank only if needed (and with unique passwords too)
2. get the actual port from each node configuration to match in the FO config.
3. `description` could be more dynamic as well (tank name? version? uacomment?)

pinheadmz:
Will is asleep push tests to his branch!

josibake:
https://github.com/bitcoin-dev-project/warnet/pull/507#discussion_r1742354256

resources/charts/fork-observer/templates/configmap.yaml
willcl-ark: I am not going to work on this so leave for a followup I guess

m3dwards: 884d94b3845d7a73d22dc4cc9106af8022900976

m3dwards:
Probably isn't related to PR but when testing I can't run `miner_std`.

```shell
test_framework.authproxy.JSONRPCException: non-JSON HTTP response with '403 Forbidden' from server
```

Update: Seems to be working on main so looking like it could be this PR

m3dwards:
reACK

pinheadmz:
ACK for now but I thought the "hard coded config" complaint was more about the RPC ports and u/pw: https://github.com/bitcoin-dev-project/warnet/pull/507#discussion_r1742832168

m3dwards:
> ACK for now but I thought the "hard coded config" complaint was more about the RPC ports and u/pw: [#507 (comment)](https://github.com/bitcoin-dev-project/warnet/pull/507#discussion_r1742832168)

Let's get this in as it's going to conflict with other PRs, we have #511 to track inter-chart configs.

PR #509, opened 2024-09-02
Update quickstart: minikube and docker-desktop

mplsgrant:
This updates quickstart to prompt the user about their choice of minikube vs docker-desktop.

One open question is about minikube on macos. Is there actually an issue with version 1.33.1? Or is that the only good version on mac? If it's the only good version, then I'll need to change the check to ensure users have that version installed.


pinheadmz:
> One open question is about minikube on macos. Is there actually an issue with version 1.33.1? Or is that the only good version on mac? If it's the only good version, then I'll need to change the check to ensure users have that version installed.

Last I checked the issue was with version 1.32 and 1.33

see https://github.com/kubernetes/minikube/issues/19173#issuecomment-2209193041

pinheadmz:
Not happy on macos even though i have docker desktop and run warnet all the time!

```
--> ./resources/scripts/quick_start.sh 

   ╭───────────────────────────────────╮
   │  Welcome to the Warnet Quickstart │
   ╰───────────────────────────────────╯

    Let's find out if your system has what it takes to run Warnet...

 ⭐️ Found docker: /usr/local/bin/docker
 ⭐️ Running Docker on Darwin
 💥 Could not find minikube version > 1.33.1. Please upgrade...
   https://minikube.sigs.k8s.io/docs/start/
 ⭐️ Found kubectl: /usr/local/bin/kubectl 
 ⭐️ Found helm: /opt/homebrew/bin/helm
 ⭐️ Found just: /opt/homebrew/bin/just 
 ⭐️ Found python3: /opt/homebrew/bin/python3 
 ⭐️ Found venv: a python3 module
There were errors in the setup process. Please fix them and try again.
```

pinheadmz:
Also, given that there is now `warnet quickstart` we need to resolve the name collision here, or split up the use case. My preference is to do more in python and less bash scripts

mplsgrant:
> My preference is to do more in python and less bash scripts

Agreed. This seems like it should be in python.

> Last I checked the issue was with version 1.32 and 1.33

Thank you. I will update to reflect that. Should solve the issue you were having where it complains about 1.33.1.

pinheadmz:
Nice, tested some of the paths but not all. Happy to see another bash script get incorporated into the actual code.

Only feedback is that this seems to do two things now:
1. check dependencies
2. init a warnet project

Selecting "minikube" in step 1 doesn't necessarily affect how warnet is run, just what dependencies are checked.

Step 1 is something a user only needs to do once, after installing warnet. Step 2 is something they can do over and over again without needing to do step 1 again.

So I wonder if we can split this into another command like `warnet setup` and THEN user calls `quickstart`.

Honestly Id be in favor of renaming `quickstart` as well or maybe combining it with `init`.

setup, quickstart, init, create -- help me I'm a new user and I have no idea which command i need

mplsgrant:
> So I wonder if we can split this into another command like `warnet setup` and THEN user calls `quickstart`.

In my mind, `quickstart` checks the user's system for prerequisites, and then if everything checks out, it invites the user to run `warnet create`. This way, quickstart has all the information about what a user needs to run warnet, and it guides the user to getting their system ready. The dependencies are documented as code.

> Honestly Id be in favor of renaming `quickstart` as well or maybe combining it with `init`.

I think merging `quickstart` into `init` and `create` is a good idea. It would mean that every time we run `warnet init` it would check the user's system for dependencies; however, this could be minimized by examining the current `docker` and `kubectl` configs. @m3dwards suggested something along that line at one point.

pinheadmz:
ACK nice

willcl-ark:
ACK from me too, great work!

> > So I wonder if we can split this into another command like `warnet setup` and THEN user calls `quickstart`.
> 
> In my mind, `quickstart` checks the user's system for prerequisites, and then if everything checks out, it invites the user to run `warnet create`. This way, quickstart has all the information about what a user needs to run warnet, and it guides the user to getting their system ready. The dependencies are documented as code.
> 
> > Honestly Id be in favor of renaming `quickstart` as well or maybe combining it with `init`.
> 
> I think merging `quickstart` into `init` and `create` is a good idea. It would mean that every time we run `warnet init` it would check the user's system for dependencies; however, this could be minimized by examining the current `docker` and `kubectl` configs. @m3dwards suggested something along that line at one point.

Yeah I really like this idea. Took a stab at it (building off this PR) here: https://github.com/willcl-ark/warnet/tree/command-split

...and I really feel like it operates much more intuitively with:

```bash
warnet setup
warnet create <dir_name>    # init in that dir
warnet init                 # init in this dir
```


Is this ready for merge then?

256dummy:


### Conflicts

* [#524](https://github.com/bitcoin-dev-project/warnet/pull/524) (Add create-network cli command by m3dwards)

mplsgrant:
Thank you! Seems like we have a few acks. Good to move forward on this.

PR #517, opened 2024-09-03
commander: remove lie

pinheadmz:
follow up!


willcl-ark:
You missed that this dockerfile doesn't work 😝 

Fixed in https://github.com/bitcoin-dev-project/warnet/pull/506 though ;)

PR #521, opened 2024-09-04
Remove dead code

josibake:
Post Italy refactors, mostly removing dead code.


willcl-ark:
Nice looking start. IMO it would be lush if `main.py` ended up looking (roughly) like this (I'm happy to tackle myself in a followup to this dedup/reorg work though):

image

josibake:
> IMO it would be lush if main.py

yeah, main has turned back into a dumping ground. What do you think about something like:

- main is a light wrapper which imports all of the other commands , and perhaps has any styling / display functions
- commands get their own .py file, e.g., deploy.py (tiny commands can be grouped into a logical file)
- utilities and warnet functions not exposed in the cli go into their own folders

this is what im working towards here, but will probably not do all of it in this PR.

josibake:
Have some other ideas on code cleanups in another branch, but going to stop here on this PR to keep it simple. As of right now, this just deletes code that was not being used and makes a few modifications in the tests to have them use slightly different functions for checking the predicates.

m3dwards:
Should there be a cli command called `connected` introduced with this? I can't immediately see it.

josibake:
> Should there be a cli command called `connected` introduced with this? I can't immediately see it.

good shout! I was experimenting with adding it in for @pinheadmz use case, and forgot to remove it. will fix in a sec

m3dwards:
Span up a network and ran some scenarios. LGTM

m3dwards (approved):

PR #522, opened 2024-09-04
scenarios: only print 'outer' help

### LINKED ISSUE #512: pass through scenario file help options
If I run `warnet run ./scenarios/mining_std.py` , it's difficult to know what options are available for the python test. We could do something like `warnet run ./scenarios/mining_std.py --help`, which would then show the arguments for that script, but this somewhat assumes all of the scripts have some sort of help option with them, which definitely won't be the case once we enable running _anything_ via `warnet run <path to executable>`.

Perhaps simpler would be let the user determine the arguments for the script without needing to go through warnet, something like `./scenarios/mining_std.py --help`. This doesn't work today because the scripts aren't executable but we can change that so that when the directory is created for the user all of the default scenarios are created as executable.

Another issue with this approach is the bitcoin python test framework class spits out a bunch of irrelevant options, e.g.:

image

So at a minimum id propose we:

- modify the WarnetTestFramework / commander code to not show help options from the bitcoin test framework
- make it so all scripts in scenarios/ are executable by default

Curious to hear other suggestions before jumping in with a PR for this?

willcl-ark:
Fixes: #512

Use a second argparser to only print "outer" help options.

This makes it clearer for users:

image

### DIFF
--- a/resources/scenarios/commander.py
+++ b/resources/scenarios/commander.py
@@ -131,6 +131,14 @@ class Commander(BitcoinTestFramework):
         self.success = TestStatus.PASSED
 
     def parse_args(self):
+        # Only print "outer" args from parent class when using --help
+        help_parser = argparse.ArgumentParser(usage="%(prog)s [options]")
+        self.add_options(help_parser)
+        help_args, _ = help_parser.parse_known_args()
+        if help_args.help:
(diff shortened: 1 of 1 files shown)


josibake:
Awesome! 

Closes #512

PR #523, opened 2024-09-04
use built commander image in CI

### LINKED ISSUE #520: CI image pipeline improvements
We should build images in CI and use those directly. Ideally, once a PR gets merged to main, it should also build and push images to dockerhup (post-merge action?).

I'm sure this is a well established patterns in other projects, so likely something we can pull from other places. The goal here is to minimise differences between what we are testing vs what's in our public docker hub and get our PR / review / merge process to not involve manual steps, like changing image names from dev -> latest before merging.

Adding as a backlog, not high priority.

willcl-ark:
Fixes: #520

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,34 @@ jobs:
       - uses: eifinger/setup-uv@v1
       - run: uvx ruff format . --check
 
+  build-image:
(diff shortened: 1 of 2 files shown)


josibake:
Can merge this now and perhaps you can rebase your command PR on top of this so we don't need to switch the tag from dev to latest?

Issue #527, opened 2024-09-04
"warnet down" takes a long time for large network


pinheadmz:
I think its because we are "helm uninstall"-ing one tank at a time in a loop. I know it won't work with "default" namespace but for sure, `namespace delete` worked a LOT faster

willcl-ark:
Is it not fork-observer❓

josibake:
Might be possible to tune our helm commands a bit, but also doing a namespace delete should work. One thing to keep in mind: if we deploy stuff with helm and then delete it with something else, we might leave helm in a confused / borked state.

pinheadmz:
we learned in italy that we cant delete the `default` namespace

willcl-ark:
Helm has state? 

I thought it was a yaml masher only?

willcl-ark:
I think I have something for this, inadvertently discovered when trying to teardown my Caddy deployment...

josibake:
It would be great to speed up `warnet down` but I don't think this is a priority before Tabconf, unless we are talking about wait times that make development and testing unbearable.

I'd also suggest we investigate tuning helm first before going straight to managing the teardown ourselves in python.

willcl-ark:
https://github.com/BarelElbaz/helm-delete-all

If we really want this to be faster, this plugin may be a solution

pinheadmz:
closed by #550

PR #530, opened 2024-09-04
faster and cleaner teardown

willcl-ark:
- Add more k8s helpers
- Use k8s API for more things

### DIFF
--- a/src/warnet/control.py
+++ b/src/warnet/control.py
@@ -12,10 +12,10 @@ from rich.table import Table
 
 from .k8s import (
     apply_kubernetes_yaml,
+    delete_all_resources,
     delete_namespace,
     get_default_namespace,
     get_mission,
-    get_pods,
 )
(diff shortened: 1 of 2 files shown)


josibake:
Not opposed to this if it is much faster (and we've confirmed that speeding up helm is not possible), but as I mentioned in #527 I'm worried about leaving things in an inconsistent state between what helm thinks is deployed/not deployed vs what the cluster thinks. It also feels like a bit of a regression since one of the motivations of adding helm was to have _less_ python code talking directly to the K8s API. This seems to be reverting back to our old approach.

Perhaps we look at tuning helm first and see if there's a way to make it faster?

willcl-ark:
Some faster/cleaner teardown is achieved as part of https://github.com/bitcoin-dev-project/warnet/pull/531/commits/fd8b8b8615768273e890e8c841f2d14b60e3f924

It can be extracted if necessary

PR #531, opened 2024-09-04
unified HTML landing page (NOT a dashboard!)

willcl-ark:
- [x] Add a caddy reverse proxy
- [x] Add a landing page
- [x] Connect fork-observer
- [x] Connect grafana (thanks @m3dwards )
- [x] Docs
- [x] Update tests

### DIFF
--- a/resources/manifests/grafana_values.yaml
+++ b/resources/manifests/grafana_values.yaml
@@ -16,6 +16,9 @@ grafana.ini:
   auth:
     disable_login_form: true
     disable_signout_menu: true
+  server:
+    # this is required to use Grafana behind a reverse proxy (caddy)
+    root_url: "%(protocol)s://%(domain)s:%(http_port)s/grafana/"
   auth.anonymous:
     enabled: true
     org_name: Main Org.
--- a/src/warnet/dashboard.py
+++ b/src/warnet/dashboard.py
@@ -1,0 +1,11 @@
+import click
+
+
+@click.command()
+def dashboard():
+    """Open the Warnet dashboard in default browser"""
+    import webbrowser
+
+    url = "http://localhost:2019"
+    webbrowser.open(url)
+    click.echo("warnet dashboard opened in default browser")
--- a/src/warnet/deploy.py
+++ b/src/warnet/deploy.py
@@ -1,3 +1,6 @@
+import os
+import subprocess
+import sys
 import tempfile
 from pathlib import Path
 
@@ -16,7 +19,7 @@ from .constants import (
     NAMESPACES_FILE,
     NETWORK_FILE,
 )
-from .k8s import get_default_namespace
+from .k8s import get_default_namespace, wait_for_caddy_ready
 from .process import stream_command
 
 
@@ -102,13 +105,17 @@ def deploy_caddy(directory: Path, debug: bool):
         return
 
     namespace = get_default_namespace()
-    cmd = f"{HELM_COMMAND} 'caddy' {CADDY_CHART} --namespace {namespace}"
+    name = "caddy"
+    cmd = f"{HELM_COMMAND} {name} {CADDY_CHART} --namespace {namespace}"
     if debug:
         cmd += " --debug"
 
     if not stream_command(cmd):
         click.echo(f"Failed to run Helm command: {cmd}")
         return
+
+    wait_for_caddy_ready(name, namespace)
+    _port_start_internal()
 
 
 def deploy_fork_observer(directory: Path, debug: bool):
@@ -242,3 +249,43 @@ def deploy_namespaces(directory: Path):
         finally:
             if temp_override_file_path.exists():
                 temp_override_file_path.unlink()
+
+
+def is_windows():
+    return sys.platform.startswith("win")
+
+
+def run_detached_process(command):
+    if is_windows():
+        # For Windows, use CREATE_NEW_PROCESS_GROUP and DETACHED_PROCESS
+        subprocess.Popen(
+            command,
+            shell=True,
+            stdin=None,
+            stdout=None,
+            stderr=None,
+            close_fds=True,
+            creationflags=subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.DETACHED_PROCESS,
+        )
+    else:
+        # For Unix-like systems, use nohup and redirect output
+        command = f"nohup {command} > /dev/null 2>&1 &"
+        subprocess.Popen(command, shell=True, stdin=None, stdout=None, stderr=None, close_fds=True)
+
+    print(f"Started detached process: {command}")
+
+
+def _port_start_internal():
+    command = "kubectl port-forward service/caddy 2019:80"
+    run_detached_process(command)
+    click.echo(
+        "Port forwarding on port 2019 started in the background. To access landing page visit localhost:2019."
+    )
+
+
+def _port_stop_internal():
+    if is_windows():
+        os.system("taskkill /F /IM kubectl.exe")
+    else:
+        os.system("pkill -f 'kubectl port-forward service/caddy-service 2019:80'")
+    click.echo("Port forwarding stopped.")
--- a/src/warnet/deploy.py
+++ b/src/warnet/deploy.py
@@ -282,11 +282,11 @@ def run_detached_process(command):
 
 
 def _port_start_internal(name, namespace):
+    click.echo("Starting port-forwarding to warnet dashboard")
     command = f"kubectl port-forward -n {namespace} service/{name} 2019:80"
     run_detached_process(command)
-    click.echo(
-        "Port forwarding on port 2019 started in the background. To access landing page visit localhost:2019."
-    )
+    click.echo("Port forwarding on port 2019 started in the background.")
+    click.echo("\nTo access the warnet dashboard visit localhost:2019 or run:\n  warnet dashboard")
 
 
 def _port_stop_internal(name, namespace):
--- a/src/warnet/deploy.py
+++ b/src/warnet/deploy.py
@@ -15,6 +15,7 @@ from .constants import (
     FORK_OBSERVER_CHART,
     HELM_COMMAND,
     LOGGING_HELM_COMMANDS,
+    LOGGING_NAMESPACE,
     NAMESPACES_CHART_LOCATION,
     NAMESPACES_FILE,
     NETWORK_FILE,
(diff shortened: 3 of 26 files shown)


josibake:
> Uses the k8s API (it would be much less code to, as I had in my first iteration, to simply apply all *.yaml files in the caddy directory with subprocess). @m3dwards is this something we should also make a chart for like fork observer? I don't really know where the boundary of raw yaml files and charts lies, but this seemed simpler as a first approach

IMO we should be using helm charts to deploy everything, unless we have a specific reason why it _can't_ be done via helm. The manifests folder is really just a leftover and should be deleted once logging has been migrated to a helm chart, so I don't think we should be adding more to it. One of the goals of using helm was to use the python K8s API less, and to get rid of all of the k8s resource yaml files in the repo in favour of using helm charts.

To me, this feels like a regression back to what we were doing before and reintroduces some of the old risks, e.g., multiple ways of doing things, difficult to manage config, hardcoded things in different places, etc.

willcl-ark:
Great, that’s what I though too

willcl-ark:
TODO:

- [x] add auto-port-forward and tear-down
- [x] docs

josibake:
Nice, thanks for going the helm route! looks much better IMO. I have a PR that cleans up the config and puts all the resources / charts / directory paths in one place, could rebase on that, or I rebase on this if it merges first.

m3dwards:
This is quite a cool idea! I like the implementation.

Could even have a command `warnet not-dashboard` that opens the webpage for you!

I also can access FO but not grafana yet.

willcl-ark:
> This is quite a cool idea! I like the implementation.

Thanks, really curious to learn more about integrating this into the bitcoincore chart itself when you have time.

> Could even have a command `warnet not-dashboard` that opens the webpage for you!

`warnet dashboard` added in 585295f377c040475f988badbc797dc997ce57e6. Nice idea.

- fork-observer test updated to now test via the caddy proxy

256dummy:


### Conflicts
No conflicts as of last run.

josibake:
Concept ACK, very cool!

josibake:
started a review, looks good. in the teardown section, there is a work around due to the fact that commanders are not in a helm chart. id prefer we helmify the commander first and then rebase this? should make this PR simpler. this is also the second PR ive looked at that has a work around due to the commanders not being configured/deployed in the same way as everything else.

willcl-ark:
> started a review, looks good. in the teardown section, there is a work around due to the fact that commanders are not in a helm chart. id prefer we helmify the commander first and then rebase this? should make this PR simpler. this is also the second PR ive looked at that has a work around due to the commanders not being configured/deployed in the same way as everything else.

Attempted in #550

m3dwards:
@willcl-ark does Grafana work for you now? If so let's merge?

willcl-ark:
I think this needs a rebase too? (On mobile)

josibake:
now that #550 is merged, I think all this needs is a rebase and g2g?

willcl-ark:
Reabsed

willcl-ark:
Jesus more conflicts since I started the rebase! :'(

m3dwards:
I'm sorry, I feel your pain.

If it makes you feel better, I had it from you this week too!

willcl-ark:
Ok rebased again. The logging test didn't work for me locally, but it's pretty flakey still regarding whether it gets any data for me (usually), so will see what CI says

src/warnet/graph.py
```diff
@@ -180,6 +184,7 @@ def inquirer_create_network(project_path: Path):
         custom_network_path,
         fork_observer,
         fork_observer_query_interval,
+        fork_observer, # This enables caddy whenever fork-observer is enabled
```
willcl-ark: Hmmm, now that logging is enabled by default, perhaps caddy must also be enabled by default in this PR @m3dwards ?

josibake: I'd say yes? if the dashboard is meant to be how a user interacts with logs , then logs: enabled implies caddy:enabled in my mind.

m3dwards: Now I'm thinking about it I think the logic should be if logging / metrics or if fork observer then yes caddy. I'm not sure we need the ability to disable caddy.

m3dwards: And technically, logging isn't enabled by default as both logs and metrics are disabled in values.yaml. It's only enabled if metrics or logs are switched to on in your network definition.

resources/manifests/grafana_values.yaml
```diff
@@ -16,6 +16,9 @@ grafana.ini:
   auth:
     disable_login_form: true
     disable_signout_menu: true
+  server:
+    # this is required to use Grafana behind a reverse proxy (caddy)
+    root_url: "%(protocol)s://%(domain)s:%(http_port)s/grafana/"
```
josibake, line 21: just for my own understanding, grafana stuff is in `manifests/` for legacy reasons, yeah? was surprised its not also being deployed by helm (not a blocker for this PR tho)

src/warnet/graph.py
willcl-ark: mmm ok. I see now in https://github.com/bitcoin-dev-project/warnet/pull/565/commits/0c4f76c624826c408918f0859b463a4502a106dc which I need to remove also (didn't spot in the rebase)

m3dwards: See: https://github.com/bitcoin-dev-project/warnet/blob/main/src/warnet/deploy.py#L56

src/warnet/dashboard.py
```diff
@@ -0,0 +1,11 @@
+import click
+
+
+@click.command()
+def dashboard():
```
josibake, line 5: meh on having a command for this. seems easier to just prompt the user in the logs? also this assumes port forwarding, which I don't think will always be the case, so this command wont work if we move to having proper ingress. lastly, id assume this webbrowser thing will end up failing for some weird reason on someone's computer (cough cough windows).

src/warnet/deploy.py
```diff
@@ -90,6 +95,30 @@ def deploy_logging_stack(directory: Path, debug: bool):
...
+    namespace = "warnet-logging"
+    name = "caddy"
+
+    # Only start if configured in the network file
+    if not network_file.get(name, {}).get("enabled", False):
+        return
+
+    cmd = f"{HELM_COMMAND} {name} {CADDY_CHART} --namespace {namespace}"
+    if debug:
+        cmd += " --debug"
```
josibake, line 115: nice!

src/warnet/deploy.py
```diff
@@ -90,6 +95,30 @@ def deploy_logging_stack(directory: Path, debug: bool):
     return True
 
 
+def deploy_caddy(directory: Path, debug: bool):
+    network_file_path = directory / NETWORK_FILE
+    with network_file_path.open() as f:
+        network_file = yaml.safe_load(f)
+
+    namespace = "warnet-logging"
+    name = "caddy"
```
josibake: "warnet-logging" should be a constant in `constants.py` imo, and `name` should definitely be coming from the helm chart and not hardcoded here.

src/warnet/deploy.py
```diff
@@ -99,7 +128,8 @@ def deploy_fork_observer(directory: Path, debug: bool):
     if not network_file.get("fork_observer", {}).get("enabled", False):
         return
 
-    namespace = get_default_namespace()
+    default_namespace = get_default_namespace()
+    namespace = "warnet-logging"
```
josibake: same comment regarding constants.py

src/warnet/deploy.py
```diff
@@ -221,3 +251,43 @@ def deploy_namespaces(directory: Path):
         finally:
             if temp_override_file_path.exists():
                 temp_override_file_path.unlink()
+
+
+def is_windows():
+    return sys.platform.startswith("win")
```
josibake, line 261: doesn't need to be addressed in this PR , but this is code smell to me and indicates we really should stop doing port forwarding in favour of ingress (see https://github.com/bitcoin-dev-project/warnet/discussions/529)

src/warnet/k8s.py
```diff
@@ -115,3 +115,21 @@ def get_default_namespace() -> str:
     command = "kubectl config view --minify -o jsonpath='{..namespace}'"
     kubectl_namespace = run_command(command)
     return kubectl_namespace if kubectl_namespace else DEFAULT_NAMESPACE
+
+
+def wait_for_caddy_ready(name, namespace, timeout=300):
```
josibake, line 232: this also feels a bit smelly to me, as this is what helm is supposed to be managing for us. I didn't see where this is being used, but im guessing it's for the port forwarding?

josibake (commented):
ACK, overall looks good. I'm not crazy about all of the code just for port forwarding, and even less excited to see arch specific code like `is_windows` but not a blocker for this PR since port forwarding is the best we have right now. I do think this PR is a pretty good example of why we should move to proper ingress, even if it costs more in a k8s environment.

There are a few other nits that I think would be really nice to address if you end up touching this again, mainly around hard coded constants that I feel should either be in constants.py or in the helm chart itself, but not blocking. Will happily re-review if you end up touching this again.

src/warnet/deploy.py
willcl-ark: Yeah I think ingress can be its own PR

src/warnet/k8s.py
willcl-ark: Yes waiting for caddy to be ready. This is a k8s thing not helm? I can't concieve of any other way you'd do this for the port forward...

src/warnet/deploy.py
willcl-ark: This is modified from the fork-observer code which shoud also be updated.

Moved the constant, but getting the component name from the network file is not currently possible as this exists as a key only (so I would just be hardcoding the key here, named "caddy").

src/warnet/deploy.py
willcl-ark: also done

pinheadmz:
will conflict with #585 but we can rebase whichever (add docs!)

src/warnet/k8s.py
josibake: Yeah, sorry if my comment was unclear, more saying this `wait_for_caddy` function seems _specific_ to port forwarding, so we wouldn't need it if we weren't using port forwarding.

m3dwards (approved):
LGTM

m3dwards:
> will conflict with #585 but we can rebase whichever (add docs!)

Will has rebased this half a dozen times so I think it's only fair to merge this one.

PR #524, opened 2024-09-04
Add create-network cli command

m3dwards:
Look @pinheadmz, a test!

PR #536, opened 2024-09-06
Finish refactor of main. Align setup, new & init commands

### LINKED ISSUE #525: quickstart: Funny order of "how many nodes"
I think "Enter the number of nodes" should be asked immediately after "How many nodes would you like?" if the response is "other"

```
[?] Do you want to create a new project? (Y/n): 
[?] Enter the project directory path: wn100
[?] Do you want to create a custom network? (Y/n): 
[?] Enter your network name: wn100
[?] How many nodes would you like?: 
   8
   12
   20
   50
 > other

[?] How many connections would you like each node to have?: 
   0
   1
   2
 > 8
   12
   other

[?] Which version would you like nodes to run by default?: 
 > 27.0
   26.0
   25.1
   24.2
   23.2
   22.2

[?] Enter the number of nodes: 100
```

willcl-ark:
Follows up on #509
Fixes: #525


josibake:
https://github.com/bitcoin-dev-project/warnet/pull/524 repurposes `warnet create` to mean create a network (with a nice guided setup). haven't looked super closely at this PR but it looks like you are repurposing it to mean warnet create <dir>?

I prefer keeping create for "create a network", perhaps we could use warnet new here for starting a new project directory? id also be fine with just having warnet init.

docs/quickstart.md
```diff
@@ -28,13 +28,23 @@ pip install -e .
...
 
-This will check you have the required dependencies and guide you through setting up and deploying your first network.
+Then create your first network:
+
+```bash
+# Create a new network in the current directory
+warnet init
+
+# Or in a directory of choice
+warnet create <directory>
```
josibake: this is out of date after changing to the `warnet new` command

src/warnet/admin.py
```diff
@@ -33,3 +36,67 @@ def init():
         f"[green]Copied network and namespace example files to {Path(current_dir) / NETWORK_DIR.name}[/green]"
     )
     richprint(f"[green]Created warnet project structure in {current_dir}[/green]")
+
+
+@click.command()
+@click.argument("kube_config", type=str)
+def auth(kube_config: str) -> None:
```
josibake: hm, this is not an `admin` command. its meant to be used by warnet users , where an admin has created a cube-config file for them, which they then run `warnet auth <file>`, which will update their `kubectl` config and switch to the appropriate context. I like that you moved this out of `main.py` so I don't think moving it back to main is a good idea but not sure where it should live. Could go into `project.py` or we could create a `users.py` so that it's clear which commands are for the user and which commands are for the admin (by putting them in `admin.py`) ?

josibake:
Concept ACK, I really like that you're moving logic out of `main.py` into more appropriately named files. One small fix regarding documentation, and moving the `auth` command out of `admin.py` and I think this is good to go.

josibake:
ACK, lgtm. thanks for taking the suggestion regarding `warnet init/new` !

PR #538, opened 2024-09-06
Support signet in warnet

### LINKED ISSUE #503: Set up Signet-like miner for warnet
For war-games, we don't want teams to be able to mine their own blocks, which means we need an admin controlled miner. Signet is likely the easiest way to do this, so we will need to:

1. Run warnet in signet mode
2. Have a signet miner controlled by the war-games admin
3. Mine the blocks into each team's namespace (randomly? round robin?)

The idea of randomly mining into each teams namespace is to ensure blocks will come from enemy teams, thus each teams need to maintain connections to malicious peers to hear about new blocks. Otherwise, you could just firewall your nodes from any attacker.

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/503

Features added:
1. Support signet in warnet simply by setting `chain` in your network chart. Ports are set dynamically everywhere
2. Include new built-in scenario `signet_miner.py` which is Commander-wrapped proxy for Bitcoin Core `/contrib/signet/miner` and accepts all the same arguments with the addition of `--tank={index}`
3. Test

Reviewers should try:
1. Running test locally: `test/signet_test.py`
2. Checking that ports are set correctly when setting chain:
```
(.venv) --> helm template resources/charts/bitcoincore/ --set chain=regtest | grep port
    # rpcport and zmq endpoints are configured by chain in configmap.yaml
    rpcport=18443
  ports:
    - port: 18443
    - port: 18444
    - port: 28333
    - port: 28332
    - port: 9332
      ports:
          port: 18443

(.venv) --> helm template resources/charts/bitcoincore/ --set chain=signet | grep port
    # rpcport and zmq endpoints are configured by chain in configmap.yaml
    rpcport=38332
  ports:
    - port: 38332
    - port: 38333
    - port: 28333
    - port: 28332
    - port: 9332
      ports:
          port: 38332

```
3. (ADVANCED and not tested personally) run the signer miner *locally*:
    - `warnet deploy test/data/signet` to start the signet-chain warnet network
    - `cd` to your local bitcoin core repo
    - Start the signet miner locally using kubectl-forwarded bitcoin-cli (this is the crazy part I dunno if it will work):

```
python3 contrib/signet/miner \
    --cli=warnet bitcoin rpc miner
    generate
    --address=tb1q6vakuyw2jhzwmnxcaxryxs6c670fr9esfvrrhj
    --grind-cmd=src/bitcoin-util grind
    --min-nbits
    --ongoing
```

The PRO for this method is that you will get the benefit of mining with the C++ compiled `bitcoin-util` as opposed to the python grinder which is used in the scenario. CON is of course it is grinding locally and not in the cluster.

### DIFF
--- a/resources/charts/bitcoincore/values.yaml
+++ b/resources/charts/bitcoincore/values.yaml
@@ -112,14 +112,6 @@ collectLogs: false
 metricsExport: false
 prometheusMetricsPort: 9332
 
-regtestConfig: |
-  rpcuser=user
-  rpcpassword=password
-  rpcport=18443
-  rpcallowip=0.0.0.0/0
-  rpcbind=0.0.0.0
-  rest=1
-
 baseConfig: |
   checkmempool=0
   acceptnonstdtxn=1
@@ -132,6 +124,15 @@ baseConfig: |
 
   zmqpubrawblock=tcp://0.0.0.0:28332
   zmqpubrawtx=tcp://0.0.0.0:28333
+
+  rpcuser=user
+  rpcpassword=password
+  # use regtest port by default for all networks
+  rpcport=18443
+  rpcallowip=0.0.0.0/0
+  rpcbind=0.0.0.0
+  rest=1
+
 
 config: ""
 
--- a/resources/charts/bitcoincore/values.yaml
+++ b/resources/charts/bitcoincore/values.yaml
@@ -36,8 +36,13 @@ service:
 regtest:
   RPCPort: 18443
   P2PPort: 18444
-  ZMQTxPort: 28333
-  ZMQBlockPort: 28332
+
+signet:
+  RPCPort: 38332
+  P2PPort: 38333
+
+ZMQTxPort: 28333
+ZMQBlockPort: 28332
 
 ingress:
   enabled: false
@@ -82,8 +87,6 @@ readinessProbe:
   failureThreshold: 1
   periodSeconds: 1
   successThreshold: 1
-  tcpSocket:
-    port: 18443
   timeoutSeconds: 1
 
 
@@ -121,17 +124,12 @@ baseConfig: |
   capturemessages=1
   fallbackfee=0.00001000
   listen=1
-
-  zmqpubrawblock=tcp://0.0.0.0:28332
-  zmqpubrawtx=tcp://0.0.0.0:28333
-
   rpcuser=user
   rpcpassword=password
-  # use regtest port by default for all networks
-  rpcport=18443
   rpcallowip=0.0.0.0/0
   rpcbind=0.0.0.0
   rest=1
+  # rpcport and zmq endpoints are configured by chain in configmap.yaml
 
 
 config: ""
--- a/resources/scenarios/signet_miner.py
+++ b/resources/scenarios/signet_miner.py
@@ -1,0 +1,568 @@
+#!/usr/bin/env python3
+# Copyright (c) 2020 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+###
+# WARNET SCENARIO SIGNET MINER
+#
+# This file is a hacked-up version of Bitcoin Core contrib/signet/miner.
+# The primary difference is that instead of using bitcoin-cli for RPCs
+# we use the authproxy from the test framework.
+###
+
+# The base class exists inside the commander container
+try:
+    from commander import Commander
+except ImportError:
+    from resources.scenarios.commander import Commander
+
+import json
+import logging
+import math
+import re
+import struct
+import sys
+import time
+import subprocess
+
+from test_framework.blocktools import get_witness_script, script_BIP34_coinbase_height # noqa: E402
+from test_framework.messages import CBlock, CBlockHeader, COutPoint, CTransaction, CTxIn, CTxInWitness, CTxOut, from_binary, from_hex, ser_string, ser_uint256, tx_from_hex # noqa: E402
+from test_framework.psbt import PSBT, PSBTMap, PSBT_GLOBAL_UNSIGNED_TX, PSBT_IN_FINAL_SCRIPTSIG, PSBT_IN_FINAL_SCRIPTWITNESS, PSBT_IN_NON_WITNESS_UTXO, PSBT_IN_SIGHASH_TYPE # noqa: E402
+from test_framework.script import CScriptOp # noqa: E402
+
+logging.basicConfig(
+    format='%(asctime)s %(levelname)s %(message)s',
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S')
+
+SIGNET_HEADER = b"\xec\xc7\xda\xa2"
+PSBT_SIGNET_BLOCK = b"\xfc\x06signetb"    # proprietary PSBT global field holding the block being signed
+RE_MULTIMINER = re.compile(r"^(\d+)(-(\d+))?/(\d+)$")
+
+
+class SignetMinerScenario(Commander):
+    def set_test_params(self):
+        self.num_nodes = 1
+
+    def add_options(self, parser):
+        parser.add_argument(
+            "--tank",
+            dest="tank",
+            type=int,
+            help="Index of tank with wallet loaded for block signing",
+        )
+        get_args(parser)
+
+
+    def run_test(self):
+        args = self.options
+        args.bcli = lambda method, *args, **kwargs: self.nodes[self.options.tank].__getattr__(method)(*args, **kwargs)
+        return do_generate(args)
+
+
+def create_coinbase(height, value, spk):
+    cb = CTransaction()
+    cb.vin = [CTxIn(COutPoint(0, 0xffffffff), script_BIP34_coinbase_height(height), 0xffffffff)]
+    cb.vout = [CTxOut(value, spk)]
+    return cb
+
+def signet_txs(block, challenge):
+    # assumes signet solution has not been added yet so does not need
+    # to be removed
+
+    txs = block.vtx[:]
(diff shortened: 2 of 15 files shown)


resources/charts/bitcoincore/values.yaml
```diff
@@ -121,17 +124,12 @@ baseConfig: |
...
-  zmqpubrawtx=tcp://0.0.0.0:28333
-
   rpcuser=user
   rpcpassword=password
-  # use regtest port by default for all networks
-  rpcport=18443
   rpcallowip=0.0.0.0/0
   rpcbind=0.0.0.0
   rest=1
+  # rpcport and zmq endpoints are configured by chain in configmap.yaml
```
josibake, line 132: interesting!

src/warnet/bitcoin.py
```diff
@@ -36,10 +36,12 @@ def rpc(tank: str, method: str, params: str):
 
 
 def _rpc(tank: str, method: str, params: str):
+    # bitcoin-cli should be able to read bitcoin.conf inside the container
+    # so no extra args like port, chain, username or password are needed
```
josibake, line 40: this is v cool, with having bitcoin.conf drive a lot of config via the configmap.

resources/scenarios/signet_miner.py
```diff
@@ -0,0 +1,568 @@
...
+from test_framework.script import CScriptOp # noqa: E402
+
+logging.basicConfig(
+    format='%(asctime)s %(levelname)s %(message)s',
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S')
+
+SIGNET_HEADER = b"\xec\xc7\xda\xa2"
+PSBT_SIGNET_BLOCK = b"\xfc\x06signetb"    # proprietary PSBT global field holding the block being signed
+RE_MULTIMINER = re.compile(r"^(\d+)(-(\d+))?/(\d+)$")
```
josibake, line 41: iirc, this puts us on our own "private signet," correct?

josibake (commented):
very cool! super excited to see that the bulk of the change is adding the signet miner. will test and comment back if I find anything

josibake:
not a critical fix, but if you're not in the warnet directory , the test fails:

```
  File "/root/wtest/../warnet/test/signet_test.py", line 52, in <module>
    test.run_test()
  File "/root/wtest/../warnet/test/signet_test.py", line 21, in run_test
    self.check_signet_miner()
  File "/root/wtest/../warnet/test/signet_test.py", line 36, in check_signet_miner
    self.warnet(
  File "/root/warnet/test/test_base.py", line 74, in warnet
    raise Exception(proc.stderr.decode().strip())
Exception: Usage: warnet run [OPTIONS] SCENARIO_FILE [ADDITIONAL_ARGS]...
Try 'warnet run --help' for help.
```

which I think is easily fixed by just referencing the file using the `files(...)` approach (this should be easier after my constants.py / common.py PR gets merged)

test/data/signet/node-defaults.yaml
```diff
@@ -0,0 +1,6 @@
+image:
+  repository: bitcoindevproject/bitcoin
+  pullPolicy: IfNotPresent
+  tag: "27.0"
+
+chain: signet
```
josibake: I'd suggest updating the default example to explicitly have the `chain: regtest` argument specified. While testing, I had to come back to your PR to remember where to put the chain field.

josibake:
Deployed a network by setting `chain: signet` on the 6_node network and then tried to naively deploy a signet miner without setting any arguments: `warnet run scenarios/signet_miner.py`. Got the following error, which I found to be a bit strange/obtuse:

```
  File "/test_framework/test_framework.py", line 131, in main
    self.run_test()
  File "/scenario.py", line 61, in run_test
    return do_generate(args)
2024-09-06 16:12:56 ERROR Unexpected exception caught during testing
  File "/test_framework/test_framework.py", line 131, in main
           ^^^^^^^^^^^^^^^^^
  File "/scenario.py", line 289, in do_generate
    if args.max_blocks is not None:
       ^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'max_blocks'
    self.run_test()
  File "/scenario.py", line 61, in run_test
    return do_generate(args)
           ^^^^^^^^^^^^^^^^^
  File "/scenario.py", line 289, in do_generate
    if args.max_blocks is not None:
       ^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'max_blocks'
```

josibake:
Might be unrelated to your change, but I can't import a descriptor directly:

```
warnet bitcoin rpc tank-0001 importdescriptors '[{"desc": "wpkh(tprv8ZgxMBicQKsPfH87iaMtrpzTkWiyFDW7SVWqfsKAhtyEBEqMV6ctPdtc5pNrb2FpSmPcDe8NrxEouUnWj1ud7LT1X1hB1XHKAgB2Z5Z4u2s/84h/1h/0h/0/*)#5j6mshps", "timestamp": 0, "active": true, "internal": false, "range": [0, 999], "next": 0, "next_index": 0}, {"desc": "wpkh(tprv8ZgxMBicQKsPfH87iaMtrpzTkWiyFDW7SVWqfsKAhtyEBEqMV6ctPdtc5pNrb2FpSmPcDe8NrxEouUnWj1ud7LT1X1hB1XHKAgB2Z5Z4u2s/84h/1h/0h/1/*)#9xl6dz3g", "timestamp": 0, "active": true, "internal": true, "range": [0, 999], "next": 0, "next_index": 0}]'

error: Error parsing JSON: [{desc:
command terminated with exit code 1
```

edit: this is the descriptor from the test data, so it should work. I was also having trouble importing it using jq, and ended up reading it in via python and dumping it again directly just like the test, and that's when I got this error.

willcl-ark:
Another big concept ACK from me! Greta work @josibake and @pinheadmz on this signet stuff!

Will try to review this later too.

Josie: passing a descriptor like that seems like a big strong escape nightmare potentially? May have to escape all " marks?

josibake:
> Josie: passing a descriptor like that seems like a big strong escape nightmare potentially? May have to escape all " marks?

yeah, ended up needing to escape all my double quotes. I think we can probably be a bit smarter than this for rpc commands that expect json in warnet , but its not related to this PR so im leaving it alone for now

resources/scenarios/signet_miner.py
pinheadmz: no the scenario is is a copy/paste from bitcoin core with a few changes to redirect the RPCs. The "private signet" is configured by the `-signetchallenge` setting, which I believe also sets a custom magic bytes. Not sure if that's the same as the SIGNET_HEADER here

pinheadmz:
> not a critical fix, but if you're not in the warnet directory , the test fails:

Probably an issue with all the tests...

test/data/signet/node-defaults.yaml
pinheadmz: isnt the default set in `charts/bitcoincore/values.yml` ?

pinheadmz:
> AttributeError: 'Namespace' object has no attribute 'max_blocks'

I'll look into this, the signet miner script (from bitcoin core) should throw if a required arg isnt set, and there are several required args for the script.

josibake:
ACK from me, ran the tests, tried to recreate the test on my own using the signet_miner.py and succeeded!

josibake:
Circling back, I was able to signet mine using `contrib/signet/miner` on my local machine. I just needed to apply the following patch to the miner:

```diff
diff --git a/contrib/signet/miner b/contrib/signet/miner
index 4216ada5fa..5ffd49235d 100755
--- a/contrib/signet/miner
+++ b/contrib/signet/miner
@@ -386,7 +386,7 @@ def do_generate(args):
             continue

         # gbt
-        tmpl = json.loads(args.bcli("getblocktemplate", '{"rules":["signet","segwit"]}'))
+        tmpl = json.loads(args.bcli("getblocktemplate", "'{\"rules\":[\"signet\",\"segwit\"]}'"))
         if tmpl["previousblockhash"] != bci["bestblockhash"]:
             logging.warning("GBT based off unexpected block (%s not %s), retrying", tmpl["previousblockhash"], bci["bestblockhash"])
             time.sleep(1)
@@ -408,8 +408,7 @@ def do_generate(args):
         logging.debug("Mining block delta=%s start=%s mine=%s", seconds_to_hms(mine_time-bestheader["time"]), mine_time, is_mine)
         mined_blocks += 1
         psbt = generate_psbt(tmpl, reward_spk, blocktime=mine_time)
-        input_stream = os.linesep.join([psbt, "true", "ALL"]).encode('utf8')
-        psbt_signed = json.loads(args.bcli("-stdin", "walletprocesspsbt", input=input_stream))
+        psbt_signed = json.loads(args.bcli("walletprocesspsbt", psbt, "true", "ALL"))
         if not psbt_signed.get("complete",False):
             logging.debug("Generated PSBT: %s" % (psbt,))
             sys.stderr.write("PSBT signing failed\n")
@@ -418,7 +417,7 @@ def do_generate(args):
         block = finish_block(block, signet_solution, args.grind_cmd)

         # submit block
-        r = args.bcli("-stdin", "submitblock", input=block.serialize().hex().encode('utf8'))
+        r = args.bcli("submitblock", block.serialize().hex().encode('utf8'))

         # report
         bstr = "block" if is_mine else "backup block"

```

I think we can probably fix this on the warnet side so that it it handles json a bit better when being passed over bitcoin cli. Also, need to quote the `cli` and `grind-cmd` args like so:

```bash
python3 contrib/signet/miner \
    --cli="warnet bitcoin rpc miner"
    generate
    --address=tb1q6vakuyw2jhzwmnxcaxryxs6c670fr9esfvrrhj
    --grind-cmd="src/bitcoin-util grind"
    --min-nbits
    --ongoing
```

PR #539, opened 2024-09-06
add CVE images with signet (based on #498)

### LINKED ISSUE #532: Rebuild old CVE images with support for signet
Per #503 , signet was introduced in 0.21.0 which means we can't use the signet approach for old CVE images. I took a look at the PR for signet and it seems relatively straightforward to back port. Going to take a stab at back porting images, starting with the oldest one first. Depending on how hard this is, I'll either:

- give up
- have all of the images back ported by Thursday of next week

This unblocks being able to use a signet miner to restrict other players from mining in a multi-player setting.

## Images to backport

- [x] 0.16.1
- [x] 0.17
- [x] 0.19.2
- [x] 0.20

josibake:
mostly a refactor of @willcl-ark 's PR to have one base image that can be used to build different older versions of Bitcoin Core. 

Added ccache, which speeds things up quite a bit and also local development, i.e., compile from your local repo instead of only from a remote. This is particularly useful when trying to build older images/ back port changes, etc.

The base image might seem like overkill for now, but I think having this will make it much easier to add new CVE images as they are announced and also makes it very easy to fix up the existing CVE images if we find issues with the signet back port.

closes #532 

---

Not covered in this PR but good follow-ups:

- Add a test for CVE images that ensures they are able to connect on signet
- Add documentation for using the docker image with local development 
- Add a test in our CI to make sure the patches are working correctly for each of the old images

### DIFF
--- a/resources/images/bitcoin/insecure/entrypoint.sh
+++ b/resources/images/bitcoin/insecure/entrypoint.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+set -e
+
+# Only run this tor section if the TOR=1 env var is set
+if [ "${TOR:-0}" -eq 1 ]; then
+
+    # Add bitcoin user to tor group to read the auth cookie
+    usermod -a -G debian-tor bitcoin
+
+    # ===========Tor setup===========
+    # Use custom torrc for warnet
+    if [ "${WARNET:-0}" -eq 1 ]; then
+        mv /etc/tor/warnet-torrc /etc/tor/torrc
+    fi
+    echo "Address $(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)" >> /etc/tor/torrc
+    mkdir -p /home/debian-tor/.tor/keys
+    chown -R debian-tor:debian-tor /home/debian-tor
+    # Start tor in the background
+    su-exec debian-tor:debian-tor tor &
+    # ===============================
+fi
+
+if [ "$(echo "$1" | cut -c1)" = "-" ]; then
+  echo "$0: assuming arguments for bitcoind"
+
+  set -- bitcoind "$@"
+fi
+
+if [ "$(echo "$1" | cut -c1)" = "-" ] || [ "$1" = "bitcoind" ]; then
+  mkdir -p "$BITCOIN_DATA"
+  chmod 700 "$BITCOIN_DATA"
+  echo "$0: setting data directory to $BITCOIN_DATA"
+  set -- "$@" -datadir="$BITCOIN_DATA"
+fi
+
+# Incorporate additional arguments for bitcoind if BITCOIN_ARGS is set.
+if [ -n "$BITCOIN_ARGS" ]; then
+  IFS=' ' read -ra ARG_ARRAY <<< "$BITCOIN_ARGS"
+  set -- "$@" "${ARG_ARRAY[@]}"
+fi
+
+# Conditionally add -printtoconsole for Bitcoin version 0.16.1
+if [ "${BITCOIN_VERSION}" == "0.16.1" ]; then
+  exec "$@" -printtoconsole
+else
+  exec "$@"
+fi
--- a/resources/images/bitcoin/insecure/entrypoint.sh
+++ b/resources/images/bitcoin/insecure/entrypoint.sh
@@ -1,25 +1,6 @@
 #!/bin/bash
 set -e
 
-# Only run this tor section if the TOR=1 env var is set
-if [ "${TOR:-0}" -eq 1 ]; then
-
-    # Add bitcoin user to tor group to read the auth cookie
-    usermod -a -G debian-tor bitcoin
-
-    # ===========Tor setup===========
-    # Use custom torrc for warnet
-    if [ "${WARNET:-0}" -eq 1 ]; then
-        mv /etc/tor/warnet-torrc /etc/tor/torrc
-    fi
-    echo "Address $(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)" >> /etc/tor/torrc
-    mkdir -p /home/debian-tor/.tor/keys
-    chown -R debian-tor:debian-tor /home/debian-tor
-    # Start tor in the background
-    su-exec debian-tor:debian-tor tor &
-    # ===============================
-fi
-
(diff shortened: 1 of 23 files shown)


willcl-ark:
Big concept ACK from me.

Will try to review this evening

willcl-ark:
I should close #498 now?

josibake:
> I should close https://github.com/bitcoin-dev-project/warnet/pull/498 now?

Id say so, so long as people like the approach here? Main differences are:

- one single parameterised image
- no longer applying patches during the image building

josibake:
Note to self: when building multiple images, it seems the BDB stage is not being cached. I _think_ this is because of changing the alpine images in between builds, in which case there is nothing that can be done, but will investigate further.

josibake:
Seems like the BDB stage is not caching due to a wget call. Not a big deal, but if we want to speed up image building we could just add the tar.gz asset to our repo which I think would make these builds cacheable

josibake:
There's a bug where the entrypoint.sh is not being copied over and I haven't been able to figure out what's going on. Putting in draft until I have a chance to fix this.

josibake:
Fixed the entrypoint.sh bug, turns out the file was being copied correctly but I had removed `bash` from the runtime dependencies list, and there was a typo in the `EXTRA_PACKAGES_RUNTIME` variable, so bash was never getting installed and thus the docker image couldn't run the entrypoint.sh script.

Have tested 16.1, 0.19.2, 0.20.0, 0.21.1 and all work in signet mode. 0.19.2 and 0.20.0 seem to be bypassing the signet challenge and doing IBD on regular signet, will update the back port with a fix. 

0.17 fails when getting a new block from a signet peer, but I think I've spotted the issue and will push a fix for that as well. As far as this PR goes, I think the Docker image itself is good to go (as I've been using it locally for testing/building images) so I think this can reviewed/merged as is, and I'll update build.md with the correct `commit_sha`s for 0.17, 0.19.2, and 0.20 as I fix the issues with them.

josibake:
cc @willcl-ark would very much appreciate a review here since you wrote a lot of the original

pinheadmz:
getting image pull errors (from josibake) for :

      tag: "24.2"
      tag: "25.1"
      tag: "26.0"
      tag: "27.0"

pinheadmz:
Did you push them? https://hub.docker.com/r/josibake/bitcoin/tags

josibake:
> Did you push them? https://hub.docker.com/r/josibake/bitcoin/tags

yep! you should see tags for all of the CVE images for both linux/arm. I've been testing them by changing my node-defaults.yaml to this:

image

pinheadmz:
I don't see any tags after 21

pinheadmz:
Oh right. Those versions have signet already 🙈

josibake:
@pinheadmz if you want to mix images from my repo and bitcoindevproject in the same network, its super easy:

image

(I also uploaded the project directory I've been using for testing here: https://github.com/josibake/warnet-projects/tree/main/cve-image-testing)

pinheadmz:
Yep I see that is very easy now. I will test like that tonight. Sure would be cool if we had a CI that built new images only if there were any changes, and used those images in the test. Otherwise pull from docker hub. And then on merge to main, push those images to hub. (@willcl-ark ?)

pinheadmz:
Ultimately these images should be tested in the existing signet test but they will fail CI until they are pushed to dockerhub, which is  not  garunteed to be in sync with GitHub.

willcl-ark:
What we do in the commander is use PullIfNotPresent, building if image files change (and load into subsequent jobs/workflows)

resources/images/bitcoin/insecure/entrypoint.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+set -e
+
+# Only run this tor section if the TOR=1 env var is set
```
willcl-ark: We could take this out, potentially (to be re-added later when we re-add Tor?)

willcl-ark (commented):
utACK, looks nice! 

Can test over the next few days I hope (busy/backlogged rn)

pinheadmz:
I tried testing the josi + default images in the signet test, but had trouble even getting all tanks running. Once or twice it got to the "edges connected" check but that failed because old nodes don't have "manual", they have "addnode". I fixed that but then couldn't even get that far again.

I am giving up for the night, here's the patch that adds the new images to the test:

https://gist.github.com/pinheadmz/024dfd00902e50a0036f0c13ab9e93b6

josibake:
> I tried testing the josi + default images in the signet test, but had trouble even getting all tanks running

I was running into this, as well. This isn't related to the signet changes, rather it's that our Commander/test code doesn't work well with old images of Bitcoin Core.

One suggestion would be we merge this and push the images to bitcoindevproject so we can start updating the signet test and other tests? FWIW, I've been testing the signet logic pretty extensively, so any failures we run into now are likely related to making our test/commander code work with old images. I think it would be nicer to have those as smaller PRs , rather than continuing to expand this one?

> Sure would be cool if we had a CI that built new images only if there were any changes, and used those images in the test

Agree this would be nice, only caveat is it will take a _looooong_ time to build these images in CI and they need to be built cross platform.

resources/images/bitcoin/insecure/entrypoint.sh
josibake: Agree, and if we re-add Tor I'm not even sure we would do it exactly the same way. Will remove to keep the entrypoint script simple.

pinheadmz:
I've also seen these images fail in Warnet outside of the test. There's also no commander involved in either case. The test is just kubernetes anyway. 

You can add my commit to your branch to test in CI? When you push your images to bdp we can edit the yaml for the test

josibake:
> You can add my commit to your branch to test in CI? When you push your images to bdp we can edit the yaml for the test

Yep! Images are pushed to BDP, will add you test in a commit and make sure its passing before merging. Regarding test failures, as mentioned 0.16.1 is not supported on arm64 (macOS), so that is something we will need to backport. Since that's not needed before Tabconf, I'll make a backlog issue for now and revisit it later.

For any other failures, I expect they will be either 1) outdated rpcs/arguments coming from our commander/test framework code or 2) issues in the actual images themselves.

For 2), if you run into something I'd say create a issue with steps to reproduce and then I can update the tagged commit_sha with a fix and rebuild the image.

PR #543, opened 2024-09-07
Give users appropriate access to resources

mplsgrant:
The users need access to various resources to launch scenarios and view scenario logs.

I also included the "fixup namespace directory constants" here because it's needed.


resources/namespaces/two_namespaces_two_users/namespaces.yaml
```diff
@@ -14,11 +14,35 @@ namespaces:
...
+            verbs: ["get"]
+          - apiGroups: [""]
+            resources: ["configmaps", "secrets"]
+            verbs: ["get"]
+          - apiGroups: [""]
+            resources: ["persistentvolumeclaims"]
+            verbs: ["get", "list"]
+          - apiGroups: [""]
+            resources: ["events"]
+            verbs: ["get"]
```
josibake, line 28: seems like these could all be set on the namespace-defaults.yaml file? that way we only specify rules in the namespaces.yaml when we want to override defaults for a particular user/namespace.

resources/charts/namespaces/values.yaml
```diff
@@ -9,8 +9,32 @@ roles:
...
+        verbs: ["get", "list"]
+      - apiGroups: [""]
+        resources: ["events"]
+        verbs: ["get"]
   - name: pod-manager
     rules:
       - apiGroups: [""]
         resources: ["pods"]
-        verbs: ["get", "list", "watch", "create", "update", "delete"]
+        verbs: ["get", "list", "watch", "create", "delete", "update"]
```
josibake, line 28: longterm, we might want to have the `values.yaml` provide _no_ permissions by default (or the most minimal possible), and then have sensible defaults in namespace-defaults.yaml , specific to the "war-games" use case. then namespaces.yaml is simply for overriding defaults as needed.

(what you have here is fine, more just thinking out loud of how we might want to structure this long term)

josibake (commented):
nice! left a few comments that I think would make this less verbose and some ideas about how to structure this long term.

from a permissions standpoint, I think its nice to have values.yaml not give anything by default, have namespace-defaults.yaml give the needed permissions for running a user in a namespace, and then namespaces.yaml can be used to apply overrides as needed.

josibake:
#544 merged, can be rebased on main

mplsgrant:
Thanks for the feeback @josibake 

I rebased on main but will leave the rest as-is for now. I like your ideas about trimming the files down.

josibake:
> I rebased on main but will leave the rest as-is for now

Looking at this again, I do think the `namespaces.yaml` file needs to be de-duplicated before merging. Per https://github.com/bitcoin-dev-project/warnet/blob/main/docs/config.md, `values.yaml` is meant to be overridden by `namespace-defaults.yaml` and `namespace-defaults.yaml` is meant to be overridden by `namespaces.yaml`. What you have currently is applying the same defaults multiple times in `namespaces.yaml`, whereas what is currently in `namespace-defaults.yaml` is incorrect/outdated.

The reason this is important is a user is expected to to be able to create a `custom-namespaces.yaml` config and specify just the namespaces and users they want without requiring any additional config, because the appropriate permissions are defined in `namespace-defaults.yaml`. In the event a user wants to change defaults for all namespaces in their custom deployment, they would edit the `namespace-defaults.yaml` file. In the event they want to make a change for a _specific_ user in a _specific_ namespace, they would apply those changes in the `custom-namespaces.yaml` file.

josibake:
Merging this as is for now (since its just changing defaults for configs) and will clean up in a follow up by me.

PR #544, opened 2024-09-07
fixup namespace directory constants

mplsgrant:
Some namespace dirs need to be fixed up so that `warnet admin init` works.


josibake (approved):
ACK, thanks for fixing. I think I broke this in my constants.py refactor :sweat_smile:

PR #545, opened 2024-09-07
add  create-kubeconfigs command

mplsgrant:
I transmuted the setup_user_contexts.sh script into python code.

I also made it so that it would query all namespaces that start "warnet-" looking for service accounts,
and then create token credentials for those accounts.

Currently, user namespaces must start with "warnet-".

Also, I included the commits from #544 and #543 because they are necessary for this.


src/warnet/admin.py
```diff
@@ -1,4 +1,5 @@
 import os
+import subprocess
```
josibake: do we need subprocess? iirc, we have a warnet command runner (run_command, or stream_command?)

src/warnet/admin.py
```diff
@@ -33,3 +34,132 @@ def init():
...
+
+# Get kubectl values
+def get_kubectl_value(jsonpath):
+    return subprocess.check_output(
+        ["kubectl", "config", "view", "--minify", "-o", f"jsonpath={jsonpath}"]
+    ).decode("utf-8")
+
+
+# Get all namespaces that start with "warnet-"
+def get_warnet_namespaces():
```
josibake, line 47: if we are going to start having conventions around namespace strings (e.g., warnet-), might be more useful to make this a "get_namespaces_by_prefix(prefix: str)" function. not urgent, though

src/warnet/admin.py
```diff
@@ -33,3 +34,132 @@ def init():
...
+                    .strip()
+                )
+            except subprocess.CalledProcessError:
+                click.echo(f"Failed to create token for ServiceAccount {sa}. Skipping...")
+                continue
+
+            # Create a kubeconfig file for the user
+            kubeconfig_file = os.path.join(kubeconfig_dir, f"{sa}-{namespace}-kubeconfig")
+
+            kubeconfig_content = f"""apiVersion: v1
```
josibake, line 135: works for now, but ideally we keep the k8s yamls out of the python code. maybe just leave a #TODO comment to move this at some point ?

josibake (commented):
+1 for replacing another bash script! LGTM, I think we can merge as is in the interest of time, but would be nice to leave some TODO hints for follow ups.

mplsgrant:
I added your comments as TODOs

pinheadmz:
Pretty sure I understand what this is doing but it would be nice to have a test or docs to read so I know who to try it out myself, especially this week with our pre-tabconf trials

src/warnet/admin.py
pinheadmz: I agree with @josibake but more strongly. I think we should avoid using any naming conventions like we did before helm (if possible) -- can we use kubernetes metadata "labels" instead?

josibake:
> Pretty sure I understand what this is doing but it would be nice to have a test or docs to read

I'll take these tomorrow and 1) add the changes I proposed in #543 and 2) add a README on how to deploy namespaces for multiple teams.

As an fyi, this all works today, these PRs are just setting/documenting sane defaults in the yaml files, so should be a very easy change.

josibake:
will rebase this on master now that #543 is merged and address the outstanding comments from @pinheadmz and @me

josibake:
This is continued in https://github.com/bitcoin-dev-project/warnet/pull/598

PR #553, opened 2024-09-08
add k8s log collector to CI

willcl-ark:
Currently, it's *very* painful to debug inconsistencies in the CI vs a local environment. Additionally, when things go wrong, we don't get any kubernetes output at all. 

Fix this by adding a job which uploads artefacts to GH for each job.

`stern` allows tailing of all pods (in all namespaces), and can dump to json (I chose `default` mode here).

See an example of the artefacts here: https://github.com/willcl-ark/warnet/actions/runs/10763746456

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,6 +91,27 @@ jobs:
         run: |
           source .venv/bin/activate
           ./test/${{matrix.test}}
+      - name: Collect Kubernetes logs
+        if: always()
+        run: |
+          echo "Installing stern..."
+          STERN_VERSION="1.30.0"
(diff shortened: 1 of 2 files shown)


josibake:
ACK, +1 for making CI failures easier to debug. I don't have any expertise in the GA area, but this seems reasonable to me.

> (I chose default mode here

why not founder mode?

willcl-ark:
> ACK, +1 for making CI failures easier to debug. I don't have any expertise in the GA area, but this seems reasonable to me.
> 
> > (I chose default mode here
> 
> why not founder mode?

Ok, I'm going full founder mode. Just ordered one of these in each colour for each of us: https://levelsio.com/products/founder-mode-hat-100-cotton

josibake:
needs a rebase and would be great to get someone else's opinion who is more familiar with GA / our CI, cc @pinheadmz

josibake:
LGTM and seems useful, if this ends up being problematic for GA we can always revert.

PR #555, opened 2024-09-09
create and load tank snapshots

josibake:
Given a url, download the blocks and start a node with this chain. This works by creating a temporary initContainer which downloads the file and untars it into the .bitcoin data dir. Once complete, the initContainer cleans itself up and the pod starts as normal.

Other nodes get this chain by doing IBD, or each node can be specified to start with its own snapshot. This could be useful for things like simulating a hard fork, chain split, etc.

## For testing

I mined a signet chain to 200 blocks and created a snapshot using `warnet snapshot miner -f "chainstate,blocks,wallets"`. The network along with the snapshot is saved as a warnet project here: https://github.com/josibake/warnet-projects/tree/main/premined-signet

To test:

* Checkout the project: `git clone git@github.com:josibake/warnet-projects.git && cd warnet-projects/premined-signet`
* Make sure you have this PR checked out and install warnet: `uv venv && source .venv/bin/activate.fish && uv pip install -e ../../warnet`
* Run `warnet deploy networks/cve_images_on_signet/`
* Load the wallet for the miner: `warnet bitcoin rpc miner loadwallet mining_wallet`
* Run `warnet bitcoin rpc miner -getinfo`

You should see something like the following:

```
Chain: signet
Blocks: 200
Headers: 200
Verification progress: 100.0000%
Difficulty: 0.001126515290698186

Network: in 5, out 0, total 5
Version: 270000
Time offset (s): 0
Proxies: n/a
Min tx relay fee rate (BTC/kvB): 0.00001000

Wallet: mining_wallet
Keypool size: 4000
Transaction fee rate (-paytxfee) (BTC/kvB): 0.00000000

Balance: 5000.00000000

Warnings: (none)
```

.. which means your miner started with a premined chain and 5000 available coins in the wallet!

## Bonus

1. Check some of the other nodes and verify they are on the same 200 block chain (or doing IBD)
2. Run the signet miner to make sure you are able to keep mining on top of the premined chain, i.e., `warnet run scenarios/signet_miner.py --tank=0 generate --min-nbits --address=tb1qnsv03jqd7tkn2crd7jqv5wxwf2mmxl2l23w7rs --ongoing`
3. After mining a few blocks on the new chain, run `warnet down && warnet deploy networks/cve_images_on_signet/` and verify you are back at your starting point of a 200 block chain  

---

TODO:

- [ ] add a test
- [x] add docs for how to create a premined chain and where to upload it

### DIFF
--- a/resources/charts/bitcoincore/values.yaml
+++ b/resources/charts/bitcoincore/values.yaml
@@ -134,3 +134,6 @@ baseConfig: |
 config: ""
 
 connect: []
+loadSnapshot:
+  enabled: false
+  url: ""
--- a/docs/snapshots.md
+++ b/docs/snapshots.md
@@ -1,0 +1,94 @@
+# Creating and loading warnet snapshots
+
+The `snapshot` command allows users to create snapshots of Bitcoin data directories from active bitcoin nodes. These snapshots can be used for backup purposes, to recreate specific network states, or to quickly initialize new bitcoin nodes with existing data.
+
+## Usage Examples
+
+### Snapshot a Specific bitcoin node
+
+To create a snapshot of a specific bitcoin node:
+
+```bash
+warnet snapshot my-node-name -o <snapshots_dir>
+```
+
+This will create a snapshot of the bitcoin node named "my-node-name" in the `<snapshots_dir>`. If a directory the directory does not exist, it will be created. If no directory is specified, snapshots will be placed in `./warnet-snapshots` by default.
+
+### Snapshot all nodes
+
+To snapshot all running bitcoin nodes:
+
+```bash
+warnet snapshot --all -o `<snapshots_dir>`
+```
+
+### Use Filters
+
+In the previous examples, everything in the bitcoin datadir was included in the snapshot, e.g., peers.dat. But there maybe use cases where only certain directories are needed. For example, assuming you only want to save the chain up to that point, you can use the filter argument:
+
+```bash
+warnet snapshot my-node-name --filter "chainstate,blocks"
+```
+
+This will create a snapshot containing only the 'blocks' and 'chainstate' directories. You would only need to snapshot this for a single node since the rest of the nodes will get this data via IBD when this snapshot is later loaded. A few other useful filters are detailed below:
+
+```bash
+# snapshot the logs from all nodes
+warnet snapshot --all -f debug.log -o ./node-logs
+
+# snapshot the chainstate and wallet from a mining node
+# this is particularly userful for premining a signet chain that
+# can be used later for starting a signet network
+warnet snapshot mining-node -f "chainstate,blocks,wallets"
+
+# snapshot only the wallets from a node
+warnet snapshot my-node -f wallets
+
+# snapshot a specific wallet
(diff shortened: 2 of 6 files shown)


resources/charts/bitcoincore/values.yaml
```diff
@@ -135,3 +135,6 @@ baseConfig: |
 config: ""
 
 connect: []
+blocksDownload:
+  enabled: false
+  url: "https://storage.googleapis.com/warnet-signet-chains/0014d33b6e11ca95c4edccd8e986434358d79e919730/premined-signet.tar.gz"
```
pinheadmz: this kind of hard-code is ok in a test but i dont think it belongs in a default

josibake: agreed, added "do not merge" to the title of the PR and added this as a todo in the description, will try to fix tomorrow.

m3dwards:
This is very cool and I'm sure will be extremely useful!

I've had an initial run through and it has worked as described in the description. I also think the init container architecture sounds good, so concept ack on that.

Few initial thoughts before I dive into code review:

- This really needs a new doc. I wouldn't have known to run a command like `warnet snapshot miner -f "blocks,chainstate,wallets"` I know you have that on TODO.
- Snapshots should perhaps be saved in snapshots directory rather than root of current directory?

josibake:
> Snapshots should perhaps be saved in snapshots directory rather than root of current directory?

agree. there is an `--output` flag for the snapshot command, but perhaps better would be to autocreate an empty snapshots directory via `warnet init` and have the default be there? the problem I see here is that `warnet` commands can be ran from anywhere , so if you run a warnet command from outside the project directory, and it tries to default to snapshots, the command will fail. perhaps a work around for now would be always require the `--output` argument so there are no surprises for the user?

m3dwards:
Or create snapshot directory should one not exist?

m3dwards:
Code looks ok to me, fancy adding a little doc and we can get this merged?

m3dwards (approved):

josibake:
> Or create snapshot directory should one not exist?

could do, but generally not a fan of creating dirs randomly on the users system, outside of a scoped warnet project directory. I think for now I'll leave it to "output dir is required" and then as we fine tune this whole project structure / snapshot management we can improve on it.

> Code looks ok to me, fancy adding a little doc and we can get this merged?

will do!

josibake:
@m3dwards added a doc, and also took your suggestion of defaulting to `warnet-snapshots` and creating the dir if the user doesn't provide one. I chose `warnet-snapshots` because it's quite annoying to type `warnet snapshot` and have your terminal try to autocomplete it to the directory `snapshots`. Probably a more clever fix for this , but this works for now.

m3dwards:
LGTM. Like the doc!

josibake:
thx! working on a test locally but its a bit tricky given interactions with local fs etc, should have it finished tomorrow and will open a PR

willcl-ark:
This has broken apidocs https://github.com/bitcoin-dev-project/warnet/actions/runs/10816493180/job/30007779318

pinheadmz:
@willcl-ark fixed #585

PR #548, opened 2024-09-07
check if docker and minikube are running

mplsgrant:
It might help with #542 if quickstart checked if docker and minikube were running.

PR #556, opened 2024-09-09
setup: add "No Backend" option

pinheadmz:
See: https://github.com/bitcoin-dev-project/warnet/pull/548#issuecomment-2336658681

If a user is just playing a wargame on a remote cluster that is already deployed, they don't need docker, docker desktop, or minikube locally. They will still need kubectl and helm.

### DIFF
--- a/src/warnet/project.py
+++ b/src/warnet/project.py
@@ -212,7 +212,7 @@ def setup():
                 choices=[
                     "Minikube",
                     "Docker Desktop",
-                    "No Backend (Interacting with remote cluster already deployed)",
+                    "No Backend (Interacting with remote cluster, see `warnet auth --help`)",
                 ],
             )
         ]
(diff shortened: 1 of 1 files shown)


src/warnet/project.py
```diff
@@ -209,7 +209,11 @@ def check_installation(tool_info: ToolInfo) -> ToolStatus:
             inquirer.List(
                 "platform",
                 message=click.style("Which platform would you like to use?", fg="blue", bold=True),
-                choices=["Minikube", "Docker Desktop"],
+                choices=[
+                    "Minikube",
+                    "Docker Desktop",
+                    "No Backend (Interacting with remote cluster already deployed)",
```
josibake: Might be worth prompting, or at least hinting to the user they will need to run `warnet auth <kubeconfig for remote cluster>` if choosing the "No Backend" option. Otherwise, they might finish the setup and then nothing works. The auth command is also nice in that it makes sure that kubectl is switched over to use the "warnet" context.

josibake (commented):
nice, looks good. non-blocking comment about prompting the user to use `warnet auth`

src/warnet/project.py
willcl-ark: Could even run it for them?

pinheadmz: i like it, added

PR #560, opened 2024-09-09
Fixup signet test

willcl-ark:
@pinheadmz was there a reason to intentially disable Ruff on this file? It appears to complete successfully for me after applying the suggested fixes.

### DIFF
--- a/resources/scenarios/signet_miner.py
+++ b/resources/scenarios/signet_miner.py
@@ -22,22 +22,44 @@ import logging
 import math
 import re
 import struct
+import subprocess
 import sys
 import time
-import subprocess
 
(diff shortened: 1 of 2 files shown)


pinheadmz:
Because it's 99% copied from bitcoin core so just trying to keep the diff minimal like the test framework

willcl-ark:
Ah Ok. Well then I don't know if this is needed, but the changes make the file nicer IMO. Happy to close though

josibake:
personally, I think we should trim down the signet miner to just what we need vs the whole kitten caboodle from bitcoin core, but as long as we are just copying from core I think its better to keep this the same as core, so closing this for now.

PR #566, opened 2024-09-10
Install logging as part of deploy

### LINKED ISSUE #526: install/connect logging should be warnet CLI commands
Because our new model is: 

`pip install warnet`
`cd some directory`
`warnet create ...`
and so forth.

So the user shouldn't need to know where on their computer the `install_logging.sh` script is

m3dwards:
Fixes: #526


josibake:
ACK, lgtm!

PR #572, opened 2024-09-10
bitcoincore: lazier probe checks to accomdate slower nodes / systems

### LINKED ISSUE #564: 0.16.1 does not support arm64 (macOS)
We introduced support for macOS after 16.1, so we will need to back port the build system changes to the 16.1 image if we want to let people use that image on macOS (e.g., docker desktop, Minikube).

I _think_ this should be possible by just back porting the build system changes? But not a priority for right now since everything at Tabconf can be ran inside a K8s cluster (linux).

pinheadmz:
closes #564 

See my comment in that issue, all the node versions we support DO run on fine on arm64/macos some of them just need a few more seconds before they can pass our live/ready checks

### DIFF
--- a/resources/charts/bitcoincore/values.yaml
+++ b/resources/charts/bitcoincore/values.yaml
@@ -78,16 +78,16 @@ livenessProbe:
     command:
     - pidof
     - bitcoind
-  failureThreshold: 3
+  failureThreshold: 12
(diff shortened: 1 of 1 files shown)


willcl-ark:
ACK

josibake:
ACK

PR #574, opened 2024-09-10
Updated install docs with guidance about testing minikube

### LINKED ISSUE #570: Docs: On the dependencies for install write a section that proves that minikube is running
Perhaps the docs should run a hello world minikube pod to prove it's up and running.

m3dwards:
I think these docs are clearer on what you have to do. I also got rid of the quickstart document as it's duplicated in the `install.md` doc.

Fixes: https://github.com/bitcoin-dev-project/warnet/issues/570


josibake:
LGTM, much cleaner! Prompting people to go to docker desktop / minikube docs is excellent. I recently suggested we might need a `CONTRIBUTING.md` doc in the top level of the repo, which would conflict with your naming here. Perhaps we can call this `dependencies.md` / `installing.md` ?

josibake:
(can also merge now and change the name later if / when we do add a contributing.md)

m3dwards:
> I recently suggested we might need a `CONTRIBUTING.md` doc in the top level of the repo

I'll just move / rename it, probably would need to be filled out a bit more later too.

PR #576, opened 2024-09-10
rpc: add regex pattern filtering to `warnet logs` rpc

ismaelsadeeq:
This PR updated the `warnet logs` rpc to now have an optional grep option which will
take a string value and filter tank logs based on regex matching with the passed string.

This can be tested by

`warnet log tank-001 --grep "pattern"`

```terminal
Usage: warnet logs [OPTIONS] [POD_NAME]

  Show the logs of a Kubernetes pod, optionally filtering by a grep pattern.

Options:
  -f, --follow     Follow logs in real-time
  -g, --grep TEXT  Pattern to grep for in logs
  --help           Show this message and exit.
```

### DIFF
--- a/src/warnet/control.py
+++ b/src/warnet/control.py
@@ -231,17 +231,18 @@ def run(scenario_file: str, additional_args: tuple[str]):
 
 
 @click.command()
-@click.argument("pod_name", type=str, default="")
-@click.option("--follow", "-f", is_flag=True, default=False, help="Follow logs")
-def logs(pod_name: str, follow: bool):
-    """Show the logs of a pod"""
+@click.argument("pod_name", type=str, required=False)
+@click.option("--follow", "-f", is_flag=True, default=False, help="Follow logs in real-time")
+@click.option("--grep", "-g", type=str, required=False, help="Pattern to grep for in logs")
+def logs(pod_name: str = "", follow: bool = False, grep: str = ""):
+    """Show the logs of a Kubernetes pod, optionally filtering by a grep pattern."""
     follow_flag = "--follow" if follow else ""
     namespace = get_default_namespace()
 
     if pod_name:
         try:
             command = f"kubectl logs pod/{pod_name} -n {namespace} {follow_flag}"
-            stream_command(command)
+            stream_command(command, grep)
             return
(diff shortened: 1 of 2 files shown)


willcl-ark:
Thanks! This is a really nice idea and looks like it works well already.

I think the implementation could be even more awesome if it worked a little more like the `logs` command: https://github.com/bitcoin-dev-project/warnet/blob/71cbcec8f3d4ea4b7cf90308795ffd8bf0a506e5/src/warnet/control.py#L249-L293

This gives you a "selector" to choose a running tank from. If you did fancy taking this approach, perhaps `--tank` could be a flag (or could rename `--select` or something) which would start the tank selector menu. The selected tank could then be the only tank in the `tanks` variable and no modification to get_mission would be needed (I think).

josibake:
Hm, currently we have `warnet logs` , which allows you to get the logs from a scenario container or a bitcoin tank. I'd prefer if we used that exclusively for accessing logs, instead of having one way to get logs from a commander, one way to get logs from a bitcoin node, and in the future when we add LN back in, a different way to get logs from a LN node.

@ismaelsadeeq any interest in adding a `--grep` argument to the `warnet logs` command? I think this would solve your use case and avoid you needing to reimplement the selector logic here.

We might also want to remove grep-logs entirely in the future in favor of `warnet logs --all --grep "my pattern"` which would stream logs from all containers and allow the user to grep for a pattern

ismaelsadeeq:
Thank you @willcl-ark @josibake for your reviews.

I will get back to this soon.

ismaelsadeeq (commented):
I have not encountered logs cli prior to your comment @josibake.
What I would like to do can even simply be done using logs rpc by piping the result of logs with grep command.

But I believe the approach you mentioned is cleaner and faster hence I updated the PR to your suggestion.

I force-pushed to implement it, and updated the PR description and title to match what the C.L does.

I think adding `--all` option to `logs` rpc, and getting rid of `bitcoin grep-logs` can be added in a follow-up after this.

ismaelsadeeq:
Will rebase this if there is need for this.
 Else happy to close if it is no longer relevant.

Issue #549, opened 2024-09-08
self.nodes[] in scenarios should accept node names


pinheadmz:
So it will match the graph specified in network.yaml

We may be able to preserve backwards compatibility by making self.nodes a dict and indexing nodes by *both* their index number and their name?

josibake:
+1, I found it pretty confusing to be specifying tank names in the config/logs/bitcoin rpc commands, but then needing to switch to index when using scenarios and tests.

could backwards compatible, but I generally prefer having one way to do things.

PR #577, opened 2024-09-10
scenarios: add recon scenario

pinheadmz:
Adds a scenario with verbose comments that illustrates how to probe the network for vulnerable nodes and use the P2PConnection in python to serve them... unusual... messages

### DIFF
--- a/resources/scenarios/reconnaissance.py
+++ b/resources/scenarios/reconnaissance.py
@@ -52,7 +52,10 @@ class Reconnaissance(Commander):
             dstaddr, dstport = victim["addr"].split(":")
         else:
             dstaddr = socket.gethostbyname(victim["addr"])
-            dstport = 18444
+            if self.chain == "regtest":
+                dstport = 18444
+            if self.chain == "signet":
+                dstport = 38333
 
         # Now we will use a python-based Bitcoin p2p node to send very specific,
         # unusual or non-standard messages to a "victim" node.
--- a/resources/scenarios/reconnaissance.py
+++ b/resources/scenarios/reconnaissance.py
@@ -6,12 +6,12 @@ import socket
 # but requires a relative path inside the python source code for other functions.
 try:
     from commander import Commander
-except Exception:
+except ImportError:
     from resources.scenarios.commander import Commander
 
 # The entire Bitcoin Core test_framework directory is available as a library
-from test_framework.messages import MSG_TX, CInv, msg_getdata, hash256
-from test_framework.p2p import P2PInterface, MAGIC_BYTES
+from test_framework.messages import MSG_TX, CInv, hash256, msg_getdata
+from test_framework.p2p import MAGIC_BYTES, P2PInterface
 
 
 # This message is provided to the user when they describe the scenario
--- a/resources/scenarios/reconnaissance.py
+++ b/resources/scenarios/reconnaissance.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+
+import socket
+
+# The base class exists inside the commander container when deployed,
+# but requires a relative path inside the python source code for other functions.
+try:
+    from commander import Commander
+except Exception:
+    from resources.scenarios.commander import Commander
+
+# The entire Bitcoin Core test_framework directory is available as a library
+from test_framework.messages import MSG_TX, CInv, msg_getdata
+from test_framework.p2p import P2PInterface
+
+
+# This message is provided to the user when they describe the scenario
+def cli_help():
+    return "Demonstrate network reconnaissance using a scenario and P2PInterface"
+
+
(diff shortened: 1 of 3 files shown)


resources/scenarios/reconnaissance.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+
+import socket
+
+# The base class exists inside the commander container when deployed,
+# but requires a relative path inside the python source code for other functions.
+try:
+    from commander import Commander
+except Exception:
```
willcl-ark: I think I added this originally, but prob best to do `except ImportError` here?

resources/scenarios/reconnaissance.py
```diff
@@ -0,0 +1,72 @@
...
+
+# The actual scenario is a class like a Bitcoin Core functional test.
+# Commander is a subclass of BitcoinTestFramework instide Warnet
+# that allows to operate on containerized nodes instead of local nodes.
+class Reconnaissance(Commander):
+    def set_test_params(self):
+        # This setting is ignored but still required as
+        # a sub-class of BitcoinTestFramework
+        self.num_nodes = 1
+
```
willcl-ark, line 39: IMO we should include adding an option here to show how it's done:

```python
    def add_options(self, parser):
        parser.add_argument(
            "--allnodes",
            dest="allnodes",
            action="store_true",
            help="When true, generate blocks from all nodes instead of just nodes[0]",
        )
```

resources/scenarios/reconnaissance.py
```diff
@@ -0,0 +1,72 @@
...
+    # Scenario entrypoint
+    def run_test(self):
+        self.log.info("Getting peer info")
+
+        # Just like a typical Bitcoin Core functional test, this executes an
+        # RPC on a node in the network. The actual node at self.nodes[0] may
+        # be different depending on the user deploying the scenario. Users in
+        # Warnet may have different namepsace access but everyone should always
+        # have access to at least one node.
```
willcl-ark, line 48: Do we have any idea how they access nodes in their namespace? Do they all start at `0` for everyone, is are they going to be:

0 - 5
5 - 10 
etc.?

would be good to be able to demonstrate how a user could find all node names that they can control

willcl-ark (commented):
Nice work, I like the idea. Left a few thoughts inline.

resources/scenarios/reconnaissance.py
josibake: one idea here is having a helper utility for allowing the user to easily import the `network.yaml` file into some sort of python config object. this wouldn't have information about other commanders, but it would allow the users to be aware of everything that is deployed in a particular network ?

resources/scenarios/reconnaissance.py
pinheadmz: holding off on this until #549

resources/scenarios/reconnaissance.py
pinheadmz: there will always be a self.nodes[0]. Commander fills that list with info from the warnet.json file which is generated by whatever the user gets from `describe pods`

pinheadmz:
pushed an update that uses rpc `getblocktemplate` to determine the private signet magic bytes for the P2PConnection. That will be pretty important for tabconf! The scenario is tested on both regtest and signet in CI now as well

resources/scenarios/reconnaissance.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+# This message is provided to the user when they describe the scenario
+def cli_help():
+    return "Demonstrate network reconnaissance using a scenario and P2PInterface"
+
+
+def get_signet_network_magic_from_node(node):
+    template = node.getblocktemplate({"rules": ["segwit", "signet"]})
+    challenge = template["signet_challenge"]
```
josibake, line 24: ha, clever!

resources/scenarios/reconnaissance.py
```diff
@@ -0,0 +1,90 @@
...
+    from resources.scenarios.commander import Commander
+
+# The entire Bitcoin Core test_framework directory is available as a library
+from test_framework.messages import MSG_TX, CInv, hash256, msg_getdata
+from test_framework.p2p import MAGIC_BYTES, P2PInterface
+
+
+# This message is provided to the user when they describe the scenario
+def cli_help():
+    return "Demonstrate network reconnaissance using a scenario and P2PInterface"
```
josibake, line 19: this doesn't do anything , fyi. I haven't looked at it yet, but I think we can add help text to the scenario via something like `add_options`? (not a blocker for this PR)

josibake (approved):
nice, LGTM. few nits, but nothing really specific to this PR so lets merge and we can address stuff in follow ups.

PR #578, opened 2024-09-10
Design doc cont

willcl-ark:
Picks up #547 while @mplsgrant is away.

- [x] Add extra design and philosophy info, along with background on why helm and k8s
- [x] Add pretty logos to break up the page a bit
- [x] Fixup the config graph which is not rendering properly and now linked to from DESIGN.md
- [x] Link to DESIGN.md from README.md?

Let me know if you think the extra Helm & K8s info in design.md is too much, and I can remove it. I think it might help people like me, that don't know much about k8s and helm though, without a "just go read their docs" approach.

ping @adamjonas in case you fancy a review too


DESIGN.md
```diff
@@ -0,0 +1,169 @@
+# Overview
+
+Warnet is designed for Bitcoin developers, researchers, and enthusiasts who want to safely experiment with and test ideas related to Bitcoin networks. By leveraging Kubernetes, Warnet enables users to deploy and simulate large-scale Bitcoin networks in a controlled environment.
```
josibake: nit: might be worth mentioning reproducibility, as this is a major selling point of warnet IMO:

```suggestion
Warnet is designed for Bitcoin developers, researchers, and enthusiasts who want to safely experiment with and test ideas related to Bitcoin networks. By leveraging Kubernetes, Warnet enables users to deploy and simulate large-scale Bitcoin networks in a controlled, reproducible environment.

**Review by josibake — COMMENTED** [review]
nice work , @mplsgrant and @willcl-ark - this is an excellent doc! I like that the reasoning for k8s + helm is detailed along with an opinion about how things should be done.

I think the overview of the python files is helpful, but a bit worried that it's over specified and will lead to doc-rot if not careful. That being said, I did find it helpful and imagine it will be useful for new contributors, as well.

Will leave time for others to chime in , but overall LGTM!

File: DESIGN.md (hunk 1 above)
Comment (willcl-ark) [reply DESIGN.md]: Takien in [a63e29c](https://github.com/bitcoin-dev-project/warnet/pull/578/commits/a63e29c7f6c60da22910efe4262c340a48f7e0a0)

**Comment by willcl-ark** [comment]
> I think the overview of the python files is helpful, but a bit worried that it's over specified and will lead to doc-rot if not careful. That being said, I did find it helpful and imagine it will be useful for new contributors, as well.

Yeah I was going to use auth as an example, but thought it would be more likely someone wanted to add a new service.

File: DESIGN.md [hunk 2]
```diff
@@ -0,0 +1,169 @@
...
+5. [Operating in the Network with "Scenarios"/Commanders](#operating-in-the-network-with-scenarioscommanders)
+
+6. [The Resources Configuration Pipeline - An Example](#the-resources-configuration-pipeline---an-example)
+
+## Key Features:
+- Safe simulation of Bitcoin networks
+- Kubernetes-based deployment for scalability and ease of management
+- User interaction capabilities through custom scenarios
+- Infrastructure as Code (IaC) approach for reproducibility
+- Stateless Configuration and Externalized Configuration for flexibility
```
m3dwards: I'm not really sure what stateless and externalised configuration is? Is there another way to describe this?

DESIGN.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Target Audience:
+- Bitcoin core developers
+- Bitcoin researchers
+- Network security specialists
+- Bitcoin protocol testers
```
m3dwards, line 38: I'd like to hope the target audience is much broader than this including:

Bitcoin application developers
Lightning developers
Students of bitcoin

DESIGN.md
```diff
@@ -0,0 +1,169 @@
...
+- Scalability: Easily scale the number of simulated nodes
+- Resource management: Efficiently allocate computational resources
+- Service discovery: Automatically manage network connections between nodes
+
+### Helm
+[Helm](https://helm.sh/) is a package manager for Kubernetes that simplifies the process of defining, installing, and upgrading complex Kubernetes applications. Warnet prefers Helm for managing Kubernetes resources due to its powerful templating and package management capabilities.
+
+Advantages of using Helm in Warnet:
+- Templating: Define reusable Kubernetes resource templates
+- Versioning: Manage different versions of deployed resources
```
m3dwards: We don't currently use helm versioning and may choose to never as we include the chart as part of our source code.

DESIGN.md
```diff
@@ -0,0 +1,169 @@
...
+- Packaging: Bundle related Kubernetes resources into a single unit (chart)
+- Simplified deployment: Use a single command to deploy complex applications
+
+### Why Helm is Preferred in Warnet
+1. Reproducibility: Helm charts ensure consistent deployments across different environments.
```
m3dwards, line 101: I think this one might be debatable as wouldn't we get reproducibility writing yaml manually with python?

DESIGN.md
```diff
@@ -0,0 +1,169 @@
...
+
+Looking more closely at the *resources* section, for example, we can focus in on the *namespaces* directory. Inside, there is an example *namespaces.yaml* and *namespace-defaults.yaml* file. These configuration files are provided to the user when the `warnet admin init` command is invoked. This provides the user the opportunity to change those configuration files by specifying a set of participants who will have access to the simulated Bitcoin network. When `warnet deploy [namespaces_folder]` command is run by the user, it will apply the configuration data to the Helm chart found in the *charts* directory of the *resources* section. The Helm chart acts as a template through which the user's configuration data is applied. In this way, there is a pipeline which starts with the user's Stateful Data which is then piped through the Helm templating system, and then is applied to the Kubernetes cluster.
```
m3dwards: Worth mentioning that these are under the charts directory? Would bitcoincore be the better example here? It's more core and will always be used whereas namespaces might not?

m3dwards (approved):
Lovely doc, really gives a great feel on the design philosophy.

Left a bunch of nits but none are important and can be ignored.

DESIGN.md
willcl-ark: stateless is like a Helm deployment no? i.e. you set the config you want and `upgrade --install` and it tries to reach that config for you?

I think I will take it out though...

DESIGN.md
willcl-ark: Yeah I was kinda transposing from a "helm advantages list" a bit loosely here

DESIGN.md
willcl-ark: I can remove, although I think we could make lots of things reproducible, perhaps even bash scripts!

This does make me think though, perhaps I've missed the primary helm advantage, in the way that is merges configurations together. Is there a technical term for this do you know?

m3dwards: Can't think of a technical term but for me helm is great for its, templating, configuration inheritance, encapsulation / modularity.

PR #579, opened 2024-09-11
bitcoin: save debug log to normal file

willcl-ark:
This is (erronerously?) set to `0` currently which is odd. Set to debug.log.

I seem to recall this was an attempt to disable debug.log back from the docker days (stdout only), but  even in that case we should have used `nodebuglogfile`.


josibake:
VINDICATED

PR #581, opened 2024-09-11
Project.py messages clarification

hodlinator:
Minor messaging adjustments to increase clarity, especially for newcomers.


josibake:
ACK, thanks for the improvements!

PR #582, opened 2024-09-11
setup_user_contexts.sh - Halt if kubectl fails

hodlinator:
Script would otherwise lie and state that "All kubeconfig files have been created".


josibake:
FYI , this script is a temporary work around that is removed in https://github.com/bitcoin-dev-project/warnet/pull/545 (altho I did notice the PR doesn't actually delete the script, which it should!)

josibake:
Closing since this file will be removed in #545

PR #590, opened 2024-09-11
Add nix development shell with a flake

nyonson:
Not sure if any nix fans here, but I found this helpful when bouncing between projects and needing to re-enable the dev shell.

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,3 +1,3 @@
 # Community Contributions
 
-`Contains community-contributed resources for Warnet. These contributions are not officially supported or maintained.
+Contains community-contributed resources for Warnet. These contributions are not officially supported or maintained.
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,0 +1,3 @@
+# Community Contributions
+
+`Contains community-contributed resources for Warnet. These contributions are not officially supported or maintained.
--- a/contrib/nix/flake.lock
+++ b/contrib/nix/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1725983898,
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,42 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+      in
+      {
+        devShells.default = pkgs.mkShell {
```
hodlinator: Maybe enough to do this since there aren't multiple?
```suggestion
        devShell = pkgs.mkShell {
```

hodlinator (approved):
ACK c82a861f52b01770106de87f1c3a85f26d34d6a2

Also on NixOS here. Slicker than the shell I shared earlier, and flake-ified, nice!

Can understand if non-nixers would rather keep system-specific file out of the project though. Despite there being an entire base58 conference on nix...

Not 100% sure about committing the *flake.lock* file, but probably good to have confirmed working versions checked in.

Edit: Could be good to reduce friction for *newcomers* who happen to use nix. Might provide an unfair advantage in wargames though. :)

flake.nix
nyonson: I was reading how the implicit version is still here just for backwards compatibility. I wouldn't mind switching to it here since like you said, only one, but figured I'd get in the habit of being explicit.

hodlinator: That's a comment regarding the .lock-file, right, not the shell?

josibake:
Hey @nyonson , thanks for the contribution! Currently, the main contributors to this project don't have much (_any_ ?) experience with Nix, so I'm hesitant to add support for it only because we won't be able to provide support. I'd be more inclined if something like this can be added in a `contrib/` directory, along with a README explaining these are community contributions and not officially supported. I'm not sure if that's feasible for something like a nix-flake, or if it has to be in the root of the directory?

flake.nix
nyonson: Ah sorry, I meant implicit default devShell, without the "default" key.

For the lock file, I was thinking the same as you that it might help avoid slightly different local versions.

nyonson:
> Hey @nyonson , thanks for the contribution! Currently, the main contributors to this project don't have much (_any_ ?) experience with Nix, so I'm hesitant to add support for it only because we won't be able to provide support. I'd be more inclined if something like this can be added in a `contrib/` directory, along with a README explaining these are community contributions and not officially supported. I'm not sure if that's feasible for something like a nix-flake, or if it has to be in the root of the directory?

Totally, I'll refactor this pr to use a contrib dir.

contrib/README.md
```diff
@@ -0,0 +1,3 @@
+# Community Contributions
+
+`Contains community-contributed resources for Warnet. These contributions are not officially supported or maintained.
```
josibake: typo:

```suggestion
Contains community-contributed resources for Warnet. These contributions are not officially supported or maintained.
```

josibake (commented):
Nice, looks good mod a small typo!

PR #592, opened 2024-09-12
add podname-indexed dict "self.tanks" to Commander

### LINKED ISSUE #549: self.nodes[] in scenarios should accept node names

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/549

This leaves `self.nodes[int]` alone, so scenarios where the pod names are unknown still have a chance of working. For example `self.nodes[0]` should always be available if you just need *some* node to do something.

We add `self.tanks[str]` which returns the same `TestNode` object but indexed by pod name. Covered in scenarios run by ci


willcl-ark:
Our CI did not automagically update the commander image which makes all tests fail outside the CI

willcl-ark:
We should prob test PRs at least once before merging :)

willcl-ark:
Addressed in https://github.com/bitcoin-dev-project/warnet/pull/597

edit: this does fix it, just needed to force-pull new image from registry

pinheadmz:
> We should prob test PRs at least once before merging :)

I tested locally by building the commander image locally! Just like CI does.

PR #594, opened 2024-09-12
Deploy default grafana dashboard automatically

pinheadmz:
This adds a new `chart` that is just a ConfigMap. When deployed it will pull all json files from `resources/charts/grafana/files` and install them as dashboards in Grafana. Also removed all traces of `connect_logging.sh` from docs and the logging test, now that port forwarding is built in.

Screenshot 2024-09-12 at 3 39 15 PM


m3dwards:
I like this and I don't think it's a bad approach. I think however there should be some concept of installing dashboards for your network which could be done as a follow up PR perhaps.

m3dwards (approved):

m3dwards:
Worked for me and I added a second file in there which also got deployed automatically which is nice. Merging.

Issue #383, opened 2024-06-18
Add method / test / docs for "Armada" deployment


pinheadmz:
See https://github.com/carlaKC/attackathon/pull/61

I think the idea of deploying attacks to an existing network could be one of our main features, and we should at least have an integration test to cover the issue Carla had.

PR #595, opened 2024-09-12
binary file via init

willcl-ark:
probably should launch a commander like this too really...

### DIFF
--- a/docs/warnet.md
+++ b/docs/warnet.md
@@ -73,6 +73,16 @@ options:
 | scenario_file   | Path   | yes        |           |
 | additional_args | String |            |           |
 
+### `warnet run-binary`
+Run a file in warnet
+    Pass `-- --help` to get individual scenario help
+
+options:
+| name            | type   | required   | default   |
+|-----------------|--------|------------|-----------|
+| file            | Path   | yes        |           |
+| additional_args | String |            |           |
+
 ### `warnet setup`
 Setup warnet
 
--- a/resources/charts/binary-runner/templates/pod.yaml
+++ b/resources/charts/binary-runner/templates/pod.yaml
@@ -1,0 +1,31 @@
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ include "binary-runner.fullname" . }}
+  labels:
+    {{- include "binary-runner.labels" . | nindent 4 }}
+    app: {{ include "binary-runner.name" . }}
+    mission: binary
+spec:
+  restartPolicy: Never
+  volumes:
+    - name: shared-data
+      emptyDir: {}
+  containers:
+    - name: {{ .Values.pod.name }}-runner
+      image: alpine
+      command: ["/bin/sh", "-c"]
+      args:
+        - |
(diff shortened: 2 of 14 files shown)


pinheadmz:
Yes nice approach. A solid step to close #383 as well. 

Dunno if we need to keep binaries and scenarios separate for sake of Warnet status or "missions"

don't forget to run test in ci by adding to test.yml !

pinheadmz:
Also what if user wants to run multiple commands on a special binary? Like, deploy an application along with a script that runs it, or any case with multiple files ?

willcl-ark:
> don't forget to run test in ci by adding to test.yml !

whoops! thanks

> Also what if user wants to run multiple commands on a special binary? Like, deploy an application along with a script that runs it, or any case with multiple files ?

They will have to build that themselves. Prob easiest to just write a dockerfile at that stage?

docs/warnet.md
```diff
@@ -65,14 +65,24 @@ options:
...
 
 options:
 | name            | type   | required   | default   |
 |-----------------|--------|------------|-----------|
 | scenario_file   | Path   | yes        |           |
 | additional_args | String |            |           |
 
+### `warnet run-binary`
+Run a file in warnet
+    Pass `-- --help` to get individual scenario help
```
pinheadmz, line 78: `-- --help` is specific to Commander/test framework, might not apply to all binaries

josibake:
BIG Concept ACK, light code review ack (haven't tested).

Really like the initContainers approach , as this gives us a ton of flexibility in the future without needing to continually re-write our python code. It also gives a single place to manage things like putting the files in the correct directory, set the write permissions (as you're currently doing), and more all in a self contained unit that cleans up after itself. This allows the resulting image to be clean and just focus on "running the thing." Also very happy to see we are not base64'ing files into `etcd` just to get them into a container. This should allow users to package arbitrarily large binaries without hitting configMap limits, or slowing down `etcd`.

resources/charts/binary-runner/templates/pod.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - name: shared-data
+      emptyDir: {}
+  initContainers:
+    - name: {{ .Values.pod.name }}-receiver
+      image: alpine
+      command: ["/bin/sh", "-c"]
+      args:
+        - |
+          echo "Waiting for binary file..."
+          while [ ! -f /data/binary ]; do
```
m3dwards, line 22: If this wait was moved into the main container we wouldn't need the initContainer?

Is there another advantage to the init container?

willcl-ark: There is no other advantage, nice spot. This is just an artefact of me trying to migrate from `configMap` and, instead of taking two steps forward, only taking one.

Removed the init container in [effe95d](https://github.com/bitcoin-dev-project/warnet/pull/595/commits/effe95d103c69eb6e52dfca667f20840754432fc)

m3dwards:
Ruff failure but otherwise looks good.

PR #597, opened 2024-09-13
ci: rebuild commander on python changes too

willcl-ark:
`paths:` specify on which files (changed) this workflow should run.

We want it to run on pushes to main, when these two files change.


josibake:
ACK

Issue #603, opened 2024-09-15
bug: warnet barfs a stack trace if running `warnet status` with an outdated kubeconfig


josibake:
Installed warnet, created a new project directory and ran `warnet status` without running `minikube start` or checking my cluster configs in `kubectl` and got this:

```
Traceback (most recent call last):
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connection.py", line 199, in _new_conn
    sock = connection.create_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/util/connection.py", line 85, in create_connection
    raise err
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/util/connection.py", line 73, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 61] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 789, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 490, in _make_request
    raise new_e
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 466, in _make_request
    self._validate_conn(conn)
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 1095, in _validate_conn
    conn.connect()
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connection.py", line 693, in connect
    self.sock = sock = self._new_conn()
                       ^^^^^^^^^^^^^^^^
  File "/Users/josibake/work/warnet-projects/tabconf-example/.venv/lib/python3.12/site-packages/urllib3/connection.py", line 214, in _new_conn
    raise NewConnectionError(
[...]

PR #604, opened 2024-09-16
Add `warnet status` exception handling

mplsgrant:
Add some status exception handling as per #603


willcl-ark (approved):
Looks good to me

Issue #609, opened 2024-09-17
setup: replace kubectl with kubernetes python library


pinheadmz:
...and remove kubectl from depenedencies

PR #607, opened 2024-09-17
Add ingress

m3dwards:
To test:

```shell
warnet deploy ./resources/networks/6_node_bitcoin
warnet dashboard
```

Tested on Minikube, Docker Desktop and Google so especially looking for Digital Ocean users.

PR #614, opened 2024-09-19
swap out kubectl

mplsgrant:
Took a stab at swapping out kubectl addressing #609

* It appears to *mostly* work; however, something is up with the signet test, and I didn't have a chance to figure out what.

* There was also an oddball issues around `stream` returning python dicts and lists instead of json. I was able to avoid the impacts of this issue.

* Some things remain less tested such as a few namespace items and the chain save-state functionality.

### DIFF
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -1,3 +1,6 @@
+import ast
+import base64
+import json
 import os
 import re
 import sys
@@ -5,12 +8,14 @@ from datetime import datetime
 from io import BytesIO
 
 import click
+from kubernetes.client.models import V1Pod, V1Service
+from kubernetes.stream import stream
 from urllib3.exceptions import MaxRetryError
 
 from test_framework.messages import ser_uint256
 from test_framework.p2p import MESSAGEMAP
 
-from .k8s import get_default_namespace, get_mission
+from .k8s import get_default_namespace, get_mission, get_static_client, kexec
 from .process import run_command
 
 
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -89,6 +89,7 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
     """
     Grep combined bitcoind logs using regex <pattern>
     """
+    sclient = get_static_client()
 
     try:
         tanks = get_mission("tank")
@@ -111,8 +112,12 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
             continue
 
         # Get logs from the specific container
-        command = f"kubectl logs {pod_name} -c {container_name} --timestamps"
-        logs = run_command(command)
+        logs = sclient.read_namespaced_pod_log(
+            name=pod_name,
+            namespace=get_default_namespace(),
+            container=container_name,
+            timestamps=True,
+        )
 
         if logs is not False:
             # Process logs
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -1,5 +1,6 @@
 import ast
 import base64
+import codecs
 import json
 import os
 import re
@@ -15,7 +16,7 @@ from urllib3.exceptions import MaxRetryError
 from test_framework.messages import ser_uint256
 from test_framework.p2p import MESSAGEMAP
 
-from .k8s import get_default_namespace, get_mission, get_static_client, kexec
+from .k8s import get_default_namespace, get_mission, get_static_client, kexec, pod_log
 from .process import run_command
 
 
@@ -94,15 +95,13 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
     """
     Grep combined bitcoind logs using regex <pattern>
     """
-    sclient = get_static_client()
-
     try:
         tanks = get_mission("tank")
     except MaxRetryError as e:
         print(f"{e}")
         sys.exit(1)
 
-    matching_logs = []
+    matching_logs: list[tuple[str, any]] = []
 
     for tank in tanks:
         pod_name = tank.metadata.name
@@ -117,18 +116,14 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
             continue
 
         # Get logs from the specific container
-        logs = sclient.read_namespaced_pod_log(
-            name=pod_name,
-            namespace=get_default_namespace(),
-            container=container_name,
-            timestamps=True,
-        )
+        log_stream = pod_log(pod_name, container_name, timestamps=True)
 
-        if logs is not False:
-            # Process logs
-            for log_entry in logs.splitlines():
-                if re.search(pattern, log_entry):
-                    matching_logs.append((log_entry, pod_name))
+        compiled_pattern = re.compile(pattern)
+
+        for log_line in iter_lines_from_stream(log_stream):
+            log_entry = log_line.rstrip("\n")
+            if compiled_pattern.search(log_entry):
+                matching_logs.append((log_entry, pod_name))
 
     # Sort logs if needed
     if not no_sort:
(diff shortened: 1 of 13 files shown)


willcl-ark:
Am I correct in thinking that all users may still require `kubectl` to be installed, so that the cluster can be configured? Or does the python lib bypass all that too? 

I had always assumed it hinged on `~/.kube/config` to know *where* to do the things it was going to do, but never actually checked it out. Checking the code it seems like there could be alternatives to `load_kube_config()` (in fact, I see [here](https://github.com/kubernetes-client/python/blob/eeeb43cce3c98eb10ae41967138a6d76dbe67ac1/kubernetes/base/config/kube_config.py#L251) references to google auth, which sounds potentially useful to us on the credential side of things?)

Users may also want `kubectl` for any additional k8s applications, such as `k9s`, to work, unless they too can be tricked by dumping a python-generated config (from this `kubernetes` lib) into the user's *~.kube/config* file ;)

If pure python is preferred, then this looks like a big move in the right direction!

src/warnet/bitcoin.py
```diff
@@ -35,15 +40,37 @@ def rpc(tank: str, method: str, params: str):
...
+        stdin=False,
+        stdout=True,
+        tty=False,
+    )
+    # The k8s lib seems to convert json into its python representation. The following dance seems to
+    # avoid the worst of it.
+    if resp.startswith("{") or resp.startswith("["):
+        literal = ast.literal_eval(resp)
+        json_string = json.dumps(literal)
+        return json_string
```
pinheadmz: the old `exec_run()` function added this option to `stream()` for this: ` _preload_content=False`

overall i think its better to restore the old `exec_run()` function in k8s.py from pre-italy and just call it here.
That's as far as I got on this project: 756df48b97acb1496b2958e12b895a00

src/warnet/bitcoin.py
```diff
@@ -89,8 +117,12 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
...
 
         # Get logs from the specific container
-        command = f"kubectl logs {pod_name} -c {container_name} --timestamps"
-        logs = run_command(command)
+        logs = sclient.read_namespaced_pod_log(
+            name=pod_name,
+            namespace=get_default_namespace(),
+            container=container_name,
+            timestamps=True,
+        )
```
pinheadmz: can also use the `_log()` internal function I added to get the log

src/warnet/bitcoin.py
```diff
@@ -163,21 +195,26 @@ def get_messages(tank_a: str, tank_b: str, chain: str):
...
     """
+    sclient = get_static_client()
+
     subdir = "" if chain == "main" else f"{chain}/"
     base_dir = f"/root/.bitcoin/{subdir}message_capture"
 
     # Get the IP of node_b
-    cmd = f"kubectl get pod {tank_b} -o jsonpath='{{.status.podIP}}'"
-    tank_b_ip = run_command(cmd).strip()
+    tank_b_pod: V1Pod = sclient.read_namespaced_pod(name=tank_b, namespace=get_default_namespace())
```
pinheadmz: why not use the get pod helpers in k8s.py?

src/warnet/bitcoin.py
pinheadmz: (or pod_log())

src/warnet/control.py
```diff
@@ -232,52 +244,61 @@ def run(scenario_file: str, additional_args: tuple[str]):
...
-    if selected:
-        pod_name = selected["pod"]
-        try:
-            command = f"kubectl logs pod/{pod_name} -n {namespace} {follow_flag}"
-            stream_command(command)
-        except Exception as e:
-            print(f"Please consider waiting for the pod to become available. Encountered: {e}")
-    else:
-        pass  # cancelled by user
+    container_name = "bitcoincore" if pod_name.startswith("tank") else None
```
pinheadmz: we can't ever rely on tank names conforming to any kind of convention

src/warnet/k8s.py
```diff
@@ -104,29 +141,54 @@ def apply_kubernetes_yaml_obj(yaml_obj: str) -> None:
...
+        else:
+            raise
 
 
-def delete_pod(pod_name: str) -> bool:
-    command = f"kubectl delete pod {pod_name}"
-    return stream_command(command)
+def get_default_namespace() -> str:
+    with open(KUBECONFIG) as file:
+        kubeconfig_data = yaml.safe_load(file)
```
pinheadmz: will this file always exist? is this the only way to get the namespace?

pinheadmz:
Still a few mentions `git grep kubectl`, but mostly just docs (should update) and bash scripts (do we still need?). The port forward command will be replaced by ingress #607

mplsgrant:
> Am I correct in thinking that all users may still require `kubectl` to be installed, so that the cluster can be configured? Or does the python lib bypass all that too?

@willcl-ark  The python lib bypasses this. It currently relies on KUBECONFIG which is genreally at ~/.kube/config

> it seems like there could be alternatives to load_kube_config()

I think the KubeConfigLoader provides access to KUBECONFIG, but it can also do config files in custom locations (maybe in memory, too?). Seems like the google credentials thing might be an easy win to modify KUBECONFIG with the correct google plugin data, but I have not had a chance to look into that yet.

> Users may also want kubectl for any additional k8s applications, such as k9s

I don't think k9s or ktop requires kubectl, but I do think we should guide users towards downloading it b/c it is so useful.

Regarding setting the kubeconfig file, right now we rely on minikube to do that. I have also updated `auth` to create a KUBECONFIG file from a credential file if the user does not have an existing one:  https://github.com/bitcoin-dev-project/warnet/pull/614/commits/c66648081ac25eb11c0c9336b52ec910fb6e9434

src/warnet/bitcoin.py
mplsgrant: addressed here via _preload_content=False: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/16f869a2a3725ae61f895e153320d6ea18a13add

src/warnet/bitcoin.py
mplsgrant: fixed up here: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/778a3ebab87474086fe39565e579169f49af2465

src/warnet/bitcoin.py
mplsgrant: fixed up here: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/6d8ffa091089dfadca4fff2e61331c1e8d109f66
and here: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/e84416bc72411f2dbf00615486428260a5085824

src/warnet/control.py
mplsgrant: fixed up here: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/04be184b91f8d3bf370532add51e80d47e28bd2f

src/warnet/k8s.py
mplsgrant: I will need to revisit this.

mplsgrant:
It appears that signet remains an issue. Will need to revisit.

src/warnet/k8s.py
mplsgrant: I replaced open(KUBECONFIG) instances with two functions that handle errors: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/b1365d3160ec51925923c6b0173ff502075d40d1

Regarding getting the namespace, I do believe that the namespace stored in the KUBECONFIG file is the source of truth for the namespace for user OR our default namespace which is "warnet" right now. We talked about changing that on the call to "default". I

mplsgrant:
> It appears that signet remains an issue. Will need to revisit.

I fixed up the signet test issue: https://github.com/bitcoin-dev-project/warnet/pull/614/commits/02562b216fd26ad201f5929e57a628fae2ccd4ac

It uses shlex to split up lines. Allegedly, shlex is command-line-escape-character aware.

PR #615, opened 2024-09-19
Add docker app image

willcl-ark:
Ok I _think_ this is better than a flatpak (more cross-platform).

Builds a docker image with:

- `warnet`
- `k9s`
- `kubectl`
- `ktop`

...installed.

I set `fish` as the default shell, as that gets you pretty tab-autocompletion.

All that's needed is to configure `~/.kube/config` inside the machine (does `warnet auth` do this?), and you're good to go.

Current CI is set to build PR branches and pushes to master (tagged as `latest`), but we can change that if desired.

I will add a test command for it after the image has built in the first run here.

### DIFF
--- a/resources/images/warnet-bundle/Dockerfile
+++ b/resources/images/warnet-bundle/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.11-alpine
+
+# Install uv and ktop by copying from their docker images
+COPY --from=ghcr.io/astral-sh/uv:0.4.12 /uv /bin/uv
+COPY --from=ghcr.io/vladimirvivien/ktop:latest /ko-app/ktop /bin/ktop
+
+RUN apk add --no-cache bash curl ca-certificates git kubectl helm k9s fish vim mc nano
+
+# Setup venv and install warnet
+ADD . /warnet
+WORKDIR /warnet
+RUN uv sync
+
+# Setup autocomplete
+RUN mkdir -p /root/.config/fish/completions
+RUN cat <<EOF > /root/.config/fish/completions/warnet.fish
+_WARNET_COMPLETE=fish_source warnet | source
+EOF
+
+# Start
+CMD ["/usr/bin/fish", "-c", "source /warnet/.venv/bin/activate.fish; cd /root; exec fish"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,10 @@
 .git
+.github
 
 **/__pycache__
 .trunk
 .venv
+.envrc
 build
 dist
 **/*.egg-info
--- a/.github/workflows/deploy-app.yml
+++ b/.github/workflows/deploy-app.yml
@@ -1,0 +1,91 @@
+name: Publish Warnet app bundle image
+on:
+  push:
+    branches:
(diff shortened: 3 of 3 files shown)


willcl-ark:
Ah, I think perhaps a new workflow from a PR (from a fork) cannot access repo secrets (username and password) causing the login failure?

The run [here](https://github.com/willcl-ark/warnet/actions/runs/10937748431?pr=31) works, and will allow testing with:

```bash
docker run -it will8clark/warnet-app:pr-31
```

You should configure *~/.kube/config* inside the container in the normal way. then you can interact with a cluster as usual.

If you are trying to use a local cluster, then you will need to allow the docker container to talk to (outer) host with `--net=host`, i.e. by using:

```bash
docker run --net=host -it will8clark/warnet-app:pr-31
```

This will allow the container, when looking up a cluster with an address like `0.0.0.0:42813`, to find it on the local cluster. This is probably also possible to set up with specific port forwarding, but you'd need to know the cluster port before starting etc.

mplsgrant:
Cool!

This seems to work. I needed to add ` bash` to the Dockerfile because we need that until we move away from `kubectl`.

I was able to connect to a remote cluster by using a config file. I was also able to connect to my local minikube by copying my minikube data dir to the container.

willcl-ark:
> Cool!
> 
> This seems to work. I needed to add ` bash` to the Dockerfile because we need that until we move away from `kubectl`.
> 
> I was able to connect to a remote cluster by using a config file. I was also able to connect to my local minikube by copying my minikube data dir to the container.

Thanks for testing. I added `bash` to the latest image (building now over in https://github.com/willcl-ark/warnet/pull/31/checks...)

mercie-ux:
Great idea bundling all the tools into one image, saves contributors a lot of setup time.
Tested locally and everything works.

resources/images/warnet-bundle/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM python:3.11-alpine
+
+# Install uv and ktop by copying from their docker images
+COPY --from=ghcr.io/astral-sh/uv:0.4.12 /uv /bin/uv
+COPY --from=ghcr.io/vladimirvivien/ktop:latest /ko-app/ktop /bin/ktop
+
+RUN apk add --no-cache bash curl ca-certificates git kubectl helm k9s fish vim mc nano
```
mercie-ux, line 7: `RUN apk add --no-cache bash curl ca-certificates git kubectl helm fish vim mc nano && \
    curl -sL https://github.com/derailed/k9s/releases/latest/download/k9s_Linux_amd64.tar.gz | tar -xz -C /usr/local/bin k9s
`

 The Dockerfile build is currently broken, `k9s` doesn't appear to be available in the Alpine package registry, so `apk add k9s fails with exit code 10`. It's possible it was available when this PR was first opened but has since been removed. I Downloaded the binary directly from GitHub releases and it fixes this.

pinheadmz: @mercie-ux if this is something you think youd use would you want to take over the PR? (Open a new updated version)

PR #616, opened 2024-09-24
Using team namespaces: Kubeconfig Step Thru

mplsgrant:
I've created this as an extension on #598. It includes 
- some light admin documentation 
- an update to user permissions
- an updated `deploy` function to create tanks in a specific namespace and "--to-all-users"
- an updated `down` function that asks before nuking a whole wargame (and a "--force" to not bother asking)
- e2e test 
- handling of namespace throughout many other functions: status, logs, etc
- connecting nodes across namespaces using dns: tank-0003.wargames-blue-team.svc.cluster.local
- fix to continue checking the init container if at first it can't be found
- solves #618


mplsgrant:
@josibake Made my changes in this Stepthru pull. The prefix logic in #598 wasn't working like I wanted.
I plan to look into this again Fri afternoon.

josibake:
Nice, took a brief look and this looks good! I think the prefix approach is fine, and then we can revisit in a follow up a more comprehensive approach based on labels/annotations, which would remove the need for any sort of naming convention in our namespaces.

josibake:
Did a walk through of the commits, looks good. I think it would be helpful to squash the commits into logical groups, e.g., one commit for adding the ns option to the cli arguments, one commit for all of the namespace prefix logic, etc. I found it a bit confusing to step through each commit where later commits were re-editing lines of code that had already been touched.

Alternatively, people can just review the files, but given the scope of the PR I think logical commits would be helpful.

mplsgrant:
@josibake Updated the commit structure as you suggested.

mplsgrant:
@josibake This PR is looking good I think. Added a flag to `deploy` so that we can easily deploy the network to multiply namespaces at once.

mplsgrant:
rebased on main

src/warnet/admin.py
```diff
@@ -33,3 +35,93 @@ def init():
...
+  user:
+    token: {token}
+contexts:
+- name: {sa}-{namespace}
+  context:
+    cluster: {cluster_name}
+    namespace: {namespace}
+    user: {sa}
+current-context: {sa}-{namespace}
+"""
```
pinheadmz: might be cleaner to implement as a dict and then use `yaml.dump()`

pinheadmz (commented):
ok this is a lot, i breezed through each commit and added one comment. going to focus on test and docs next

src/warnet/bitcoin.py
```diff
@@ -23,23 +24,24 @@ def bitcoin():
...
 
 
-def _rpc(tank: str, method: str, params: str):
+def _rpc(tank: str, method: str, params: str, namespace: Optional[str] = None):
     # bitcoin-cli should be able to read bitcoin.conf inside the container
     # so no extra args like port, chain, username or password are needed
-    namespace = get_default_namespace()
-
+    if not namespace:
+        namespace = get_default_namespace()
```
pinheadmz: i wonder if the repetition of this can be smoothed out with a function decorator like `@needs_namespace` that can be applied to all relevant commands

src/warnet/admin.py
```diff
@@ -33,3 +37,93 @@ def init():
...
+
+    cluster_name = get_kubeconfig_value("{.clusters[0].name}")
+    cluster_server = get_kubeconfig_value("{.clusters[0].cluster.server}")
+    cluster_ca = get_kubeconfig_value("{.clusters[0].cluster.certificate-authority-data}")
+
+    os.makedirs(kubeconfig_dir, exist_ok=True)
+
+    # Get all namespaces that start with prefix
+    # This assumes when deploying multiple namespacs for the purpose of team games, all namespaces start with a prefix,
+    # e.g., tabconf-wargames-*. Currently, this is a bit brittle, but we can improve on this in the future
```
pinheadmz, line 77: I'm still pretty worried about the tech debt from this mechanism since it is so ubiquitous in this PR. Hard-coding the k8s namespaces and logging namesapce as well... and forcing the user to use a naming convention... all seems to go against our design principles!

pinheadmz:
test fails locally bc im using docker desktop:

```

Network connected
2024-10-03 08:48:27 | INFO    | test     | Authenticating and becoming bob...
2024-10-03 08:48:27 | INFO    | test     | Stopping network
2024-10-03 08:48:27 | DEBUG   | test     | Executing warnet command: down --force
2024-10-03 08:48:28 | DEBUG   | test     | Waiting for predicate with timeout 60s and interval 1s
2024-10-03 08:48:29 | INFO    | test     | Waiting for all tanks to reach 'stopped': {'total': 1, 'running': 1}
  File "/Users/matthewzipkin/Desktop/work/warnet/test/namespace_admin_test.py", line 124, in <module>
    test.run_test()
  File "/Users/matthewzipkin/Desktop/work/warnet/test/namespace_admin_test.py", line 32, in run_test
    self.authenticate_and_become_bob()
  File "/Users/matthewzipkin/Desktop/work/warnet/test/namespace_admin_test.py", line 69, in authenticate_and_become_bob
    assert get_kubeconfig_value("{.current-context}") == "minikube"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

```

src/warnet/bitcoin.py
```diff
@@ -126,16 +139,29 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
 @click.argument("tank_a", type=str, required=True)
 @click.argument("tank_b", type=str, required=True)
 @click.option("--chain", default="regtest", show_default=True)
-def messages(tank_a: str, tank_b: str, chain: str):
+@click.option("--namespace_a", default=None, show_default=True)
+@click.option("--namespace_b", default=None, show_default=True)
```
josibake: Is this necessary if we instead just passed the fully tank hostname, e.g., `tank.namespace.svc`?

src/warnet/bitcoin.py
```diff
@@ -160,23 +186,25 @@ def messages(tank_a: str, tank_b: str, chain: str):
         print(f"Error fetching messages between nodes {tank_a} and {tank_b}: {e}")
 
 
-def get_messages(tank_a: str, tank_b: str, chain: str):
+def get_messages(tank_a: str, tank_b: str, chain: str, namespace_a: str, namespace_b: str):
```
josibake, line 195: same comment regarding fully tank hostname. Might be slightly annoying to have to type out the full name if you're only operating on one namespace (majority use case), but the plus side is simpler code?

src/warnet/control.py
```diff
@@ -284,35 +335,43 @@ def filter(path):
...
+        namespace = get_default_namespace()
 
     if pod_name == "":
         try:
             pods = get_pods()
-            pod_list = [item.metadata.name for item in pods.items]
+            pod_list = [
+                f"{item.metadata.name}: {item.metadata.namespace}"
+                for item in pods
+                if item.metadata.namespace not in [LOGGING_NAMESPACE, INGRESS_NAMESPACE]
```
josibake: would be nice if this logic was inside some sort of `get_namespaces` function, instead of the RPC command itself. It seems like we will eventually want to get rid of any sort of hardcoding/namespace conventions, so hiding all of the hardcoded logic inside functions will make this _much_ easier to refactor in the future.

src/warnet/deploy.py
```diff
@@ -42,13 +49,31 @@ def validate_directory(ctx, param, value):
     callback=validate_directory,
 )
 @click.option("--debug", is_flag=True)
-def deploy(directory, debug):
+@click.option("--namespace", type=str, help="Specify a namespace in which to deploy the network")
+@click.option("--to-all-users", is_flag=True, help="Deploy network to all user namespaces")
```
josibake, line 54: I'm not crazy about this option.. it feels like we are building a use case specific convenience flag into the CLI. What's the advantage of having this vs just writing a simple bash script that deploys the network to each namespace by calling `warnet deploy /path/to/network --namespace X; warnet deploy path/to/network --namespace Y` ?

josibake: I also quite like that `warnet deploy` is somewhat generic in that it can deploy namespaces / networks. but now this flag is specific to only deploying networks.

src/warnet/k8s.py
```diff
@@ -359,3 +388,43 @@ def write_file_to_container(pod_name, container_name, dst_path, data):
...
+
+    except ApiException as e:
+        if e.status == 403:
+            ns = sclient.read_namespace(name=get_default_namespace())
+            return [ns]
+        else:
+            return []
+
+
+def get_namespaces_by_prefix(prefix: str) -> list[V1Namespace]:
```
josibake: if the plan is to eventually move to labels, we could call this function `get_namespaces_by_type` or something similar, and then have this function be a black box. Right now its using prefix, but we could refactor just this function in the future to actually use labels/annotations.

src/warnet/service_accounts.py
```diff
@@ -0,0 +1,38 @@
...
+)
+from .k8s import get_static_client
+
+
+@click.group(name="service-accounts")
+def service_accounts():
+    """Service account commands"""
+
+
+@service_accounts.command()
```
josibake: What's the use case for having this? Is this just to avoid running a kubectl command?

test/namespace_admin_test.py
```diff
@@ -0,0 +1,124 @@
...
+from test_base import TestBase
+
+from warnet.k8s import get_kubeconfig_value
+from warnet.process import run_command
+
+
+class NamespaceAdminTest(TestBase):
+    def __init__(self):
+        super().__init__()
+        self.namespace_dir = (
```
pinheadmz, line 23: the test in https://github.com/bitcoin-dev-project/warnet/pull/622 would be a good model for this one as well. after running the test locally I end up with a `kubeconfigs/` directory in the repo root. The test should operate in /tmp

src/warnet/admin.py
josibake: Agree that the prefixes is an anti pattern. I was looking at using namespace labels so they we could do things like `get_namespaces_by_type(WARNET_TANKS)` / `get_namespaces_by_type(LOGGING)` / `get_namespaces_by_type(ALL)`, etc but that felt like it would take more work to get right. If the tradeoff is we have prefixed namespaces and can use namespaces for Tabconf, vs we don't have namespaces for Tabconf at all, I'm more inclined to proceed as is and have a backlog to refactor to using labels (happy to own following up on that).

src/warnet/admin.py
```diff
@@ -33,3 +37,93 @@ def init():
...
+    help="Directory to store kubeconfig files (default: kubeconfigs)",
+)
+@click.option(
+    "--token-duration",
+    default=172800,
+    type=int,
+    help="Duration of the token in seconds (default: 48 hours)",
+)
+def create_kubeconfigs(kubeconfig_dir, token_duration):
+    """Create kubeconfig files for all ServiceAccounts in warnet team namespaces starting with <prefix>."""
```
josibake: help is outdated (mentions passing a prefix argument)

josibake:
Some testing notes:

1. deployed namespaces
2. created kube-configs
3. authenticated as Alice (`warnet auth <kubeconfig>`
4. ran `warnet run scenario/mining_std.py --allnodes`

error:

```
warnet run scenarios/miner_std.py --allnodes
Including: __init__.py
Including: commander.py
Including: miner_std.py
Including: test_framework
Including: test_framework/__init__.py
Including: test_framework/address.py
Including: test_framework/authproxy.py
Including: test_framework/bdb.py
Including: test_framework/blockfilter.py
Including: test_framework/blocktools.py
Including: test_framework/coverage.py
Including: test_framework/descriptors.py
Including: test_framework/ellswift.py
Including: test_framework/key.py
Including: test_framework/messages.py
Including: test_framework/muhash.py
Including: test_framework/netutil.py
Including: test_framework/p2p.py
Including: test_framework/psbt.py
Including: test_framework/ripemd160.py
Including: test_framework/script.py
Including: test_framework/script_util.py
Including: test_framework/secp256k1.py
Including: test_framework/segwit_addr.py
Including: test_framework/siphash.py
Including: test_framework/socks5.py
Including: test_framework/test_framework.py
Including: test_framework/test_node.py
Including: test_framework/test_shell.py
Including: test_framework/util.py
Including: test_framework/wallet.py
Including: test_framework/wallet_util.py
Failed to deploy scenario commander: miner_std
Error: Error: create: failed to create: secrets is forbidden: User "system:serviceaccount:wargames-red-team:alice" cannot create resource "secrets" in API group "" in the namespace "wargames-red-team"
```

will need to dig into the commander image more, but kinda weird that the commander image needs to create "secrets"? Easy fix is to just change the default permissions for the role to allow users to create secrets in their namespaces, but would also like to understand why this permission is needed in the first place.

josibake:
Also, when running `warnet down`, it doesn't actually delete any resources (expected, per the permissions we gave Alice), but it tells the user that it did do the teardown successfully (lies!):

```
warnet down
Preparing to bring down the running Warnet...
The listed namespaces will be affected:
  wargames-red-team
[?] Do you want to bring down the running Warnet? (y/N): y

Bringing down the warnet...

Initiated uninstall for: tank-0001 in namespace wargames-red-team
Initiated uninstall for: tank-0003 in namespace wargames-red-team
Initiated uninstall for: tank-0002 in namespace wargames-red-team
Initiated deletion of pod: tank-0002 in namespace wargames-red-team
Initiated deletion of pod: tank-0001 in namespace wargames-red-team
Initiated deletion of pod: tank-0003 in namespace wargames-red-team
Teardown process initiated for all components.
Note: Some processes may continue in the background.
Warnet teardown process completed.
```

mplsgrant:
rebased on main

mplsgrant:
> Also, when running `warnet down`, it doesn't actually delete any resources (expected, per the permissions we gave Alice), but it tells the user that it did do the teardown successfully (lies!):

@josibake I supposed they are lies by omission. After all, it did initiate the deletion. The teardown process did complete, but part of that process is not to allow Alice to delete nodes.

That said, we should probably query the user's permissions and mention that they can't do what they are trying to do. 

Or, we let `down` run in the foreground and report back on its success.

mplsgrant:
> will need to dig into the commander image more, but kinda weird that the commander image needs to create "secrets"?

@josibake The other move right now might be to get rid of Alice and her hobbled permissions. She makes the `down` command not work, and we get this secrets error in her account. Do we really need a user with her permissions?

mplsgrant:
> test fails locally bc im using docker desktop:

@pinheadmz Thank you. Fixed here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/c638a1abbb1ca5b9fa32ae0c4fc37ff8905c1977 "testing: remove mention of minikube"

josibake:
> Do we really need a user with her permissions?

Permissions are configuration based (mostly), so I'd expect either warnet to tell me I'm not allowed to do something or complete the action successfully, based on my permissions. FWIW, I switched to Bob and also tried to run a scenario. Didn't get the permissions error but the container immediately went to a succeeded status, which seems wrong?

test/namespace_admin_test.py
mplsgrant: Thank! Fixed in https://github.com/bitcoin-dev-project/warnet/pull/616/commits/64d1b40bd4f9bd345cb481dcdaa875e08c1b00c0 "tesing: use a temporary directory"

src/warnet/bitcoin.py
mplsgrant: I would like to be able to specify the bare tank name and namespace. Maybe we could also parse the tank name to determine if it is a dns name, and act accordingly. 

I would also like to give `messages` the `logs` treatment.

src/warnet/service_accounts.py
mplsgrant: Yes. Also, if we move towards an `-o json` flag then this becomes more powerful and helpful within the code base.

src/warnet/admin.py
mplsgrant: Thank you. Fix here https://github.com/bitcoin-dev-project/warnet/pull/616/commits/9b1d070e4bdc58e7bef8821c33bc6c4572a84e88

src/warnet/service_accounts.py
josibake: Hm, I'm not seeing how this is helpful. From what I can tell, this command would only be ran by an admin, and it's unclear what they would do with the output of this command , without needing to run more kubectl commands. If the argument is we are going to keep expanding our codebase to do k8s administrative work, I'd say this seems like we are re-writing kubectl in warnet and only for an admin user, which doesn't seem like a good idea to me.

mplsgrant: Sounds good. Removed it here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/d29b30e03259b2bc286f63b8b349232311fcd4c2 and https://github.com/bitcoin-dev-project/warnet/pull/616/commits/78fa9050c51395e3fa6b25ca792f3b5623e656cc

src/warnet/admin.py
mplsgrant: Fixed here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/f50dcf82335ce8c24a23b0e8a2775b7df5686537 "admin: make kubeconfig a dict"

src/warnet/bitcoin.py
mplsgrant: Fixed here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/2a906e28b5ed17efe74a75976c4781507b501af4 "DRY out the namespace check"

src/warnet/control.py
mplsgrant: Fixed. Simplified the logic using labels here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/f7e43c47f83d763660cd4ac550e1f42ed59c799c "constants: use labels to select pods for `log`"

src/warnet/deploy.py
mplsgrant: What if we had a `deploy_to_users()` function in the `admin` section? I would like to avoid using bash scripts.

src/warnet/k8s.py
mplsgrant: Refactored here: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/049e2dfebfe3e5307c3688aaa176226c7072826a "rename to `get_namespaces_by_type`"

mplsgrant:
> Also, when running `warnet down`, it doesn't actually delete any resources (expected, per the permissions we gave Alice), but it tells the user that it did do the teardown successfully (lies!):

@josibake Fixed here by checking user permissions: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/8ad00cd2f42f41667ee4d7c92588d57727d04946 and https://github.com/bitcoin-dev-project/warnet/pull/616/commits/4c62c79294e0dad5cc1ac5386b279a1a92322f50
"k8s: add can_delete_pods function" and "Using team namespaces: Kubeconfig Step Thru"

m3dwards:
Test now passes for me on Docker Desktop. Was having multiple local k8s that people use for dev useful here to drive out a problem??? haha.

test/namespace_admin_test.py
```diff
@@ -0,0 +1,127 @@
...
+        self.log.info(self.warnet(f"deploy {self.network_dir} --to-all-users"))
+        self.wait_for_all_tanks_status()
+        self.log.info("Waiting for all edges")
+        self.wait_for_all_edges()
+
+    def authenticate_and_become_bob(self):
+        self.log.info("Authenticating and becoming bob...")
+        assert get_kubeconfig_value("{.current-context}") == self.initial_context
+        self.log.info(self.warnet("auth kubeconfigs/bob-wargames-red-team-kubeconfig"))
+        assert get_kubeconfig_value("{.current-context}") == "bob-wargames-red-team"
```
pinheadmz: just so I'm clear: a context is something like `docker-desktop` but also its `username + namespace` ?

docs/admin.md
```diff
@@ -0,0 +1,70 @@
...
+## The *namespaces* directory
+This directory contains a Helm chart named *two_namespaces_two_users*.
+
+Modify this chart based on the number of teams and users you have.
+
+Deploy the *two_namespaces_two_users* chart.
+
+```shell
+$ cd namespaces
+$ warnet deploy two_namespaces_two_users
```
m3dwards: Perhaps:

```suggestion
$ warnet deploy namespaces/two_namespaces_two_users
```

test/namespace_admin_test.py
mplsgrant: My understanding is...

context is a cluster ("minikube", "digital-ocean-string", etc) + username ("john") + optionally a namespace ("default", "wargames-red-team", etc)

Also, we store contexts in the kubeconfig file.

docs/admin.md
mplsgrant: Thank you. Fixed: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/4460f20a6e927fcb9e49bbf573f15dafd5ed50e4 "admin.md: update namespace deploy command"

mplsgrant:
rebased on main

m3dwards:
I think the test should either cleanup it's modifications to your kubeconfig or use a temporary kubeconfig.

m3dwards:
Miner scenario runs for me when I'm authenticated as one of the users.

m3dwards:
I've not poured over the code, more looked at this from a functional perspective. It's working for me, left a few comments.

pinheadmz:
Using `warnet admin init` to update my battle repo:
- i wish it ONLY created the `namespaces/` directory. i already have `networks/` and i dont need another copy of `6_node...`. It also doesn't copy the default `scenarios/` so its incomplete.
- i wish it did the inquirer thing like `warnet create` does. otherwise i just get plopped an example namespace, and i have to reverse-engineer that example to make what i need

m3dwards:
> * i wish it ONLY created the `namespaces/` directory

aren't you likely to create a network so having a reference there is handy? I mean we could copy namespaces in too, the point of the init is to copy some examples?

As for not copying scenarios, I can understand why it doesn't. It wouldn't be expected that an admin would be running scenarios?

m3dwards:
I think this PR should be updated to remove `--create-namespace` from the helm commands. This prevents a user from deploying their own network.

pinheadmz:
> aren't you likely to create a network so having a reference there is handy? I mean we could copy namespaces in too, the point of the init is to copy some examples?

i had already `warnet create`  a few weeks ago and i am fiddling with my networks. today i decide to add namespaces and i end up with this:

```
/
  networks/
  scenarios/
  admin/
    networks/
    namespaces/
```

pinheadmz:
getting this now, but i have moved some files around ...?


```
(.venv) --> warnet deploy namespaces/armies/
Deploying namespace: wargames-red
Release "wargames-red" does not exist. Installing it now.
NAME: wargames-red
LAST DEPLOYED: Fri Oct  4 13:26:51 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/bin/warnet", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/.venv/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/src/warnet/deploy.py", line 62, in deploy
    _deploy(directory, debug, namespace, to_all_users)
  File "/Users/matthewzipkin/Desktop/work/warnet/src/warnet/deploy.py", line 83, in _deploy
    deploy_namespaces(directory)
  File "/Users/matthewzipkin/Desktop/work/warnet/src/warnet/deploy.py", line 298, in deploy_namespaces
    temp_override_file_path.unlink()
  File "/opt/homebrew/Cellar/python@3.12/3.12.5/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 1342, in unlink
    os.unlink(self)
PermissionError: [Errno 1] Operation not permitted: '.'
```

pinheadmz:
sorry for spamming the PR comments but I think this is a necessary fix that should be included:

```diff
diff --git a/src/warnet/deploy.py b/src/warnet/deploy.py
index 63c888e3..f32156c4 100644
--- a/src/warnet/deploy.py
+++ b/src/warnet/deploy.py
@@ -273,7 +273,7 @@ def deploy_namespaces(directory: Path):
     for namespace in namespaces_file["namespaces"]:
         click.echo(f"Deploying namespace: {namespace.get('name')}")
         try:
-            temp_override_file_path = Path()
+            temp_override_file_path = ""
             namespace_name = namespace.get("name")
             namespace_config_override = {k: v for k, v in namespace.items() if k != "name"}
 
@@ -294,7 +294,7 @@ def deploy_namespaces(directory: Path):
             click.echo(f"Error: {e}")
             return
         finally:
-            if temp_override_file_path.exists():
+            if temp_override_file_path:
                 temp_override_file_path.unlink()
 
 
```

docs/admin.md
```diff
@@ -0,0 +1,69 @@
...
+```shell
+$ warnet auth alice-wargames-red-team-kubeconfig
+```
+
+### Set up a network for your users
+Before letting the users into your cluster, make sure to create a network of tanks for them to view.
+
+
+```shell
+$ warnet admin deploy networks/mynet --to-all-users
```
pinheadmz: I think this is just `deploy` not `admin deploy` ?

m3dwards:
> i had already `warnet create`

This is a good point. If you want networks, you can just call warnet init.

pinheadmz:
Another thing that would make the battle easier to set up (for me) is if the namespaces list also included a "battlefield" for the admin to use. Then I can confidently hardcode tank names like `miner.battlefield` etc my flow is looking like this:

1. generate network files: 120-tank battlefield, 3-node armada
2. generate namespaces: 1 for each team, 1 battlefield for admin
3. generate config files for each namespace
4. `warnet auth battlefield && warnet deploy networks/battlefield` as admin
5. then players will `warnet auth wargames-red-team` and then can run scenarios

pinheadmz:
I think the initial warning here should list literally ALL namespaces that will be down'ed

```
(.venv) --> warnet down
Service account can delete pods in namespace 'default'.
Preparing to bring down the running Warnet...
The listed namespaces will be affected:
  default
[?] Do you want to bring down the running Warnet? (y/N): y

Bringing down the warnet...
Initiated uninstall for: wargames-white in namespace default
Initiated uninstall for: wargames-violet in namespace default
Initiated uninstall for: wargames-black in namespace default
Initiated uninstall for: wargames-battlefield in namespace default
Initiated uninstall for: wargames-yellow in namespace default
Initiated uninstall for: wargames-orange in namespace default
Initiated uninstall for: wargames-blue in namespace default
Initiated uninstall for: wargames-green in namespace default
Initiated uninstall for: wargames-brown in namespace default
Initiated uninstall for: wargames-red in namespace default
Initiated uninstall for: wargames-grey in namespace default
Teardown process initiated for all components.
Note: Some processes may continue in the background.
Warnet teardown process completed.

```

pinheadmz:
Ok i see, so my official request is that the hard coded `namespaces/` chart we template from always includes an admin service account that can admin ALL namespaces as well as warnet-logging


> Another thing that would make the battle easier to set up (for me) is if the namespaces list also included a "battlefield" for the admin to use. Then I can confidently hardcode tank names like `miner.battlefield` etc my flow is looking like this:
> 
>     1. generate network files: 120-tank battlefield, 3-node armada
> 
>     2. generate namespaces: 1 for each team, 1 battlefield for admin
> 
>     3. generate config files for each namespace
> 
>     4. `warnet auth battlefield && warnet deploy networks/battlefield` as admin
> 
>     5. then players will `warnet auth wargames-red-team` and then can run scenarios

pinheadmz:
Ok i see, so my official request is that the hard coded `namespaces/` chart we template from always includes an admin service account that can admin ALL namespaces as well as warnet-logging


> Another thing that would make the battle easier to set up (for me) is if the namespaces list also included a "battlefield" for the admin to use. Then I can confidently hardcode tank names like `miner.battlefield` etc my flow is looking like this:
> 
>     1. generate network files: 120-tank battlefield, 3-node armada
> 
>     2. generate namespaces: 1 for each team, 1 battlefield for admin
> 
>     3. generate config files for each namespace
> 
>     4. `warnet auth battlefield && warnet deploy networks/battlefield` as admin
> 
>     5. then players will `warnet auth wargames-red-team` and then can run scenarios

mplsgrant:
> I think the test should either cleanup it's modifications to your kubeconfig or use a temporary kubeconfig.

@m3dwards I have created a cleanup solution and pushed it.

mplsgrant:
> I think this PR should be updated to remove `--create-namespace` from the helm commands. This prevents a user from deploying their own network.

@m3dwards updated. Thank you!

src/warnet/deploy.py
pinheadmz: i like `--to-all-users` personally

pinheadmz:
`warnet messages` doesn't seem to work across namespaces. Auth'ed as red team, checking messages from miner in battlefield: (also tried `miner.default`)

```
(.venv) --> warnet bitcoin messages armada-0 miner
Error fetching messages between nodes armada-0 and miner: Error from server (NotFound): pods "miner" not found
```

pinheadmz:
i need some emergency exit after using `warnet auth` to an underprivleged namesapce. Even swithcing to my "docker-desktop" context im stuck:

```
(.venv) --> k get pods
No resources found in wargames-red namespace.
```


... and I cant `warnet down` !

mplsgrant:
> `warnet messages` doesn't seem to work across namespaces. Auth'ed as red team, checking messages from miner in battlefield: (also tried `miner.default`)

You would need to add the --namespace_a and --namespace_b flags I believe

pinheadmz:
would be super cool if it were just `warnet bitcoin messsges tank-0000-red
miner.default` (using the dot notation for namespace)

mplsgrant:
> would be super cool if it were just `warnet bitcoin messsges tank-0000-red miner.default` (using the dot notation for namespace)

@pinheadmz Good idea. That now works, except you an leave off "default" as it will automatically tack that on if it doesn't see a namespace.

m3dwards:
Anything outstanding?

pinheadmz:
Did this get addressed ?

https://github.com/bitcoin-dev-project/warnet/pull/616#issuecomment-2394264911

mplsgrant:
> Did this get addressed ?
> 
> [#616 (comment)](https://github.com/bitcoin-dev-project/warnet/pull/616#issuecomment-2394264911)

@pinheadmz I have now pushed a commit to address that.

mplsgrant:
@m3dwards 
Another open item is how we handle `--create-namespace`. It currently remains in the code base.

mplsgrant:
Also, note that I had updated the `auth` function and `create_kubeconfigs` in the latest commits: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/0d45104f2fae6bd2a39633e1962ef5d3090badef https://github.com/bitcoin-dev-project/warnet/pull/616/commits/e00ee5eaf6a3259dbb8e8dcae2e091209ed98b1e https://github.com/bitcoin-dev-project/warnet/pull/616/commits/4f1245a4e7e14e856617282a28d6a89f439ed77b https://github.com/bitcoin-dev-project/warnet/pull/616/commits/4d77dff7142a777cd061fd02b974711a9cefe5e4

I also had updated the `messages` function to take "tank.namespace" parameters: https://github.com/bitcoin-dev-project/warnet/pull/616/commits/191da0b7c6f078110d0e92da5fd4201ae5e7f493

mplsgrant:
rebased on main

PR #619, opened 2024-09-25
Make test_framework available for users when creating scenarios

pinheadmz:
This changes two important things about how scenarios are written:
1. Scenario files must be in a directory next to `test_framework`
2. Scenario files must have a `main()` function, instead of (or in addition to) the `__name__ == "__main__"` logic:

example:
```python
def main():
    ConnectDag().main()
```

### Projects
When a user creates a project, they will get a `scenarios/` directory which includes our default scenarios like `miner_std.py` as well as a complete copy of `test_framework`. They are welcome to edit the framework, the code in that directory is actually deployed.

### Running Scenarios
When executing `warnet run <path>` this PR bundles up the scenario file, `commander.py` and the entire `test_framework` directory into a `.pyz` archive which is compressed, base64 encoded, and added as raw data in a config map that is deployed along with a vanilla, unmodified image of `python:3.12-slim`. Deploying our (current) biggest scenario `signet_miner.py` results in a config map of 118,459 bytes, which is still significantly below the [1 MB recommended maximum for configMap](https://github.com/kubernetes/kubernetes/issues/19781).

### Side effects

The grossest thing about this PR in my opinion is that all the scenarios we use only for tests have to be moved as well, so they are in a directory with `test_framework`. To deal with that, I adjusted the "exclude" logic in `copy_scenario_defaults()` to filter by regex and then exclude `TEST_*.py`.

### TODO
- [ ] add a test that calls `warnet init` on an empty directory, runs a network and scenario from there to fully test the user flow (reccomend reviewers do this!)


src/warnet/control.py
```diff
@@ -198,9 +198,27 @@ def run(scenario_file: str, additional_args: tuple[str]):
     # Create in-memory buffer to store python archive instead of writing to disk
     archive_buffer = io.BytesIO()
 
+    # No need to copy the entire scenarios/ directory into the archive
+    def filter(path):
+        if any(
+            needle in str(path) for needle in [
+                ".pyc",
+                ".csv",
```
mplsgrant: Do we need to categorically reject csv files? It's a common format, and I can imagine a user thinking they could use it.

pinheadmz:
fuck, right...
`OSError: [Errno 7] Argument list too long: 'helm'`
I guess a command line option with 100,000 bytes of base64 is overkill. Abandoning.

PR #621, opened 2024-09-26
 Make test_framework available for users when creating scenarios

pinheadmz:
This changes two important things about how scenarios are written:
1. Scenario files must be in a directory next to `test_framework`
2. Scenario files must have a `main()` function, instead of (or in addition to) the `__name__ == "__main__"` logic:

example:
```python
def main():
    ConnectDag().main()
```

### Projects
When a user creates a project, they will get a `scenarios/` directory which includes our default scenarios like `miner_std.py` as well as a complete copy of `test_framework`. They are welcome to edit the framework, the code in that directory is actually deployed.

### Running Scenarios
When executing `warnet run <path>` this PR bundles up the scenario file, `commander.py` and the entire `test_framework` directory into a `.pyz` archive which is compressed, base64 encoded, and added to a volume via an initContainer that is deployed along with a vanilla, unmodified image of `python:3.12-slim`. 

### Side effects

The grossest thing about this PR in my opinion is that all the scenarios we use only for tests have to be moved as well, so they are in a directory with `test_framework`. To deal with that, I adjusted the "exclude" logic in `copy_scenario_defaults()` to filter by regex and then exclude `testscenario_*.py`.

### TODO
- [ ] add a test that calls `warnet init` on an empty directory, runs a network and scenario from there to fully test the user flow (reccomend reviewers do this!)


m3dwards:
LGTM. I changed the test_framework and I checked that this change was reflected the next time I ran a scenario. Will be very useful!

PR #622, opened 2024-10-01
charts: split off defaultConfig from [node] config

pinheadmz:
The `config` value could be set in both `node-defaults.yaml` and in `network.yaml` with the latter clobbering the default. This adds a second value `defaultConfig` which is concatenated with `config` so tanks can run bitcoin.conf settings from a combination of:

`baseConfig` set in values.yaml (all nodes in every warnet)
`defaultConfig` set in node-defaults.yaml (all nodes in your network)
`config` set in network.yaml (per-node individual settings)

### DIFF
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -1,14 +1,14 @@
 import os
 import re
+import subprocess
 import sys
 from datetime import datetime
 from io import BytesIO
 
 import click
-from urllib3.exceptions import MaxRetryError
-
 from test_framework.messages import ser_uint256
 from test_framework.p2p import MESSAGEMAP
+from urllib3.exceptions import MaxRetryError
 
 from .k8s import get_default_namespace, get_mission
 from .process import run_command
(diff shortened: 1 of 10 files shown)


src/warnet/bitcoin.py
```diff
@@ -187,7 +187,6 @@ def get_messages(tank_a: str, tank_b: str, chain: str):
                 file_path = f"{base_dir}/{dir_name}/{file}"
                 # Fetch the file contents from the container
                 cmd = f"kubectl exec {tank_a} -- cat {file_path}"
-                import subprocess
```
m3dwards, line 190: Wasn't there a reason this was in here?

m3dwards:
I think we should probably include a config section in 6_node network as an example.

src/warnet/bitcoin.py
pinheadmz: its up top now, i thought that might be what the linter was complaining about since it does not inform me, the guy trying to pass CI, what it was complaining about 🔥

pinheadmz:
> I think we should probably include a config section in 6_node network as an example.

doen!

pinheadmz:
@m3dwards added a test! I upcycled your existing graph_test to deploy the network it creates and verify that `config` doesn't clobber `defaultConfig` in a new project created by the inquirer UI. That test could be expanded to cover more of those `warnet create` options and use cases too!

m3dwards:
ACK. Tested and config worked for me.

PR #625, opened 2024-10-01
tool: pin ruff version

willcl-ark:
Hopefully this will fix the ruff inconsistencies between local and CI envs.

Install the correct version using uv:

```bash
# install
$ uv tool install ruff@0.6.8

# use
$ uvx ruff@0.6.8 check .
```

If this is the only version you have installed then it may be possible to still use the shorthand, if the version comes back correct:

```bash
$ uvx ruff --version
ruff 0.6.8

$ uvx ruff check.
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,19 +16,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: eifinger/setup-uv@v1
-      - run: uvx ruff check .
+      - name: Install a specific version of uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.4.4"
+          enable-cache: true
+      - run: uvx ruff@0.6.8 check .
 
   ruff-format:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v2
+      - name: Install a specific version of uv
+        uses: astral-sh/setup-uv@v3
         with:
-          version: "latest"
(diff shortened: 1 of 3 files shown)


pinheadmz:
```
--> uv tool install ruff@0.6.8
error: Failed to parse: `ruff@0.6.8`
  Caused by: Expected path (`/Users/matthewzipkin/0.6.8`) to end in a supported file extension: `.whl`, `.zip`, `.tar.gz`, `.tar.bz2`, `.tar.xz`, or `.tar.zst`
ruff@0.6.8
     ^^^^^
```

pinheadmz:
I do like this:

```
(.venv) --> uvx ruff check .
ruff failed
  Cause: Required version `==0.6.8` does not match the running version `0.6.9`

```

pinheadmz:
getting this gem now trying to pip install -e:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
bmon-infra 0.0.1 requires clii>=1.0.0, which is not installed.
bmon-infra 0.0.1 requires fscm@ git+https://git@github.com/jamesob/fscm.git, which is not installed.
bmon-infra 0.0.1 requires mitogen@ git+https://git@github.com/jamesob/mitogen.git, which is not installed.

```

pinheadmz:
ok the command that worked for me was `uvx install tool 'ruff==0.6.8'`

pinheadmz (approved):
Id suggest adding to the docs somewhere how to install the right version, then lets merge this!

willcl-ark:
> Id suggest adding to the docs somewhere how to install the right version, then lets merge this!

Done

Issue #623, opened 2024-10-01
`warnet logs` needs to specify the bitcoincore container when prometheus exporter is present


pinheadmz:
```
(.venv) --> warnet logs
[?] Please choose a pod: 
   commander-minerstd-1727811004
   commander-txflood-1727811067
 ❯ tank-0000-red
   tank-0001-red
   tank-0002-red
   tank-0003-red
   tank-0004-red
   tank-0005-red
   tank-0006-red
   tank-0007-red
   tank-0008-red
   tank-0009-red
   tank-0010-red

a container name must be specified for pod tank-0000-red, choose one of: [bitcoincore prometheus]
```

pinheadmz:
damn this command too:

```
(.venv) --> warnet bitcoin rpc tank-0016-orange getpeerinfo
Defaulted container "bitcoincore" out of: bitcoincore, prometheus
error: unable to upgrade connection: container not found ("bitcoincore")


```

PR #626, opened 2024-10-01
Add container selection logic

mplsgrant:
This fixes the logs container selection issues in #623

### DIFF
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -9,7 +9,8 @@ from test_framework.messages import ser_uint256
 from test_framework.p2p import MESSAGEMAP
 from urllib3.exceptions import MaxRetryError
 
-from .k8s import get_default_namespace, get_mission, get_pod
+from .constants import BITCOINCORE_CONTAINER
+from .k8s import get_default_namespace, get_mission
 from .process import run_command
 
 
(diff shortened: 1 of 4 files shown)


src/warnet/bitcoin.py
```diff
@@ -39,10 +38,19 @@ def _rpc(tank: str, method: str, params: str):
     # bitcoin-cli should be able to read bitcoin.conf inside the container
     # so no extra args like port, chain, username or password are needed
     namespace = get_default_namespace()
+
+    try:
+        pod = get_pod(tank)
+        container_names = [container.name for container in pod.spec.containers]
+        container_name = container_names[0]
```
pinheadmz: What determines the order of containers in a pod? Can we guarantee the one we want is always `[0]` because of the yaml file? Or should we hard code the name from the chart `bitcoincore` ?

mplsgrant: The first container has been the bitcoincore container and commander container from what I have seen, but I agree that we have no way of guaranteeing that.

I changed the logic so that we use hard coded values from the constants file which match the helm charts.

pinheadmz:
looks good! testing ...

pinheadmz (approved):
👍  ACK

mplsgrant:
Thanks @pinheadmz I'll merge

PR #627, opened 2024-10-02
scenarios: miner_std should accept single --tank option

pinheadmz:
Currently miner_std either tries to mine from ALL nodes or only the 0th node in the network. Since it also naively calls rpc `createwallet` and some older node versions do not support that command, the scenario fails. I think we can also modify the scenario to be more flexible with older versions but it also makes sense to be able to just pick a single mining node. The signet miner scenario already has this option.


m3dwards:
uTACK

PR #630, opened 2024-10-02
delete old items: graph resource & setup scripts

mplsgrant:
Remove the graph directory from resources and the setup scripts for minikube and users.


pinheadmz (approved):
CLEAN UP CLEAN UP EVERYBODY CLEAN UP

PR #632, opened 2024-10-04
chart: set rpcpassword as a value, add to labels for scenarios

pinheadmz:
The configmap helm logic then adds the chart value to bitcoin.conf and to the pod metadata labels, same mechanism we currently use for rpcport. This allows chart-writers to override rpcpassword on an individual per-tank basis. When deploying a scenario, the metadata with the password is read from the metadata labels and written to warnet.json in the commander pod. Assuming namespaces limit users ability to read pod metadata, this should restrict rpc access to users in multiplayer mode. 

I also changed the default password in the base chart values so if anything was hard coded we should see tests fail.

UPDATE: that was so smart of me, "password" was hard coded in the prometheus exporter


m3dwards:
All worked for me.

PR #639, opened 2024-10-08
add inquirer option to set pullPolicy

willcl-ark:
- add inquirer option to set pullPolicy
- ask follow-up questions in order. Previous selecting "other" meant waiting until later in the question phase before you were asked how many "other" actually meant.
    - Use inquirer [shortcut functions](https://python-inquirer.readthedocs.io/en/latest/usage.html#shortcut-functions) for all questions to simplify and allow intermediate questions in order.


willcl-ark:
Any objections to this? I'd like to merge, especially for the second commit.

@pinheadmz @m3dwards @mplsgrant

mplsgrant:
Will the first commit pull images every time we `helm`? It seems like doing version numbers on the images might let us get the newest version without Always pulling.

*Edit*: Oh, I see it defaults to "no". That seems fine.

src/warnet/graph.py
```diff
@@ -98,119 +98,90 @@ def custom_graph(
...
+        validate=lambda _, x: len(x) > 0,
+    )
+    nodes = inquirer.list_input(
+        message=click.style("How many nodes would you like?", fg="blue", bold=True),
+        choices=["8", "12", "20", "50", "other"],
+        default="12",
+    )
+    if nodes == "other":
+        nodes = inquirer.text(
+            message=click.style("Enter the number of nodes:", fg="blue", bold=True),
```
mplsgrant: >Enter the number of nodes:

Leave off the ":" because inquirer will add one automatically.

src/warnet/graph.py
```diff
@@ -98,119 +98,90 @@ def custom_graph(
...
-                "Which version would you like nodes to run by default?", fg="blue", bold=True
-            ),
-            choices=SUPPORTED_TAGS,
-            default=DEFAULT_TAG,
+        choices=["0", "1", "2", "8", "12", "other"],
+        default="8",
+    )
+    if connections == "other":
+        connections = inquirer.text(
+            message=click.style("Enter the number of connections?", fg="blue", bold=True),
```
mplsgrant: > Enter the number of connections?

Remove the "?"

willcl-ark: fixed in 711156a73898eda58a2e306fcc21a58356966d95

src/warnet/graph.py
willcl-ark: fixed in 711156a73898eda58a2e306fcc21a58356966d95

willcl-ark:
> Will the first commit pull images every time we `helm`? It seems like doing version numbers on the images might let us get the newest version without Always pulling.
> 
> _Edit_: Oh, I see it defaults to "no". That seems fine.

Yeah, I guess I *could* technically drop this commit now that we have actual version numbers on the vuln images... I still think it's kinda useful though to fetch new versions automagically if desired (and if not you just `enter` right past it)

mplsgrant:
> Yeah, I guess I _could_ technically drop this commit now that we have actual version numbers on the vuln images... I still think it's kinda useful though to fetch new versions automagically if desired (and if not you just `enter` right past it)

Agree. ACK https://github.com/bitcoin-dev-project/warnet/pull/639/commits/711156a73898eda58a2e306fcc21a58356966d95

willcl-ark:


### Conflicts

* [#707](https://github.com/bitcoin-dev-project/warnet/pull/707) (warnet init: add `metricsExport:true` if grafana is selected by pinheadmz)

PR #644, opened 2024-10-09
fix run --debug issue for service accounts

mplsgrant:
- Make it so Service Accounts ...
  - ... have access to `get` `pod/status` which solves our issue with Service Accounts calling `run ... --debug`
  - ... can get access to `patch` which lets Bob modify pods
  - ... can `update` configmaps and secrets which also helps Bob modify pods 
- Duplicate the p2p test in the scenarios file so that it runs it with `--debug`
- Modify the namespace_admin test so that it inherits from the scenarios test, allowing Bob to run all the scenarios.
- Adjust `block[best_block]` to `>=1`


test/namespace_admin_test.py
```diff
@@ -86,18 +77,28 @@ def setup_service_accounts(self):
...
-        self.log.info(self.warnet(f"deploy {self.network_dir} --to-all-users"))
+    def setup_network(self):
+        if self.current_context == self.bob_context:
+            self.log.info(f"Allowing {self.current_context} to update the network...")
+            assert self.this_is_the_current_context(self.bob_context)
+            self.warnet(f"deploy {self.network_dir}")
+        else:
+            self.log.info("Deploy networks to team namespaces")
+            assert self.this_is_the_current_context(self.initial_context)
+            self.log.info(self.warnet(f"deploy {self.network_dir} --to-all-users"))
```
pinheadmz, line 88: This deploys 2x12=24 tanks which may be a bit too much for local testing, but working ok, no change needed

test/namespace_admin_test.py
```diff
@@ -159,6 +171,11 @@ def cleanup_kubeconfig(self):
         except Exception as e:
             raise K8sError(f"Could not write to KUBECONFIG: {KUBECONFIG}") from e
 
+    def bob_runs_scenario_tests(self):
+        assert self.this_is_the_current_context(self.bob_context)
+        super().run_test()
```
pinheadmz, line 176: At first I thought this would deploy another 12 tanks for a total of 36 running together, but i think since the namespace already has 12 tanks running it just kinda redploys those?

pinheadmz: Screenshot 2024-10-09 at 10 15 57 PM

pinheadmz (commented):
I'm still having an issue on this branch using --debug, to reproduce:

```
  warnet deploy resources/namespaces/two_namespaces_two_users/
  warnet admin create-kubeconfigs
  warnet deploy resources/networks/6_node_bitcoin/ --to-all-users
  warnet auth kubeconfigs/alice-wargames-red-team-kubeconfig
  warnet run resources/scenarios/reconnaissance.py --debug

...
Failed to deploy scenario commander: reconnaissance
Error: Error: create: failed to create: secrets is forbidden: User
"system:serviceaccount:wargames-red-team:alice" cannot create resource "secrets" in API group "" in
the namespace "wargames-red-team"

```

mplsgrant:
> I'm still having an issue on this branch using --debug, to reproduce:

@pinheadmz alice has the "pod-viewer" role, which means she can't `run` anything. If you `auth` with bob's credentials, then `--debug` should work.

test/namespace_admin_test.py
mplsgrant: I think I had a lower tank count in an earlier version of this test; however, since I decided to extend the Scenarios test, it meant that I would need to use whatever it is using.

test/namespace_admin_test.py
mplsgrant: Bob re-runs the helm commands, and if I understand it correctly, this is why he needs the `patch` permission. I actually want to keep this as part of the test because it seems like a good thing to make sure pod-manager service accounts like bob have access to it.

pinheadmz:
> If you `auth` with bob's credentials, then `--debug` should work.

indeed! Great work.

pinheadmz:
confirmed test fails on main, passes on branch. actual code review nexrt :-P

test/namespace_admin_test.py
mplsgrant: I suppose I could override network in my namespace_admin test. That would let us use fewer nodes if we wanted to.

pinheadmz:
ACK 🚀

Issue #642, opened 2024-10-08
No such file or directory when running `warnet auth`


danielabrozzoni:
I got a "No such file or directory" when running `warnet auth`, I think it might be linked with the fact that I didn't have a `~/.kube` directory. Creating `/tmp/kube` and explicitly setting `export KUBECONFIG=/tmp/kube/kube.yaml` fixed it.
Error log:
```
warnet git:main  
(.venv) ❯ warnet auth config.yaml 
Traceback (most recent call last):
  File "/home/daniela/Developer/warnet/src/warnet/k8s.py", line 534, in write_kubeconfig
    with tempfile.NamedTemporaryFile("w", dir=dir_name, delete=False) as temp_file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4rf5qybw37b4lh1g0xczlv14sqdbmnpm-python3-3.11.9/lib/python3.11/tempfile.py", line 580, in NamedTemporaryFile
    file = _io.open(dir, mode, buffering=buffering,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4rf5qybw37b4lh1g0xczlv14sqdbmnpm-python3-3.11.9/lib/python3.11/tempfile.py", line 577, in opener
    fd, name = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4rf5qybw37b4lh1g0xczlv14sqdbmnpm-python3-3.11.9/lib/python3.11/tempfile.py", line 257, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/daniela/.kube/tmphbqp4j2t'

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/daniela/Developer/warnet/.venv/bin/warnet", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/daniela/Developer/warnet/.venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
[...]

PR #647, opened 2024-10-10
users: check for ~/.kube and create if needed

mplsgrant:
Within the `auth` function, check for and create (if needed) a folder appropriate to the KUBECONFIG path.

Addresses #642


pinheadmz (approved):
tested ACK thank you

PR #649, opened 2024-10-11
speedy startup

willcl-ark:
Deploy everything everywhere, all at once (except fork observer).

- no more waiting for logging to install before we even start the next thing
- could probably speed up very large networks by multiprocessing individual tank deployment

Fork observer uses k8s api (via `get_mission()`) to create its config, so it has to wait until at least that deploy is finished.

### DIFF
--- a/src/warnet/constants.py
+++ b/src/warnet/constants.py
@@ -96,6 +96,12 @@ LOGGING_CONFIG = {
     },
 }
 
+LOGGING_CRD_COMMANDS = [
+    "helm repo add prometheus-community https://prometheus-community.github.io/helm-charts",
+    "helm repo update",
+    "helm upgrade --install prometheus-operator-crds prometheus-community/prometheus-operator-crds",
+]
+
 # Helm commands for logging setup
 # TODO: also lots of hardcode stuff in these helm commands, will need to fix this when moving to helm charts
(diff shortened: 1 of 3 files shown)


pinheadmz:
ooh concept ack nice!

willcl-ark:
ping @mplsgrant: I can't understand why the namespace test should fail with these changes (although, I don't totally understand the test that well).

If you get a minute, would appreciate your eyes on it :) (p.s. it still fails even if I drop the commit touching the namespace deploy code...)

mplsgrant:
@willcl-ark It looks like it's deploying ingress regardless of whether there is an ingress flag enabled. Our Service Accounts (like bob) can't ~~do~~ *deploy* ingress. Basically, if you become Bob in the battle of galen, and you try to deploy the armada, then it will complain about not being able to deploy ingress (because of the permissions bob has).

mplsgrant:
I should also add that the namespace_admin.py test actually runs the entire scenarios_test.py.

willcl-ark:
TODO:

- [x] add multiprocessing to scenarios stop all

willcl-ark:
undrafted to see if that will make CI work...

pinheadmz (approved):
tested ACK, its fast! tytyty

PR #653, opened 2024-10-15
fix namespace bug in _logs()

pinheadmz:
There is a leftover bug from #651 that prevented the Battle signet miner command from working:

Command:
> (.venv) --> warnet run scenarios/admin/signet_miner.py --source_dir=scenarios --tank=0 generate  --min-nbits --address=$(warnet bitcoin rpc miner getnewaddress) --ongoing --debug

Error:
> Error getting pods. Could not determine primary container: cannot access local variable 'pod_namespace' where it is not associated with a value

### DIFF
--- a/src/warnet/control.py
+++ b/src/warnet/control.py
@@ -398,12 +398,12 @@ def _logs(pod_name: str, follow: bool, namespace: Optional[str] = None):
         ]
         selected = inquirer.prompt(q, theme=GreenPassion())
         if selected:
-            pod_name, pod_namespace = selected["pod"].split(": ")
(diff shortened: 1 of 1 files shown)


mplsgrant:
ACK. Works for me.

*edit* ~~The namespace component works for me, but the command in op does give me this error:~~ 

```terminal
Waiting for commander pod to start...
/bin/sh: 1: Syntax error: "(" unexpected
Deleting pod...
pod "commander-signetminer-1729005660" deleted
```

*update*: I needed to create the wallet and have the miner tank up and running, and that allowed the command in op to run!

PR #664, opened 2024-11-06
Restore LN functionality with helm

pinheadmz:
- [x] Add charts for lnd. Deploy nodes and connect to bitcoind. Manually create channels and payments with RPC
- [x] Import channels from yaml or json with policy configurations
- [x] Run ln_init scenario to establish entire LN graph with policies


resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - mountPath: /root/.lnd/lnd.conf
+          name: config
+          subPath: lnd.conf
+        - mountPath: /root/.lnd/tls.key
+          name: tlskey
+          subPath: tls.key
+        - mountPath: /root/.lnd/tls.cert
+          name: tlscert
+          subPath: tls.cert
+    {{- if .Values.circuitBreaker }}
```
m3dwards, line 59: Is the plan to have this removed before this PR is merged? Or shortly after? Circuit breaker feels like a ~service~ plugin.

resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
```diff
@@ -0,0 +1,88 @@
...
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+    - configMap:
+        name: {{ include "lnd.fullname" . }}
+      name: config
+    - configMap:
+        name: {{ include "lnd.fullname" . }}
+      name: tlskey
+    - configMap:
+        name: {{ include "lnd.fullname" . }}
```
m3dwards: Not sure if this is finished but it looks like the same config map is mounted three times under three names.

resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
pinheadmz: good catch, this is leftover from Italy. ill need circuit breaker together with simln eventually either in this PR or a follow up, so this'll get cleaned up before merge

resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
pinheadmz: ah thanks will clean up

resources/scenarios/ln_init.py
```diff
@@ -14,171 +16,391 @@ def add_options(self, parser):
...
+                    expected_ch = expected[i]
+                    capacity = expected_ch["capacity"]
+                    # We assert this because it isn't updated as part of policy.
+                    # If this fails we have a bigger issue
+                    assert int(actual_ch["capacity"]) == capacity
+                    # policy actual/expected source/target
+                    polas = Policy.from_lnd_describegraph(actual_ch["node1_policy"])
+                    polat = Policy.from_lnd_describegraph(actual_ch["node2_policy"])
+                    poles = Policy(**expected_ch["source_policy"])
+                    polet = Policy(**expected_ch["target_policy"])
```
mplsgrant, line 381: It looks like `expected_ch` pulls from the channels config map. I don't think there is enough info in the config map to construct a `Policy`. Would it be sufficient to use the information from the config map and compare it to `actual_ch`?

````json
[{"capacity":50000,"id":{"block":300,"index":2},"push_amt":25000,"target":"tank-0005-ln"}]

````

Comment (pinheadmz) [reply resources/scenarios/ln_init.py:381]: >  I don't think there is enough info in the config map to construct a Policy.

There isn't?

> Would it be sufficient to use the information from the config map and compare it to actual_ch?

The main reason we need `Policy` is to normalize all the terrible naming conventions used by LND. The network.yaml and corresponding channel configmaps conform to the [BOLT](https://github.com/lightningcn/lightning-rfc/blob/master/07-routing-gossip.md#the-channel_update-message) and the `Policy` class is used to translate between (both of) LND's non-spec names and then whatever CLN and Eclair use as well...

Comment (mplsgrant) [reply resources/scenarios/ln_init.py:381]: I may be confused, but I'm seeing this in the channel config map: 

````JSON
[{"capacity":50000,"id":{"block":300,"index":2},"push_amt":25000,"target":"tank-0005-ln"}]
````

And it seems like `Policy` wants this:

````python
        cltv_expiry_delta: int,
        htlc_minimum_msat: int,
        fee_base_msat: int,
        fee_proportional_millionths: int,
        htlc_maximum_msat: int,
````

Comment (pinheadmz) [reply resources/scenarios/ln_init.py:381]: Ah yeah specifying a channel policy is not required, lnd will just use its defaults. So you can have channel config maps without `source_policy` and/or `target_policy`. So probably I need to just not try to check channels without a policy defined at all... In fact if you try to deploy `test/data/ln` the ln_init scenario will fail for this reason! And that isn't covered by a test.
Good catch! thanks.

File: test/ln_test.py [hunk 4]
```diff
@@ -6,121 +6,80 @@
...
+        graphs = []
+        for n in range(10):
+            ln = f"tank-{n:04d}-ln"
+            res = self.warnet(f"ln rpc {ln} describegraph")
+            graphs.append(json.loads(res)["edges"])
+
+        def check_policy(node: int, index: int, field: str, values: tuple):
+            self.log.info(f"Checking policy: Node={node} ch={index} Expected={field}:{values}")
+            graph = graphs[node]
+            assert len(graph) == 13
```
mplsgrant, line 48: This assertion appears to fail spuriously: https://github.com/mplsgrant/warnet/actions/runs/12283986192/job/34278744847#step:9:192

I image if we wait for the edges to materialize it will avoid this assertion error:

````python
        def ln_node_has_n_edges(ln: str, n: int, ) -> Optional[str]:
            res = self.warnet(f"ln rpc {ln} describegraph")
            edges = json.loads(res)["edges"]
            if len(edges) != n:
                return None
            return edges

        for n in range(10):
            ln = f"tank-{n:04d}-ln"
            partial_fn = partial(ln_node_has_n_edges, ln, 13)
            edges = self.wait_for_predicate(partial_fn)
            graphs.append(edges)

        def check_policy(node: int, index: int, field: str, values: tuple):
            self.log.info(f"Checking policy: Node={node} ch={index} Expected={field}:{values}")
            graph = graphs[node]
            assert len(graph) == 13
            ch = graph[index]
            a = int(ch["node1_policy"][field])
            b = int(ch["node2_policy"][field])
            assert values == (a, b) or values == (
                b,
                a,
            ), f"policy check failed:\nActual:\n{ch}\nExpected:\n{field}:{values}"
````

Side note: I also think we should let `wait_for_predicate` to return a value:

````python
    def wait_for_predicate(self, predicate, timeout=5 * 60, interval=5):
        self.log.debug(f"Waiting for predicate with timeout {timeout}s and interval {interval}s")
        while timeout > 0:
            try:
                result = predicate()
                if result:
                    return result
            except Exception:
                pass
            sleep(interval)
            timeout -= interval
        import inspect

        raise Exception(
            f"Timed out waiting for Truth from predicate: {inspect.getsource(predicate).strip()}"
````

resources/scenarios/ln_init.py
pinheadmz: fixed https://github.com/bitcoin-dev-project/warnet/pull/664/commits/ddc93316b350d55dac6a799a2bfaa3beb8005d3f

test/ln_test.py
pinheadmz: ln_init handles all the waiting, at this point in the test we should be able to guarantee the graph for every node. Hard to tell but in your test run it looks like ln_init didn't output anything... I'll watch for this in future test runs

pinheadmz:
attempting rebase on main with speedy parallel deploy... 🤞

mplsgrant:
Changes look good, tests work, and my plugins rebase works cleanly on [c18298d](https://github.com/bitcoin-dev-project/warnet/pull/664/commits/c18298d2bde972f12e58095a277d8d38195b0bf6)

PR #665, opened 2024-11-11
Add the hook api

mplsgrant:
## Overview
This hooks system is intended to allow the user to extend Warnet with python code plugins which will have hooks into key points in the Warnet workflow.

## For the user
This hooks API lets users write plugins that look like this:

``` python
@pre_status
def thing_i_want_to_do():
    """When the user runs `warnet status`, this function will run first."""
    print("I just want to say hi")
```

The various decorators available to users can be found in the hooks_api.py file which is also found in the `plugins` directory.

To get a `plugins` directory, simply run `warnet init`, and warnet will generate one for you.

### For the Warnet developers
This new hooks system lets us decorate our internal Warnet functions with `@api` like this:

``` python
@click.command()
@api
def status():
    ....
```

Doing so causes Warnet to produce the corresponding `hooks_api.py` file when the user runs `warnet init`

## Feedback
I am looking for feedback on this approach. Earlier I had considered authoring an abstract class/interface. After implementing the hooks system using decorators, I noticed that it became much easier to get new API endpoints.


mplsgrant:
closed in favor of #670

PR #668, opened 2024-11-13
update install docs

### LINKED ISSUE #666: doc: reporting a couple confusions in dependencies installation
Following https://github.com/bitcoin-dev-project/warnet/blob/main/docs/install.md on my own to test the docs. I'm running Debian stable.

I followed "[Running warnet locally](https://github.com/bitcoin-dev-project/warnet/blob/main/docs/install.md#running-warnet-locally)" > "[Docker desktop](https://github.com/bitcoin-dev-project/warnet/blob/main/docs/install.md#docker-desktop)". Then i followed "[Testing kubectl and helm](https://github.com/bitcoin-dev-project/warnet/blob/main/docs/install.md#testing-kubectl-and-helm)" to ensure my dependencies install was correct before moving on to installing Warnet itself. I was confused that it tells to try `kubectl` command when i wasn't instructed to install it. Also the `helm install` command failed for some reason.

Asking @pinheadmz it turns out all this wasn't actually necessary since the next step takes care to install `helm` and `kubectl` locally in my virtual env.

Conclusion: the simplest route should probably be emphasized in the doc. Install Docker Desktop, create a Python venv, `pip install warnet`, `warnet setup`, done. Installing the deps globally can still be discussed but it should probably not be the default route when following the docs?

Side note: my first `warnet setup` failed because i didn't enable Kubernetes in the settings of Docker desktop, might be worth mentioning it's necessary. Although the error i got was very clear and the fix straightforward.

willcl-ark:
Closes: #666

May not directly address:

> Conclusion: the simplest route should probably be emphasized in the doc. Install Docker Desktop, create a Python venv, pip install warnet, warnet setup, done. Installing the deps globally can still be discussed but it should probably not be the default route when following the docs?

But I think the re-ordering should help significantly.

If we really want to address this comment, perhaps we should re-introduce `warnet quickstart` or something?

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,51 +1,8 @@
 # Installing Warnet
 
-Warnet requires Kubernetes (k8s) and helm in order to run the network. Kubernetes can be run remotely or locally (with minikube or Docker Desktop). `kubectl` and `helm` must be run locally to administer the network.
-
-## Dependencies
-
-### Remote (cloud) cluster
-
-The only two dependencies of Warnet are `helm` and `kubectl` configured to talk to your cloud cluster.
-
-### Running Warnet Locally
-
-If the number of nodes you are running can run on one machine (think a dozen or so) then Warnet can happily run on a local Kubernetes. Two supported k8s implementations are Minikube and K8s as part of Docker Desktop.
-
-#### Docker Desktop
-
-[Docker desktop](https://www.docker.com/products/docker-desktop/) includes the docker engine itself and has an option to enable Kubernetes. Simply installing it and enabling Kubernetes should be enough.
-
-[Helm](https://helm.sh/docs/intro/install/) is also required to be installed.
-
-#### Minikube
-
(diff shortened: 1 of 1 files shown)


mplsgrant (approved):
This appears to address #666 
ACK

pinheadmz:
LGTM - but considering the experience weve had thus far i think the very top of the page should look something like this:

---

# TLDR

```
pip install warnet
warnet setup
```

willcl-ark:
> LGTM - but considering the experience weve had thus far i think the very top of the page should look something like this:
> 
> 
> 
> ---
> 
> 
> 
> # TLDR
> 
> 
> 
> ```
> 
> pip install warnet
> 
> warnet setup
> 
> ```

I can add this section at the top if you're serious? I do agree it would be pretty nice myself...

(But I'd add venv create/activate too)

willcl-ark:
Hmmm as I sketch it out, I'm not sure a TLDR section is necessary with the new structure. And once we include a python venv, and a k8s cluster, we just end up repeating what is already in this change?

pinheadmz:
OK! ty <3

PR #671, opened 2024-12-06
log correct number of outbound connections

cndolo:
Currently, `connections_in` is used for both inbound and outbound peers. The PR corrects this by using `connections_out` for outbound peers.


pinheadmz (approved):
tested ACK

THANK YOU Charmaine! Great catch here. I will merge to main and also build and push the revised bitcoin-exporter image to docker hub. You may need to remove your local copy of the image to force docker to pull the update.

Example dashboard using 12 node below. You can see how long it takes for all nodes to connect! 

Screenshot 2024-12-06 at 10 36 35 AM

PR #669, opened 2024-11-21
Use docker buildx bake to build images

willcl-ark:
This is much cleaner than maintaining a list of complete shell commands.

PR #673, opened 2024-12-09
image build: use cmake

### LINKED ISSUE #672: `warnet image build` still builds with autotools
` ./autogen.sh: not found`

We gotta fix this before we can use Warnet to test pull requests

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/672

Seems to be working for me so far locally, will test a few other options as well. 

Should we add a test to CI for this? It might just take too long to run...

### DIFF
--- a/src/warnet/image_build.py
+++ b/src/warnet/image_build.py
@@ -23,10 +23,7 @@ def build_image(
     arches: str,
     action: str,
 ):
-    if not build_args:
-        build_args = '"--disable-tests --without-gui --disable-bench --disable-fuzz-binary --enable-suppress-external-warnings --disable-dependency-tracking "'
-    else:
-        build_args = f'"{build_args}"'
+    build_args = '"-DWITH_ZMQ=ON "' if not build_args else f'"{build_args}"'
 
     build_arches = []
     if not arches:
--- a/resources/images/bitcoin/Dockerfile
+++ b/resources/images/bitcoin/Dockerfile
@@ -7,8 +7,8 @@ ARG BUILD_ARGS
 RUN --mount=type=cache,target=/var/cache/apk \
     sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/https\:\/\/alpine.global.ssl.fastly.net/g' /etc/apk/repositories \
     && apk --no-cache add \
-    autoconf \
-    automake \
+    cmake \
+    python3 \
     boost-dev \
     build-base \
     chrpath \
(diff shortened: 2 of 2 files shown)


willcl-ark:
We do need this, but it will break rebuilding <= v28.0 images...

I think we should simply add a *new* dockerfile with this in. When combined with https://github.com/bitcoin-dev-project/warnet/pull/669 having multiple dockerfiles becomes easy to handle IMO. To build v28.0 you'd just run:

```
docker buildx bake bitcoin-28
```

and for >28.0 you'd do:

```
docker buildx bake bitcoin-29
# or perhaps we can get working...?
docker buildx bake <commit sha>
```

and let the `bake` magic [handle which dockerfile is used](https://github.com/willcl-ark/warnet/blob/ee71928f95d885e7300b40a0b8ff17cefc02965a/docker-bake.hcl#L49-L65) for you.

pinheadmz:
Ah brilliant mate. Will review baker first then rebase this. 👨‍🍳

danvergara:
I did this PR to leverage `bake` and `cmake` together 

https://github.com/willcl-ark/warnet/pull/32

src/warnet/image_build.py
```diff
@@ -23,10 +23,7 @@ def build_image(
     arches: str,
     action: str,
 ):
-    if not build_args:
-        build_args = '"--disable-tests --without-gui --disable-bench --disable-fuzz-binary --enable-suppress-external-warnings --disable-dependency-tracking "'
-    else:
-        build_args = f'"{build_args}"'
+    build_args = '"-DWITH_ZMQ=ON "' if not build_args else f'"{build_args}"'
```
deadmanoz, line 26: Won't be able to build images <= `v28.X` if this change is made.

I tested building `v28.1` using the new `bake` functional:
- pre this change - builds fine
- post this change - build fails

```
28.35 + ./configure 'LDFLAGS=-L/lib/' 'CPPFLAGS=-g0 -I/include/ --param ggc-min-expand=1 --param ggc-min-heapsize=32768' '--prefix=/opt/bitcoin' '-DWITH_ZMQ=ON'
28.48 configure: error: unrecognized option: '-DWITH_ZMQ=ON'
```

deadmanoz:
@pinheadmz, @willcl-ark do you wish for users to be able to use `warnet image build` for arbitrary Bitcoin Core commits (pre the CMake switchover of https://github.com/bitcoin/bitcoin/pull/30454)?

If so, we'll need to support the 2 different styles of `build_args` (and reference the different Dockerfiles) to do so.

https://gist.github.com/hebasto/2ef97d3a726bfce08ded9df07f7dab5e

pinheadmz:
@deadmanoz yes but I think it's fair to assume that from now on, all pull requests will be cmake-based. However, I guess it's possible that someone may want to modify an old version of bitcoin and test that in warnet...

I wonder if there's an easy trick we can use to determine which build system to execute, maybe just like the presence of `autogen.sh` in the root directory ?

PR #674, opened 2024-12-10
Simple replacement cycling attack

a-mpch:
## Issue

I'm trying to replicate replacement cycling attacks from Antoine Riard's example and paper and how implementations were made.

## Solution
Built heavily inspired in #422 but doing a simplification of the attack. We don't use HTLCs but we use the same condition of a "shared utxo" and assume the time sensitive. Using this simplifcation I can build a simple transaction cycling and pinning the transaction to the mempool.

### Next

- [ ] Build Stochastic scenario 
- [ ] Build current mitigations using the stochastic scenario
- [ ] Other node topologies?

### Would highly appreciate
I think doing the random_interval hack and the process of the cycling actually check mempool might be an ok approach. Any input on this would be highly appreciated

### DIFF
--- a/resources/scenarios/miner_std.py
+++ b/resources/scenarios/miner_std.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from random import randint
 from time import sleep
 
 from commander import Commander
@@ -34,6 +35,12 @@ class MinerStd(Commander):
             default=60,
             type=int,
             help="Number of seconds between block generation (default 60 seconds)",
+        )
+        parser.add_argument(
+            "--random-interval",
+            dest="random_interval",
+            action="store_true",
+            help="Should the interval be randomized between 0 and the interval value",
         )
         parser.add_argument(
             "--mature",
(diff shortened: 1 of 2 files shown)


mplsgrant:
This looks really cool. Thank you for putting it together. I would like to take a closer look later, but I want to quickly mention that I'm not sure it would make sense to incorporate this directly into Warnet. The original PR that you referred to may be a little misleading in the sense that it merges a scenario directly into Warnet. As more and more people start developing new scenarios, I believe the goal would be for the authors to host those scenario files in their own repository. That way, they are easier to distribute and do not rely on the Warnet maintainers approving anything.

For example, I can do a `warnet init` on some directory, and that will populate the directory with a scenarios folder, network folder, etc. I would then make the changes I want (add scenarios, modify the network file, etc). And then I would commit that to a repo on my own github account. That way, users can clone that repo and start running it.

a-mpch:
hey @mplsgrant, thanks for taking your time!
It makes sense what you are referring to. But I'm wondering how to gather feedback on scenarios if they live in my repo. Discussions? Issue? Happy to help how to navigate that.
 
Broader context, I come from [delving bitcoin post](https://delvingbitcoin.org/t/op-expire-mitigating-replacing-cycling-attacks/1282/4?u=mpch) and maybe misinterpreted the support / working scenario / PRs.

a-mpch:
This was moved to https://github.com/a-mpch/replacement-cycling-attack

PR #686, opened 2025-03-02
cli: add 'port_forward' subcommand for ln

f3r10:
Add a subcommand for port forwarding from ln pods to a local port so a ln pod can be referenced from the host machine.

### DIFF
--- a/src/warnet/ln.py
+++ b/src/warnet/ln.py
@@ -7,7 +7,7 @@ from .k8s import (
     get_default_namespace_or,
     get_pod,
 )
-from .process import run_command
+from .process import run_command, stream_command
 
 
 @click.group(name="ln")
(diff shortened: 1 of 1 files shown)


pinheadmz:
Thanks for contribution! And sorry about  the linter, everyone hates it (sigh) see https://github.com/bitcoin-dev-project/warnet/blob/main/docs/developer-notes.md#formatting--linting

Since you're using `stream_command` it's blocking -- which is fine if you only need one at a time, how have you been dealing with simln and (I assume) a cluster of LN nodes? It might make more sense to even spawn a pool of threads for all the nodes, maybe even spit out a config file for simln

f3r10:
yes, you are right it would be better to spawn a pool of threads for all the nodes, I will update that part.

Issue #678, opened 2025-01-29
plugin.py needs shebang or chmod but should really probably just be ingested by python instead of run directly in a shell


pinheadmz:
Error log:

```

Queuing postNetwork plugin command: leaderboard with {'entrypoint': '../../plugins/leaderboard'}
Starting postNetwork plugins
Process Process-5:
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/matthewzipkin/Desktop/work/warnet/src/warnet/process.py", line 7, in run_command
    raise Exception(result.stderr)
Exception: bash: line 1: networks/scrimmage/../../plugins/leaderboard/plugin.py: Permission denied
```

mplsgrant:
fixed by  #689

PR #689, opened 2025-03-06
let python3 run `plugin.py` instead of bash

mplsgrant:
`plugin.py` should just be ingested by python instead doing what we do today which is chmod-ing `plugin.py` and calling it via bash.

Also, when deciding between using `python` and  `python3`, I did some research and `python3` is allegedly the sure-fire way to call python on both linux and macos.

Addresses #678

### DIFF
--- a/src/warnet/deploy.py
+++ b/src/warnet/deploy.py
@@ -166,7 +166,7 @@ def run_plugins(directory, hook_value: HookValue, namespace, annex: Optional[dic
                 }
 
                 cmd = (
-                    f"{network_file_path.parent / entrypoint_path / Path('plugin.py')} entrypoint "
+                    f"python3 {network_file_path.parent / entrypoint_path / Path('plugin.py')} entrypoint "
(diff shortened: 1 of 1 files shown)


src/warnet/deploy.py
```diff
@@ -166,7 +166,7 @@ def run_plugins(directory, hook_value: HookValue, namespace, annex: Optional[dic
                 }
 
                 cmd = (
-                    f"{network_file_path.parent / entrypoint_path / Path('plugin.py')} entrypoint "
+                    f"python3 {network_file_path.parent / entrypoint_path / Path('plugin.py')} entrypoint "
```
pinheadmz: there's a trick to this, see how bitcoin test runner calls python from inside python:

https://github.com/bitcoin/bitcoin/blob/ee68b05f3d639b28b9d23b48324bc5f20ccfe49c/test/functional/test_runner.py#L783

pinheadmz (commented):
concept ACK

pinheadmz:
LGTM, merge when you're happy

PR #694, opened 2025-03-11
Add Local Test Runner Script

macgyver13:
This PR adds a script for executing github/workflow tests to mimic the pipeline testing process

- add local_test_runner script
- add delay to graph_test expect - increases reliability of test

Let me know if this should be named differently, documented or included in a different location.


willcl-ark:


### Conflicts
No conflicts as of last run.

willcl-ark:
Interesting!

Do you know how this compares with e.g. https://github.com/nektos/act (which I've actually never tried in this repo)?

macgyver13:
I was able to get the act system running on macOS silicon using these arguments:
`act --container-architecture linux/arm64 -j test`
`act --container-architecture linux/amd64 -j test`

Each test fails with either of these 2 docker errors depending on architecture (Not sure this will be worth resolving for mac):
```
| * minikube v1.35.0 on Ubuntu 22.04 (docker/arm64)
| * The "docker" driver should not be used with root privileges. If you wish to continue as root, use --force.
| * If you are running minikube within a VM, consider using --driver=none:
| *   https://minikube.sigs.k8s.io/docs/reference/drivers/none/
|
| X Exiting due to DRV_AS_ROOT: The "docker" driver should not be used with root privileges.
```

```
| * minikube v1.35.0 on Ubuntu 22.04 (docker/amd64)
| * Unable to pick a default driver. Here is what was considered, in preference order:
|   - docker: Not healthy: Cannot use amd64 minikube binary to start minikube cluster with Docker driver on arm64 machine
|   - docker: Suggestion: Download and use arm64 version of the minikube binary <https://minikube.sigs.k8s.io/docs/start/>
| * Alternatively you could install one of these drivers:
|   - kvm2: Not installed: exec: "virsh": executable file not found in $PATH
|   - podman: Not installed: exec: "podman": executable file not found in $PATH
|   - qemu2: Not installed: exec: "qemu-system-x86_64": executable file not found in $PATH
|   - virtualbox: Not installed: unable to find VBoxManage in $PATH
|
| X Exiting due to DRV_NOT_HEALTHY: Found driver(s) but none were healthy. See above for suggestions how to fix installed drivers.
```

test/graph_test.py
```diff
@@ -39,21 +40,29 @@ def directory_exists(self):
             self.sut = pexpect.spawn("warnet create")
             self.sut.expect("name", timeout=10)
             self.sut.sendline("ANewNetwork")
+            time.sleep(0.1)
```
pinheadmz: Why do you need all these sleeps?

pinheadmz: (doesnt the pexpect module handle its own waiting?)

test/local_test_runner.py
```diff
@@ -0,0 +1,69 @@
+import subprocess
```
pinheadmz, line 2: you can chmod a+x this file and commit those permissions

pinheadmz: #! env python
probably also needed

pinheadmz:
Besides these nits, concept ACK and tested ACK - I tried to run the script on a slow underpowered VM i had laying around and it did run two tests successfully in series then failed which is not your fault ;-)

```
Exception: Timed out waiting for Truth from predicate: def check_scenarios():
            scns = scenarios_deployed()
            if len(scns) == 0:
                return True
            for s in scns:
                exit_status = get_pod_exit_status(s["name"], s["namespace"])
                self.log.debug(f"Scenario {s['name']} exited with code {exit_status}")
                if exit_status != 0:
                    return False
            return True
******** testing failed

```

test/graph_test.py
macgyver13: I did a 10 test sample with and without time.sleep, 30% failure in both cases so I am going to reverse the graph_test.py change

test/local_test_runner.py
macgyver13: I am unable to execute ./test/local_test_runner.py <test> without using `#!/usr/bin/env python` Do you want me to use that or is `#! env python` some general approach that should work?

pinheadmz: sorry do what works - my comment is garbage

pinheadmz:
LGTM @willcl-ark any additional thoughts?

Issue #699, opened 2025-03-26
`warnet auth` error with existing kube config


pinheadmz:
existing file, not sure its origin but this is on ubuntu using minikube.

```
$ cat ~/.kube/config
apiVersion: v1
clusters: null
contexts: null
current-context: ""
kind: Config
preferences: {}
users: null
```

warnet auth fails trying to "merge" the incoming warnet auth file (for battle of galen erso) with this existing file (I think)

```
Traceback (most recent call last):
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/bin/warnet", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stacie/Documents/bitcoin-dev-project/warnet/MIT_Hackathon_Prep/.venv/lib/python3.12/site-packages/warnet/users.py", line 44, in auth
[...]

PR #700, opened 2025-03-26
user auth: check for None, replace with []

mplsgrant:
Check for None in kubeconfig and replace it with [] in order to avoid #699

### DIFF
--- a/src/warnet/users.py
+++ b/src/warnet/users.py
@@ -41,6 +41,8 @@ def auth(auth_config):
     if not is_first_config:
         for category in ["clusters", "users", "contexts"]:
             if category in auth_config:
+                if not base_config[category]:
+                    base_config[category] = []
                 merge_entries(category, base_config, auth_config)
 
     new_current_context = auth_config.get("current-context")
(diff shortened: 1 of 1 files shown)


pinheadmz:
LGTM @satsie if you're able to reproduce the issue again (resetting that stub kubeconfig file?) and verify this PR fixes the issue that would be really great. 

Either way we should merge this and cut a release before the MIT event this week.

src/warnet/users.py
```diff
@@ -41,6 +41,8 @@ def auth(auth_config):
     if not is_first_config:
         for category in ["clusters", "users", "contexts"]:
             if category in auth_config:
+                if not base_config[category]:
```
macgyver13: This will generate a KeyError exception if category is not in base_config right?

`if category in base_config and not base_config[category]` should work

mplsgrant: Nice! That works smoothly. Thank you.

pinheadmz (approved):
tested ACK thank you!

PR #713, opened 2025-04-23
Add CLN support

macgyver13:
Replaces #685 (Continues work started by Dan, but only focuses on CLN support at this time, he faced stability issues with eclair, will revisit that integration in the future)

This PR adds CLN pod support to warnet and SimLN plugin. Noteable changes:
1. SimLN plugin [copies](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-be478977151fc21dd7b931e628451e54dc2ecd34dfe1f05c80c1f298dcec48b3) certs/key from source cln pod to simln pod
2. added sleep(0.25) before most thread.start() to greatly increase reliability of ln_init.py
3. [deploy.py](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-57566e58fe1069e9fc91b8a84d803037bbac59f8bc0bcbf35c9bdaaff1a56d54) now searches node-defaults.yaml to determine if ln_init is required


test/data/LN_10.json
```diff
@@ -987,7 +987,7 @@
         {
             "node1_pub": "02b705ad9fff5b30e69dd3810d100372332c1a750db5a50edf7353c77ab486643e",
             "node2_pub": "037659a0ac8eb3b8d0a720114efc861d3a940382dcfa1403746b4f8f6b2e8810ba",
-            "channel_id": "891818279431438336",
+            "channel_id": "908655100992880641",
```
pinheadmz: You shouldn't have to alter this json file -- it's an example of input derived from a chunk of mainnet.

src/warnet/graph.py
```diff
@@ -269,7 +269,7 @@ def _import_network(graph_file_path, output_path):
         source = pk_to_tank[edge["node1_pub"]]
         channel = {
             "id": {"block": block, "index": index},
-            "target": pk_to_tank[edge["node2_pub"]] + "-ln",
+            "target": pk_to_tank[edge["node2_pub"]] + "-lnd",
```
pinheadmz: Is the change to `-lnd` required? My thinking originally was that `-ln` could apply to any implementation. There's other kubernetes-y ways of labeling pods with what specific LN client they are using, for the sake of interop.

test/data/LN_10.json
pinheadmz: Just reverting these channel IDs, and `ln_test` passed locally for me

pinheadmz (commented):
Starting to review this, will add more comments and address your questions once ive got my head around it

test/data/LN_10.json
macgyver13: I agree this change was made before I realized that adding a .25 delay in the thread start addressed the underlying issue - I will reverse this.

src/warnet/graph.py
macgyver13: Good question - The answer is yes it is possible to use another attribute to make the decision on how to communicate with the node.  Pretty much anywhere you see a `"-cln" in` reference I would need to make the type determination a different way.
I will come up with a plan for this.

macgyver13: Resolved in commit 10531d5

resources/charts/bitcoincore/charts/cln/values.yaml
```diff
@@ -0,0 +1,114 @@
...
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+livenessProbe:
+  exec:
+    command:
+      - "/bin/sh"
+      - "-c"
+      - "lightning-cli --network=regtest getinfo >/dev/null 2>&1"
```
pinheadmz: we can't hard code a network because warnet also runs on custom signet chains. for lnd we just look for a PID

resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
```diff
@@ -23,7 +23,7 @@ If release name contains chart name it will be used as a full name.
 Expand the name of the chart.
 */}}
 {{- define "lnd.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}-ln
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
```
pinheadmz: these `-ln` should be kept, too. Otherwise there will be inconsistencies in the meta data:

```
app=tank-0002-ln
app.kubernetes.io/instance=tank-0002
```

resources/charts/commander/templates/rbac.yaml
```diff
@@ -15,8 +15,8 @@ metadata:
     app.kubernetes.io/name: {{ .Chart.Name }}
 rules:
   - apiGroups: [""]
-    resources: ["pods", "configmaps"]
-    verbs: ["get", "list", "watch"]
+    resources: ["pods", "configmaps", "pods/exec"]
+    verbs: ["get", "list", "watch", "exec"]
```
pinheadmz: I *think* this is probably fine, commanders in a wargame shouldnt have any adversarial access to pods, but I think ultimately we should take advantage of the cln rest interface, and send our commands that way instead of executing the cli tool inside the pod (more on that in ln.py ...)

resources/scenarios/ln_framework/ln.py
```diff
@@ -73,41 +110,272 @@ def to_lnd_chanpolicy(self, capacity):
...
+        self.impl = "cln"
+
+    def rpc(
+        self,
+        method: str,
+        params: list[str] = None,
+        namespace: Optional[str] = "default",
+        max_tries=5,
+    ):
+        cmd = ["lightning-cli", method]
```
pinheadmz: I'd prefer to see a REST interface here, like we do for lnd:

https://docs.corelightning.org/reference/post_rpc_method_resource

We should be able to store the "rune" in a config map and retrieve it when commander.py is run. Lemme know if you've already looked into this rabbit hole, or how dark it is down there...

resources/plugins/simln/charts/simln/values.yaml
```diff
@@ -4,6 +4,8 @@ image:
   tag: "0.2.3"
   pullPolicy: IfNotPresent
 
+restartPolicy: Never
```
pinheadmz: why ?

resources/plugins/simln/plugin.py
```diff
@@ -179,6 +187,54 @@ def _generate_activity_json(activity: Optional[list[dict]]) -> str:
     return json.dumps(data, indent=2)
 
 
+def transfer_cln_certs(name):
+    dst_container = "init"
+    cln_root = "/root/.lightning/regtest"
```
pinheadmz: avoid hard coded network name here if possible

resources/plugins/simln/plugin.py
```diff
@@ -179,6 +187,54 @@ def _generate_activity_json(activity: Optional[list[dict]]) -> str:
...
+                ln_name,
+                "cln",
+                f"{cln_root}/client-key.pem",
+                name,
+                dst_container,
+                f"/working/{ln_name}-client-key.pem",
+            )
+
+
+def copyfile(pod_name, src_container, source_path, dst_name, dst_container, dst_path):
```
pinheadmz: seems like this entire function can go in k8s.py along with the read_ and write_ utils.

resources/scenarios/commander.py
```diff
@@ -132,7 +158,7 @@ def tank_connected(self, tank):
                 if count >= tank.init_peers:
                     break
                 else:
-                    sleep(1)
+                    sleep(5)
```
pinheadmz, line 161: how come ?

resources/scenarios/commander.py
```diff
@@ -92,6 +94,30 @@ def auth_proxy_request(self, method, path, postdata):
 AuthServiceProxy._request = auth_proxy_request
 
 
+# Create a custom formatter
+class ColorFormatter(logging.Formatter):
```
pinheadmz, line 98: wow really nice! 🚀

resources/scenarios/commander.py
```diff
@@ -203,7 +228,8 @@ def setup(self):
             self.tanks[tank["tank"]] = node
 
         for ln in WARNET["lightning"]:
-            self.lns[ln] = LND(ln)
+            ln.setLogger(self.log)
```
pinheadmz: this is cool, i think (maybe in Class LNNode) each ln process should identify itself in the log, using `getLogger(name)` or whatever it is, otherwise you can't tell who threw the error:

```
LNInit   - LN Channel {'capacity': 50001, 'id': {'block': 300, 'index': 4}, 'push_amt': 25003, 'target': 'tank-0003-ln', 'source': 'tank-0001-ln'} not available, removing

```

resources/scenarios/ln_framework/ln.py
```diff
@@ -11,6 +18,36 @@
 INSECURE_CONTEXT.verify_mode = ssl.CERT_NONE
 
 
+# execute kubernetes command
+def run_command(name, command: list[str], namespace: Optional[str] = "default") -> str:
```
pinheadmz: yeah seems overkill to need this, I really hope we can cln to work via REST

resources/scenarios/ln_framework/ln.py
```diff
@@ -73,41 +110,272 @@ def to_lnd_chanpolicy(self, capacity):
...
+    @abstractmethod
+    def graph(self) -> dict:
+        pass
+
+    @abstractmethod
+    def update(self, txid_hex: str, policy: dict, capacity: int) -> dict:
+        pass
+
+
+class CLN(LNNode):
```
pinheadmz, line 133: Really great work on this, thanks. this is the real task!

resources/scenarios/ln_init.py
```diff
@@ -40,28 +45,24 @@ def gen(n):
         ##
         self.log.info("Getting LN wallet addresses...")
         ln_addrs = []
+        ln_nodes = []
```
pinheadmz: what was wrong with `self.lns.values()`?

resources/scenarios/ln_init.py
```diff
@@ -135,13 +134,19 @@ def get_ln_uri(self, ln):
...
-            connections.append((node, prev_node))
+            if node != prev_node:
+                connections.append((node, prev_node))
             prev_node = node
         # Explicit connections between every pair of channel partners
         for ch in self.channels:
+            node_names = self.node_names(ln_nodes)
+            if ch["source"] not in node_names or ch["target"] not in node_names:
+                self.log.error(f"LN Channel {ch} not available, removing")
+                self.channels.remove(ch)
```
pinheadmz: this is more of a user input error isn't it? a channel is defined without both a source and target? I wonder if that should be a critical error, because the user is probably not going to get what they intended.

resources/scenarios/ln_init.py
```diff
@@ -6,6 +6,8 @@
 from commander import Commander
 from ln_framework.ln import Policy
 
+THREAD_JOIN_TIMEOUT = 20
```
pinheadmz: Im not sure I agree with this, for example if a channel doesn't open within 20 seconds, the scenario will just silently continue, and the user won't have the network graph they wanted. Were you finding some operations to hang longer than 20 seconds?

resources/scenarios/ln_init.py
```diff
@@ -135,13 +134,19 @@ def get_ln_uri(self, ln):
         # (source: LND, target_uri: str) tuples of LND instances
         connections = []
         # Cycle graph through all LN nodes
-        nodes = list(self.lns.values())
+        nodes = list(ln_nodes)
         prev_node = nodes[-1]
         for node in nodes:
-            connections.append((node, prev_node))
+            if node != prev_node:
```
pinheadmz: why is this needed? how would `==` occur?

resources/scenarios/ln_init.py
```diff
@@ -188,7 +199,9 @@ def connect_ln(self, pair):
         # so their channel ids are deterministic
         ch_by_block = {}
         for ch in self.channels:
-            # TODO: if "id" not in ch ...
+            if "id" not in ch or "block" not in ch["id"]:
+                self.log.info(f"LN Channel {ch} not found")
+                continue
```
pinheadmz: I think this is another place where we shouldn't just silently fail. User wants graph, user gets graph or we abort.

In fact, I'm wondering why you had to make so many changes to `ln_init.py` at all. Everything lnd-specific *should* have been already abstracted out. What would fail on your branch if this file was totally restored? Maybe if you have ln_init optimizations they can go in a separate PR?

src/warnet/deploy.py
```diff
@@ -365,11 +365,22 @@ def deploy_network(directory: Path, debug: bool = False, namespace: Optional[str
...
+                break
+        if needs_ln_init:
             break
 
+    default_file_path = directory / DEFAULTS_FILE
+    with default_file_path.open() as f:
+        default_file = yaml.safe_load(f)
+    if any(default_file.get("ln", {}).get(key, False) for key in supported_ln_projects):
+        needs_ln_init = True
+
```
pinheadmz, line 382: yes, good save thanks.

src/warnet/ln.py
```diff
@@ -46,9 +49,13 @@ def pubkey(
...
 
-def _pubkey(pod: str):
-    info = _rpc(pod, "getinfo")
-    return json.loads(info)["identity_pubkey"]
+def _pubkey(pod_name: str):
+    info = _rpc(pod_name, "getinfo")
+    pod = get_pod(pod_name)
+    pubkey_key = "identity_pubkey"
+    if "cln" in pod.metadata.labels["app.kubernetes.io/name"]:
+        pubkey_key = "id"
```
pinheadmz, line 57: this kind of thing is ok for now and probably will be ok for eclair as well -- but its a tad codesmelly. feels like implementation-specific methods should stay in their classes, even as jsut static functions

test/data/ln_mixed/network.yaml
```diff
@@ -0,0 +1,66 @@
+nodes:
+  - name: tank-0001
+    addnode:
+      - tank-0003
+    ln:
+      cln: true
```
pinheadmz: just wondered what would happen in the edge case

```
ln:
  cln: true
  lnd: true
```

test/ln_mixed_test.py
```diff
@@ -0,0 +1,125 @@
...
+        # pay from lightning
+        self.log.info(self.warnet(f"ln rpc {sender} payinvoice -f {inv['bolt11']}"))
+
+        def wait_for_success():
+            return self.node(recipient).channelbalance() == init_balance + 1000
+
+        self.wait_for_predicate(wait_for_success)
+
+    # def pay_invoice_node(self, sender: str, recipient: str):
+    #     print("pay invoice using ln framework")
```
pinheadmz: could just delete the unused pasted stuff, or are you going to implement it as well?

I guess, similar to my comment about `ln_init.py`, the existing ln tests should "just work" on any network, whether it be all-lnd or lnd-cln mixed.

pinheadmz (commented):
really really REALLY great work, thank you. I reviewed all the changes and left comments throughout. 

Next I'm going to try answer your open questions in the PR description, and then try to figure out whats up with the test failures.

resources/scenarios/ln_framework/ln.py
macgyver13: I ultimately abandoned REST with CLN for stability reasons not access. It has been a few weeks. Let me circle back now that I have a better picture and report any findings.

resources/plugins/simln/charts/simln/values.yaml
macgyver13: Made diagnostics easier. Not required. I'll remove.

pinheadmz:
Now that i've read through the entire patch i think i might have opinions about these questions, lets see...

----

>    [commander/templates/rbac.yaml](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-98d066ccb4fd2a9b28f0ff032aa4849d91f9e579bff6c1808e0482333de5d69e) now requires exec rule

I think you only needed this to use `lightning-cli` inside cln containers, and im hoping we dont need to do that, but this isn't a bad change necessarily

----

>    SimLN plugin [copies](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-be478977151fc21dd7b931e628451e54dc2ecd34dfe1f05c80c1f298dcec48b3) certs/key from source cln pod to simln pod

yes I think this is the only way to do it for now

----

>   [ ln_framework/ln.py::run_command](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-832df137c46d9c31e91ff9b8bc15a1752b1173a74b670d0b91d82354dcf73344) has to determine if being executed inside or outside cluster to make connection to cln node correctly via kubernetes.stream

everything inside `scenarios/` should only ever be executed inside the cluster. The only exception is if a user needs help: `warnet run ... -- --help` which requires the scenario file to be run locally just to gather the argument parser comments.

----

>    added sleep(0.25) before most thread.start() to greatly increase reliability of ln_init.py

Not sure why you needed this? What happens otherwise?

----

>    [deploy.py](https://github.com/bitcoin-dev-project/warnet/compare/main...macgyver13:warnet:add-ln-nodes#diff-57566e58fe1069e9fc91b8a84d803037bbac59f8bc0bcbf35c9bdaaff1a56d54) now searches node-defaults.yaml to determine if ln_init is required

yes thank you yes!

----

>    lnd nodes need to be indexed in the same block before cln nodes, otherwise the expected channel id will not be correct

Hm I didn't notice this. Did you see how we use decreasing transaction fees in the channel opens to ensure they are in the expected order in a block? Hopefully when you open a channel from cln you can also set the fee rate, and the process will continue to be implementation-agnostic

----

> There is a new test ln_mixed_test.py that is partially implemented until the following decision is determined - how should LNNode access LND nodes from testing instance outside of cluster?:

>    port forward? - seems like a lot of overhead / clunky
>    use rpc instead of REST? - duplicates warnet/ln.py
>    use kubectl to issue curl commands? - messy / clunky
>    use simln plugin which will act inside cluster avoiding all the issues? - a drawback for simln is the activity is sendkey based instead of invoice / pay

I don't think this should ever happen - was it happening before? Scenarios are executed entirely inside the cluster.

resources/scenarios/ln_framework/ln.py
pinheadmz: oh interesting, let me know more about that and i'll look in to it as well. or we can ping cln contributors?

test/data/ln_mixed/network.yaml
macgyver13: I bet helm would add both charts 😬

resources/scenarios/ln_framework/ln.py
pinheadmz: I think this is why the wargames test is failing - it tries to run a scenario with limited permission and is crashing here:

```

Failed to deploy scenario commander: generate_one_allnodes
Error: Error: 2 errors occurred:
        * roles.rbac.authorization.k8s.io "commander-generateoneallnodes-1745521172" is forbidden: user "system:serviceaccount:wargames-red:warnet-user"
(groups=["system:serviceaccounts" "system:serviceaccounts:wargames-red" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["configmaps"], Verbs:["exec"]}
{APIGroups:[""], Resources:["pods"], Verbs:["exec"]}
{APIGroups:[""], Resources:["pods/exec"], Verbs:["list" "watch" "exec"]}
        * roles.rbac.authorization.k8s.io "commander-generateoneallnodes-1745521172" not found


```

resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
macgyver13: which name do you want to match?  This is what I get with the pre PR setup...
app.kubernetes.io/name=lnd-ln
meta.helm.sh/release-name: tank-0003
app=tank-0003-ln
Name:             tank-0003-ln
app.kubernetes.io/instance=tank-0003

I was thinking this makes most sense (just an opinion):
app.kubernetes.io/name=lnd
meta.helm.sh/release-name: tank-0003-ln
app=tank-0003-ln
Name:             tank-0003-ln
app.kubernetes.io/instance=tank-0003

pinheadmz: it's the `release-name` i guess but actually it's not logging where i thought it was logging, so it might not be relevant at all. did you change it here for a reason?

resources/scenarios/commander.py
macgyver13: Now: tank-0001-ln - INFO: I have an issue

resources/scenarios/commander.py
macgyver13: I was trying to cut down on frequency of these messages?  Is there any reason to state this so often?
LNInit   - Tank tank-0005 connected to 0/1 peers
LNInit   - Tank tank-0004 connected to 0/1 peers
LNInit   - Tank tank-0002 connected to 0/1 peers
LNInit   - Tank tank-0001 connected to 0/1 peers
LNInit   - Tank tank-0003 connected to 0/1 peers

pinheadmz: oh I gotcha - no its fine then

resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
macgyver13: I saw an opportunity to standardize the string assembly when updating cln template, I'm good with whatever you want here

pinheadmz: ok lets leave it with the `-ln` then just in case there is some other place where we need to distinguish between `tank-0000` and `tank-0000-ln`

resources/scenarios/ln_init.py
macgyver13: I'll revert the safeguards I added and they can go in a future PR.

resources/scenarios/ln_init.py
macgyver13: reverting to lns.values() and not managing "responding" nodes

resources/scenarios/ln_init.py
macgyver13: removed

resources/scenarios/ln_init.py
macgyver13: removed

resources/scenarios/ln_init.py
macgyver13: removed

resources/scenarios/ln_framework/ln.py
macgyver13: Using REST with CLN nodes - works - proceeding with acquiring RUNE from nodes

resources/charts/commander/templates/rbac.yaml
macgyver13: removed

resources/scenarios/ln_framework/ln.py
macgyver13: removed

resources/plugins/simln/plugin.py
macgyver13: moved

resources/scenarios/ln_framework/ln.py
macgyver13: removed

test/ln_mixed_test.py
macgyver13: removed ln_mixed

test/data/ln_mixed/network.yaml
macgyver13: tested this, lnd started when both were defined

pinheadmz:
@macgyver13 is this still intended to be a draft? What's next?

macgyver13:
> @macgyver13 is this still intended to be a draft? What's next?

Updating documentation and PR description to match current state. 
I don't plan to add more for testing or CLN channel policy updates in this PR.

pinheadmz (approved):
Woooo big one! Thank you!

PR #732, opened 2025-07-25
Improve installation docs

cndolo:
- First commit corrects the path to the venv in the top-level `README`.
- Second commit changes the path to the entrypoint to a perhaps more generic one. The reason for that is that the given relative path to the simln plugin might not always be right (it wasn't for me, at least). While at it, we also correct the spelling of simln to be consistent with the rest.


pinheadmz:
Thanks!

PR #737, opened 2025-08-13
Prepare LN for large networks

pinheadmz:
1. Move all the try/catch/retry logic from `ln.py` to `ln_init.py`. Meaning the API calls have no error handling on their own and propagate back up to the caller. This is to ensure consistency in the `ln_init` scenario which is opinionated about what is allowed to fail. Ideally the scenario completes successfully, terminates with an error, or runs endlessly because something is broken. In all those cases, `ln_init` needs to directly handle all API calls to LN nodes.
2. Refactor `ln_init` for large networks (specifically `test/data/LN_100.json` with >500 channels) ensuring LN nodes always have enough funds to open all their channels, that channel opens are always tx output 0 and channel open TXs are ordered as expected in their blocks.


pinheadmz:
@macgyver13 I touched a lot of your LN API code, lemme know if you see anything smelly or frustrating in here,,,

pinheadmz:
Tested with 10, 50, 100 node networks + simln: https://github.com/pinheadmz/ln100-warnet

On digital ocean cluster: 100 vCPUs / 200 GB Memory / 1.22 TB Disk

macgyver13:
> @macgyver13 I touched a lot of your LN API code, lemme know if you see anything smelly or frustrating in here,,,

Concept ACK - I like the separation of concerns, nice work - much needed clean up 👏 
I am a little surprised the frequent use of reset_connection in CLN and LND - was that required for stability?

I will test later this week and even consider testing eclair with these changes to see if an underlying issue was causing instability with eclair nodes in warnet.

There are some nice improvements in ln.py in the eclair PR that may make sense to follow this PR if you see value - around the "use_rpc" concept.

pinheadmz:
> reset_connection

Well, the way it was before we only reset connection if there was a problem and my new idea is, ln.py doesn't deal with problems at all. I also had to deal with this already in scenarios based on the functional test framework:

https://github.com/bitcoin-dev-project/warnet/blob/e4ed4b6b540c7fafe728908b90695901ed7b4adb/resources/scenarios/commander.py#L87-L94

Unlike the test framework, we can't expect an HTTP connection to survive the entire lifespan of a warnet, so we refresh the connection on every call.

src/warnet/deploy.py
```diff
@@ -204,6 +204,8 @@ def check_logging_required(directory: Path):
         return True
     if default_file.get("metricsExport", False):
         return True
+    if default_file.get("lnd", False).get("metricsExport"):
```
macgyver13: I think you need to use {} for the default value
if default_file.get("lnd", {}).get("metricsExport"):

src/warnet/deploy.py
```diff
@@ -216,6 +218,8 @@ def check_logging_required(directory: Path):
             return True
         if node.get("metricsExport", False):
             return True
+        if node.get("lnd", False).get("metricsExport"):
```
macgyver13: ^^ same problem

src/warnet/deploy.py
pinheadmz: yep thanks! Fixed on next push

pinheadmz:
merging for now so i can move on to the next task

rastavipx123:
[_r17_-transcript.txt](https://github.com/user-attachments/files/23829220/_r17_-transcript.txt)

PR #742, opened 2025-09-17
Support extra services in caddy / Warnet dashboard

### LINKED ISSUE #679: plugins should be able to add more websites to caddy
For example a leaderboard plugin, could be added with `preDeploy` that adds a webserver pod that has a certain label. Then when the caddy chart is deployed, the cluster is checked for labeled pods, and those are added to `caddyConfig` and `htmlConfig` in `caddy/values.yaml`

pinheadmz:
Closes https://github.com/bitcoin-dev-project/warnet/issues/679

Abstracts the fork observer and grafana services that were hard coded into the caddy chart. Also adds any extra configured reverse proxy. The test simply adds the REST interface from one of the tanks by adding this to network.yaml:

```yaml
services:
  - title: Ringo REST
    path: /ringo/
    host: ringo.default
    port: 18443
```

In conjunction with a plugin, any additional web services can now be added to a warnet network

### DIFF
--- a/resources/charts/caddy/templates/configmap.yaml
+++ b/resources/charts/caddy/templates/configmap.yaml
@@ -6,6 +6,34 @@ metadata:
     {{- include "caddy.labels" . | nindent 4 }}
 data:
   Caddyfile: |
-    {{- .Values.caddyConfig | nindent 4 }}
+    :80 {
+      respond /live 200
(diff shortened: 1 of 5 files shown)


pinheadmz:
Screenshot 2025-09-19 at 11 30 12 AM

PR #743, opened 2025-09-23
Accomdate LN nodes in "wargames" namespaces

pinheadmz:
Sending RPCs from `warnet ln rpc` and also sending commands to ln nodes from scenarios, we need to include the namespace of the node in the domain name / URI, so that an `--admin` scenario can get addresses from tanks like `armada-1-ln.wargames-red` and send funds to it from tanks like `miner.default`


pinheadmz:
Screenshot 2025-09-23 at 10 15 52 AM

PR #758, opened 2025-11-12
Bump test framework to v30.0

davidgumberg:
Bumps the included test framework to Bitcoin Core 30.0, and update example scenarios. Presently, bumping and fixing the tests happen in separate commits for ease of review, but I'm happy to squash if commits being atomic is preferred.

This PR also includes an unrelated change I found helpful: Enabling CI tests on all branches so contributors to warnet will have CI run on their branches without having to open a PR against `main`. 

Opening as a draft as this is still failing CI at the moment.


resources/scenarios/ln_init.py
```diff
@@ -511,7 +511,7 @@ def matching_graph(self, expected, ln):
 
 
 def main():
-    LNInit().main()
+    LNInit(__file__).main()
```
pinheadmz: Instead of adding this to every scenario, and since we don't use the value anyway, why not just add a dummy to Commander constructor:

```py
class Commander(BitcoinTestFramework):
    def __init__(self):
        super().__init__(__file__) # or None
```

pinheadmz (commented):
concept ACK, did quick review in the browser. I'll dig in to the CI failures next.
THANKS!

resources/scenarios/ln_init.py
davidgumberg: Right, that's much better, I've taken this.

davidgumberg: Update: Ah, I've just remembered why I did it this way: https://github.com/bitcoin/bitcoin/blob/v30.0/test/functional/test_framework/test_framework.py#L134-L136:

```python
class BitcoinTestMetaClass(type):
    """Metaclass for BitcoinTestFramework.

    Ensures that any attempt to register a subclass of `BitcoinTestFramework`
    adheres to a standard whereby the subclass overrides `set_test_params` and
    `run_test` but DOES NOT override either `__init__` or `main`. If any of
    those standards are violated, a ``TypeError`` is raised."""

    def __new__(cls, clsname, bases, dct):
        if not clsname == 'BitcoinTestFramework':
            if '__init__' in dct or 'main' in dct:
                raise TypeError("BitcoinTestFramework subclasses may not override "
                                "'__init__' or 'main'")
```

pinheadmz:
I fixed the CI failures on main and rebased. Then added a few more commits to accomdate upstream changes to the test framework. This is a breaking change and Galen Erso and Wrath of Nalo will need to either be rebased or pin specific older versions of warnet (see #763)

pinheadmz (approved):
```
--> diff ~/Desktop/work/bitcoin/test/functional/test_framework/ ~/Desktop/work/warnet/resources/scenarios/test_framework/
```

Diff is a match, CI passes. LFG, worry about breaking changes later. Main should be considered unstable now

PR #766, opened 2026-01-17
Check if minikube tunnel is running in setup

ekzyis:
Hello!

In a minikube environment, [`bash scripts/deploy.sh regtest4` in wrath-of-nalo](https://github.com/bitcoin-dev-project/wrath-of-nalo/blob/aea397cfdb8f05b3b09f31f44c16f40167e60104/scripts/deploy.sh) depends on `minikube tunnel` running. The same is true for `warnet dashboard`.

However, unlike `warnet dashboard`, scripts/deploy.sh (specifically [update_grafana_dashboard.py](https://github.com/bitcoin-dev-project/wrath-of-nalo/blob/aea397cfdb8f05b3b09f31f44c16f40167e60104/scripts/upload_grafana_dashboard.py)) does not return a useful error if the ingress IP address was not found. It prints this error instead:

> Error getting ingress IP: 'NoneType' object is not subscriptable

and then tries to resolve `None` as a domain. Compare this to how `warnet dashboard` mentions this in that case:

> Error: Could not get the IP address of the dashboard
> If you are running Minikube please run 'minikube tunnel' in a separate terminal

I didn't want to duplicate this error handling to update_grafana_dashboard.py. Instead, I thought it'd be nice to let `warnet setup` also handle this, similar to how it's already checking if minikube is running.

This is what this PR is doing! The check depends on `pgrep` with the `-f` option being available.

Here's a video of how I tested this:

https://github.com/user-attachments/assets/25efe0f8-024c-4c48-a4ed-36a060faaff2

I will include suggestions for my own code in a review. This PR is just meant to show the least invasive change to handle this in `warnet setup`, assuming that's even the best place to handle it.

### DIFF
--- a/src/warnet/project.py
+++ b/src/warnet/project.py
@@ -80,6 +80,21 @@ def setup():
             # Minikube command not found
             return False, ""
 
+    def is_minikube_tunnel_running() -> tuple[bool, str]:
+        try:
+            result = subprocess.run(
+                ["pgrep", "-f", "minikube tunnel"],
+                capture_output=True,
+                text=True
+            )
+            if result.returncode == 0:
+                return True, "minikube tunnel is running"
+            else:
+                return False, ""
+        except FileNotFoundError:
+            # pgrep command not found
+            return False, "command not found: pgrep"
+
     def is_docker_running() -> tuple[bool, str]:
         try:
             result = subprocess.run(
@@ -292,6 +307,12 @@ def setup():
         install_instruction="Please make sure minikube is running",
         install_url="https://minikube.sigs.k8s.io/docs/start/",
     )
+    minikube_tunnel_running_info = ToolInfo(
+        tool_name="Running Minikube Tunnel",
+        is_installed_func=is_minikube_tunnel_running,
+        install_instruction="Please make sure minikube tunnel is running",
+        install_url="https://minikube.sigs.k8s.io/docs/commands/tunnel/"
+    )
     kubectl_info = ToolInfo(
         tool_name="Kubectl",
         is_installed_func=is_kubectl_installed_and_offer_if_not,
@@ -355,6 +376,7 @@ def setup():
                 if is_platform_darwin():
                     check_results.append(check_installation(minikube_version_info))
                 check_results.append(check_installation(minikube_running_info))
+                check_results.append(check_installation(minikube_tunnel_running_info))
         else:
             click.secho("Please re-run setup.", fg="yellow")
             sys.exit(1)
--- a/src/warnet/project.py
+++ b/src/warnet/project.py
@@ -83,9 +83,7 @@ def setup():
     def is_minikube_tunnel_running() -> tuple[bool, str]:
         try:
             result = subprocess.run(
-                ["pgrep", "-f", "minikube tunnel"],
-                capture_output=True,
-                text=True
+                ["pgrep", "-f", "minikube tunnel"], capture_output=True, text=True
             )
             if result.returncode == 0:
                 return True, "minikube tunnel is running"
@@ -311,7 +309,7 @@ def setup():
         tool_name="Running Minikube Tunnel",
         is_installed_func=is_minikube_tunnel_running,
         install_instruction="Please make sure minikube tunnel is running",
-        install_url="https://minikube.sigs.k8s.io/docs/commands/tunnel/"
+        install_url="https://minikube.sigs.k8s.io/docs/commands/tunnel/",
     )
     kubectl_info = ToolInfo(
         tool_name="Kubectl",


src/warnet/project.py
```diff
@@ -80,6 +80,21 @@ def is_minikube_running() -> tuple[bool, str]:
             # Minikube command not found
             return False, ""
 
+    def is_minikube_tunnel_running() -> tuple[bool, str]:
+        try:
+            result = subprocess.run(
+                ["pgrep", "-f", "minikube tunnel"],
```
ekzyis: Need to check if we can depend on `pgrep -f` being available, or if there's a more common way to check if a process is running on systems that would run `minikube`.

src/warnet/project.py
```diff
@@ -355,6 +376,7 @@ def check_installation(tool_info: ToolInfo) -> ToolStatus:
                 if is_platform_darwin():
                     check_results.append(check_installation(minikube_version_info))
                 check_results.append(check_installation(minikube_running_info))
+                check_results.append(check_installation(minikube_tunnel_running_info))
```
ekzyis, line 377: Maybe this shouldn't be checked independently but inside the check if minikube is running?

That way, we wouldn't always print "Running Minikube is not satisfied" and "Running Minikube Tunnel is not satisfied" together.

src/warnet/project.py
```diff
@@ -80,6 +80,21 @@ def is_minikube_running() -> tuple[bool, str]:
...
             return False, ""
 
+    def is_minikube_tunnel_running() -> tuple[bool, str]:
+        try:
+            result = subprocess.run(
+                ["pgrep", "-f", "minikube tunnel"],
+                capture_output=True,
+                text=True
+            )
+            if result.returncode == 0:
```
ekzyis, line 88: According to `man pgrep`, checking if the exit code is 0 should be enough:

```
EXIT STATUS
       0      One or more processes matched the criteria. For pkill and pidwait, one or more processes must also have been successfully signalled or waited for.
       1      No processes matched or none of them could be signalled.
       2      Syntax error in the command line.
       3      Fatal error: out of memory etc.
```

I have noticed other commands also check the output though. I could do that here, too. That would require `-a`.

pinheadmz:
Hm I'm not so sure we need this. The webserver isn't always deployed in warnet so minkube tunnel wouldnt always be required. I think the error message returned by `warnet dashboard` is still the best way to deal with this.

In Galen Erso, the upload dashboard script isnt that important anyway but youre right its not just an admin thing, if players are testing locally. So I guess we could add `warnet dashboard` to the `deloy` script there.

ekzyis:
Thank you for the feedback, and sorry for the slow reply!

> I guess we could add `warnet dashboard` to the `deploy` script there.

Did you mean to run `warnet dashboard` before uploading the dashboard, so the user could at least see the hint regarding `minikube tunnel` when scrolling up?

```diff
diff --git a/scripts/deploy.sh b/scripts/deploy.sh
index 2254221..9dcb091 100755
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -12,4 +12,5 @@ warnet deploy armadas/$1 --to-all-users
 warnet run scenarios/arm_armada.py --debug --admin
 warnet run scenarios/miner_std.py --tank=miner --admin
 warnet run scenarios/ln_activity.py
+warnet dashboard
 ./scripts/upload_grafana_dashboard.py
\ No newline at end of file
```

This would then also need to be added to the deploy script in wrath-of-nalo (not a problem, just mentioning it).

Another idea would be to move the hint from `warnet dashboard` to `get_ingress_ip_or_host()`. This is the function the dashboard upload scripts also call. I'm not sure why it's part of the `warnet dashboard` command.

ekzyis:
We realized that `wait_for_ingress_controller` only waits for the ingress-nginx-controller pod to be ready, but the pod being ready does not mean that the load balancer is up and has a hostname or IP.[^1] This is what `minikube tunnel` does.

[^1]: I verified this by running `wait_for_ingress_controller` in a python REPL and noticing that it correctly reports the pod as being ready, but the load balancer status is independent of the pod status.

So the dashboard upload scripts in battle-of-garlen-erso and wrath-of-nalo are using `wait_for_ingress_controller` with the expectation that `get_ingress_ip_or_host` will work when it didn't throw, but that's a wrong expectation.

Therefore, instead of this PR, I will create a `wait_for_ingress_endpoint` utility function that will not only wait for the ingress pod to be ready, but also for the load balancer. After a timeout, it will throw.

This function can then be used instead of `wait_for_ingress_controller` in battle-of-garlen-erso and wrath-of-nalo.

Closing!

PR #770, opened 2026-01-22
Print dashboard URL when opening

fridokus:
When running in a VPS or similar, a "default browser" may not be available.

Print the URL so that a user can open it elsewhere if desired.


Mathias-U1F596:
That solves #771.

Mathias-U1F596 (approved):

pinheadmz:
Tested ACK 
thanks!

PR #782, opened 2026-02-10
graph: allow 0 connections in new network

### LINKED ISSUE #780: Create nodes with 0 connections
When creating a network with `warnet init`, selecting “0 connections” should create nodes with no addnode entries. Currently, it always creates at least one connection.
While I understand that it might be intentional to always have one connection, this defeats the purpose of selecting zero connections.
I think it should be possible to create zero connections and have no peers, since that is an available option.
Select “0” for connections

Expected
```
nodes:
  - image:
    tag: '29.0'
    name: tank-0000
```
Got
```
  nodes:
  - addnode:
   - tank-0001
   image:
     tag: '29.0'
   name: tank-0000
```
  
I think we should only create a connection if:

`int(entry["connections"]) > 0`

pinheadmz:
closes https://github.com/bitcoin-dev-project/warnet/issues/780

### DIFF
--- a/src/warnet/graph.py
+++ b/src/warnet/graph.py
@@ -60,24 +60,25 @@ def custom_graph(
                 image = {"tag": entry["version"]}
             node = {"name": f"tank-{index:04d}", "addnode": [], "image": image}
 
-            # Add round-robin connection
-            next_node = (index + 1) % total_count
-            node["addnode"].append(f"tank-{next_node:04d}")
-            connections.add((index, next_node))
-
-            # Add random connections
+            # Connect to other nodes
             available_nodes = list(range(total_count))
+            # Do not connect to self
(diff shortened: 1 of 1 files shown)


pinheadmz:
cc @naiyoma

naiyoma:
tested, this works,  thanks 

```
caddy:
  enabled: true
fork_observer:
  configQueryInterval: 10
  enabled: true
nodes:
- addnode: []
  image:
    tag: '29.0'
  name: tank-0000
- addnode: []
  image:
    tag: '29.0'
  name: tank-0001
- addnode: []
  image:
    tag: '29.0'
  name: tank-0002
- addnode: []
  image:
    tag: '29.0'
  name: tank-0003

```

PR #786, opened 2026-02-13
Added support of lnd seed phrase in the pod config.

randomlogin:
It's possible to set the lnd seed mnemonic in the pod config in the following way:

```
cipherSeedMnemonic: [above, rookie, pioneer, pen, truck, board, spatial, pond, destroy, civil, vote, zebra, coral, love, height, price, nephew, better, bar, wife, chaos, clock, pottery, rifle]
```

If this field is absent, it fallbacks to seed generation via rpc.

For (any) backward compatibility, if the seed is provided in the config, it is also written to `/tmp/genseed.json` (thought without encrypted field).


pinheadmz:
I dont mind _also_ writing the phrase to the data dir in plain text, but I think it should still be generated randomly when the pod starts. I also don't totally understand the use case yet. What do you need it for?

pinheadmz:
Actually I just realized this request has been satisfied inadvertently by https://github.com/bitcoin-dev-project/warnet/pull/776 which saves the seed to disk during startup to make wallet init more reliable:

```sh
(.venv) --> k exec armada-1-ln -- cat /tmp/genseed.json
{"cipher_seed_mnemonic":["abstract", "pigeon", "hammer", "kiss", "seed", "square", "story", "sure", "pen", "cigar", "escape", "fee", "involve", "lava", "lend", "wrist", "kind", "palm", "enough", "carbon", "heavy", "uncle", "illegal", "hungry"], "enciphered_seed":"APSVoj1sL6b1ptCiZR00Kkdi+5//83qT6SuRNqfZHEN6"}
```

I'll add this command to the hints in Wrath of Nalo.

Issue #655, opened 2024-10-16
warnet windows support


2Fast2BCn:
I'm trying to get warnet to work on my windows (with latest docker desktop installed).
I followed the install tutorial.
Then, I'm running: `.\.venv\Scripts\Activate.ps1` in powershell
Followed by
```
warnet --help
Usage: warnet [OPTIONS] COMMAND [ARGS]...
...
```
all good so far, even `warnet status` tells me:
```
Total Tanks: 0 | Active Scenarios: 0
Network connected
```

but then I try to deploy my own project folder `twn` with `warnet deploy .\network\networks\twn\` and I get:
```
[WinError 2] The system cannot find the file specified
It looks like kubectl is not installed. Please install it to continue: https://kubernetes.io/docs/tasks/tools/
```

however in the same shell I can run 
```
(.venv) PS C:\development\warnet> kubectl
kubectl controls the Kubernetes cluster manager.
```

willcl-ark:
Thanks for this report.

I don't know much about windows, but I am wondering if this works under WSL? (or is this how you were trying to run it here?)

Jhoyola:
I have exactly same situation on windows 11. Tried to run with git bash and the deploy command says _It looks like kubectl is not installed. _ even tough I have kubernetes running and `kubectl` command works fine on git bash in same dir.

Jhoyola:
It is definetly about the `subprocess.run()` that is used to run kube commands having `shell=True` and windows path to git bash having space in it (/Program Files/). Also windows PATH variable might return variable results for "bash".

There are multiple locations in the code that assume unix style paths, so I think it's not at all trivial to add full Windows support.

On WSL seems to work for the parts that I tested so far: creating tanks, and seeing their logs and communications.

pinheadmz:
@Jhoyola are you able to work on this? I have a windows machine to test but its not a platform I develop on day-to-day.

Jhoyola:
@pinheadmz I can try to find time to do some work on it.

Jhoyola:
Actually for me it works quite ok so far on WSL, and full Windows compatibility seems a bit difficult actually. So I'm not going to priorize this now.

PR #787, opened 2026-02-17
Node data persistence in persistent volume claims

Jhoyola:
At Torq / Mion we intend to move to using warnet for persistent regtest lightning network for development locally. This requires to have the nodes keep their state indefinetly, to not need to constantly recreate the network, connect nodes to our software etc. I tried to do it by defining `volumes` and `volumeMounts`, but for bitcoin core the _.bitcoin_ directory is already mounted so it conflicted on making a new mount to the same dir. This PR aims to add built-in, easy and configurable persistence option.

### Overview

- Uses persistent volume claims. Either auto generated or pre-defined
- A new helm configuration value set "persistence"
  - Simply add `enabled: true` to create automatic PVC for the node
  - Use `existingClaim: <existing-pvc-name>` to use pre-created PVC instead of creating new one
  - If not set, everything should work as it did before
- Down command prompts to delete the PVCs if they exist
- Added documentation on how to use this

Tested in Windows WSL environment.

### DIFF
--- a/docs/persistence.md
+++ b/docs/persistence.md
@@ -11,11 +11,14 @@ Persistence is available for:
 
 Persistence is configured per-node in the network graph definition. Add a `persistence` section to any node's configuration. This creates a new PVC for that node, which is then mounted to the appropriate data directory inside the container.
 
+Also add `restartPolicy: Always` to the node's configuration to ensure that the pod is restarted if it is deleted or crashes. This is important to ensure proper restart after restart of the kubernetes cluster.
+
 ### Bitcoin Core Node
 
 ```yaml
 bitcoin_core:
   image: bitcoincore-27.1:latest
+  restartPolicy: Always
   persistence:
     enabled: true
     size: 20Gi # optional, default is 20Gi
@@ -29,6 +32,7 @@ bitcoin_core:
 <lnd or cln>:
   image:
     tag: <node-version-tag>
+  restartPolicy: Always
   persistence:
     enabled: true
     size: 10Gi # optional, default is 10Gi
--- a/docs/persistence.md
+++ b/docs/persistence.md
@@ -11,7 +11,7 @@ Persistence is available for:
 
 Persistence is configured per-node in the network graph definition. Add a `persistence` section to any node's configuration. This creates a new PVC for that node, which is then mounted to the appropriate data directory inside the container.
 
-Also add `restartPolicy: Always` to the node's configuration to ensure that the pod is restarted if it is deleted or crashes. This is important to ensure proper restart after restart of the kubernetes cluster.
+Also add `restartPolicy: Always` to the node's configuration to ensure that the pod is restarted if it is deleted or crashes. This is important to ensure proper restart after restart of the kubernetes cluster. If there is a risk of hard resets of the cluster, add `reindex=1` to bitcoin_core config to reindex the blockchain on startup and fix potential corrupted chain state.
 
 ### Bitcoin Core Node
 
@@ -24,6 +24,8 @@ bitcoin_core:
     size: 20Gi # optional, default is 20Gi
     storageClass: "" # optional, default is cluster default storage class
     accessMode: ReadWriteOncePod # optional, default is ReadWriteOncePod. For compatibility with older Kubernetes versions, you may need to set this to ReadWriteOnce
+  config: |
+    reindex=1
 ```
 
 ### Lightning Node
--- a/docs/persistence.md
+++ b/docs/persistence.md
@@ -1,0 +1,55 @@
+# Data Persistence
+
+By default, Warnet nodes use ephemeral storage, meaning all data is lost when a pod is deleted or restarted. This document describes how to enable persistent storage to be able to use warnet for persistent development environment, such that blockchain data, wallet information, and other node state can survive pod restarts and network redeployments. This is done with Kubernetes Persistent Volume Claims (PVCs), which persist independently of pod lifecycle.
+
+Persistence is available for:
+- **Bitcoin Core** nodes
+- **LND** nodes
+- **CLN** nodes
+
+## Enabling Persistence
+
+Persistence is configured per-node in the network graph definition. Add a `persistence` section to any node's configuration. This creates a new PVC for that node, which is then mounted to the appropriate data directory inside the container.
+
+### Bitcoin Core Node
+
+```yaml
+bitcoin_core:
+  image: bitcoincore-27.1:latest
+  persistence:
+    enabled: true
+    size: 20Gi # optional, default is 20Gi
+    storageClass: "" # optional, default is cluster default storage class
+    accessMode: ReadWriteOnce # optional, default is ReadWriteOnce
+```
+
+### Lightning Node
+
+```yaml
+<lnd or cln>:
+  image:
+    tag: <node-version-tag>
+  persistence:
+    enabled: true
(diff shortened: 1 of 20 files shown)


resources/charts/bitcoincore/charts/cln/templates/pod.yaml
```diff
@@ -80,6 +82,18 @@ spec:
     - configMap:
         name: {{ include "cln.fullname" . }}
       name: config
+    - name: shared-volume
```
pinheadmz: e966d99d11f121110b8d3f3c4eb83770fccac08f

This can be named something more specific like `cln-datadir` and below, `lnd-datadir`. I think the reason it is called "shared volume" on main is because for lnd it is shared with circuit breaker and other containers that read the macaroons, etc. But that was a poor choice of name and we can fix it now.

resources/charts/bitcoincore/charts/cln/values.yaml
```diff
@@ -82,6 +82,15 @@ startupProbe:
       - "-c"
       - "lightning-cli createrune > /working/rune.json"
 
+# Node data persistence configuration. Create persistent volume claim, or use an existing one.
+persistence:
+  enabled: false
+  storageClass: ""
+  accessMode: ReadWriteOnce
```
pinheadmz: e966d99d11f121110b8d3f3c4eb83770fccac08f

I'm not super familiar with these options, but is there a reason you chose `ReadWriteOnce` which is accessible by the entire kubernetes node and not `ReadWriteOncePod` which seems to be a bit more specific for the use case?

docs/persistence.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Existing PVCs
+
+To use custom made PVC or PVC from previous deployment, use the `existingClaim` field to reference an existing PVC by name:
+
+```yaml
+persistence:
+  enabled: true
+  existingClaim: "bitcoin-node-001-data"
```
pinheadmz: 763d94009ca6fcc76f1708c6d175f9a18db64728

How do you get the name of the claim from the last lifecycle?

pinheadmz (commented):
Concept ACK so far, a few questions below.

It would be great if possible if you could add a test that demonstrates the data recovery. I feel like there's one step missing in the docs (see below) that I could use to test the feature.

Have you also tried the persistent volumes with a cloud provider like Digital Ocean? I'm just wondering if there's any obstacles with that. I imagine the volumes persist as long as the cluster is running.

Also - tested using Windows WSL! I know we connected on https://github.com/bitcoin-dev-project/warnet/issues/655 already, I'm just impressed you are using windows ;-)

And finally thanks for the contribution! Your PR is clean, well organized, and the commits show understanding of the repo which is really great. Please keep in touch about how you're using Warnet at Mion and if there's anything else you need.

docs/persistence.md
Jhoyola: The name is decided based on tank name, namespace and node type. I added it to the documentation.

resources/charts/bitcoincore/charts/cln/values.yaml
Jhoyola: I'm not very good with Kubernetes, but good suggestion, ReadWriteOncePod is better. I changed it to that.

Jhoyola:
I didn't test with any cloud setup. I'm actually quite new to kubernetes and never ran it in cloud. I would imagine this works nicely there too.

One thing I'm still unsure about: Now it's possible to leave the volumes intact even after `warnet down` command. But currently I don't think there is a way of restarting the network gracefully with existing data. The deploy command kind of works, but returns some errors because it expects it to be the first startup. And anyways if wallets, channels etc. already exist, it's not correct to recreate them I think. Would it make sense to create some "restart" command?

I at least added a small fix for the `ensure_miner` to allow to load existing miner wallet without throwing an error.

src/warnet/control.py
```diff
@@ -206,6 +217,36 @@ def delete_pod(pod_name, namespace):
...
+        for pvc in pvc_list:
+            table.add_row(pvc["namespace"], pvc["name"])
+        console.print(table)
+        pvc_confirmed = "pvc_confirmed"
+        pvc_answers = inquirer.prompt(
+            [
+                inquirer.Confirm(
+                    pvc_confirmed,
+                    message=click.style(
+                        "Do you also want to delete all PVCs containing persistent data? This will result in permanent loss of data.",
```
pinheadmz: Also just noticed this message is too long for click I guess? Just barely

Image

Jhoyola: Shortened it in rebase

Jhoyola:
Now by adding `restartPolicy: Always` and loading unloaded miner wallet in `ensure_miner` the nodes seem operational after each computer restart. The `ensure_miner` fix is important to be able to start miner after pod restart.

Jhoyola:
For me the ln_test passes locally on this branch so maybe there is some race condition or similar?

pinheadmz:
> For me the ln_test passes locally on this branch so maybe there is some race condition or similar?

Our ci is notoriously flaky, working with kubernetes has this "eventual consistency" to it which isn't easy to reproduce all the time. Anyway I restarted CI on your PR and everything looks good.

I'm going to add a test myself and push to your branch (if you don't mind) then we'll be RFM.

Jhoyola:
> I'm going to add a test myself and push to your branch (if you don't mind) then we'll be RFM.

Thanks, yes, sounds good.

Jhoyola:
~~Please don't merge yet.~~ I found a problem. Sometimes after restart the core nodes reset blockchain to block 0, eventough all the other data like wallets is persisted. I try to debug that.

Edit: I think the problem is me shutting down computer without allowing graceful exit of bitcoin core, which might corrupt the chainstate. I want to not worry about always doing graceful shutdown so I try to add some fix for it.

Edit: Fixed by adding reindex=1 to my node config. Added that to documentation.

pinheadmz:
Ok I'll take a look at that. I also noticed the `warnet-logging` pods use PVC by default which I didn't notice before but now with the extra shutdown warning, I'm looking into making those self-destruct at helm uninstall.

pinheadmz:
I rebased this on main after merging a few helm cleanups and extending the test base to confirm PVC deletion. Hoping this passes CI although im having trouble running everyhting cleanly locally. Assuming this works i'll add an actual test for persistence and push that.

pinheadmz:
I think we may need to also disable the lnd `StartupProbe` which initializes a brand new wallet and identity key, on a persistent restart.

Jhoyola:
> I think we may need to also disable the lnd `StartupProbe` which initializes a brand new wallet and identity key, on a persistent restart.

I totally missed that. Good that you noticed. I didn't test well enough with LND. With CLN I checked and confirmed that the channel still works, but yeah, the LND wallet is wiped, which is not ok for this case.

Would it make sense to do something like:

```
"if persistence.enabled && wallet.db exists && admin.macaroon exists":
  keep trying to unlock the wallet and never wipe/reset it

else:
  do the current behavior of init new wallet
```

pinheadmz:
I'll add a check if `admin.macaroon` exists already, I think that should be a reliable flag

pinheadmz:
OK I think I smoothed out all the issues with persistence and LND. Pushed a test, too. I need to do one last check with a full scale warnet like https://github.com/bitcoin-dev-project/wrath-of-nalo and if that goes well we can merge this.
THANKS!

PR #788, opened 2026-02-17
Configurable ports when service type is NodePort

Jhoyola:
For a simple and static way to connect to the node RPCs from host machine, the service type can be set to "NodePort". This PR makes it possible to fix static ports for each node to preconfigure a development environment based on warnet.

### DIFF
--- a/docs/connection-from-host.md
+++ b/docs/connection-from-host.md
@@ -1,0 +1,29 @@
+# Connection from host
+
+### Using NodePort
+
+To connect to a node from your host machine, you can use the NodePort service type. This exposes the node's desired ports on the host machine, allowing you to connect to them directly.
+
+For example to connect to a Bitcoin Core node using the RPC port, add this to the node's configuration in the network graph definition:
+
+```yaml
+nodes:
+  - name: tank-0001
+    service:
+      type: NodePort
+      rpcNodePort: 30443
+```
+
+Then you can connect to the node with `localhost:30443`. Or in non-local cluster `<kubernetes-cluster-ip>:30443`.
+
+All the different port options can be seen in values.yaml files. The exposed port values must be in the range 30000-32767. If left empty, a random port in that range will be assigned by Kubernetes.
+
+To check which ports are open on the host machine, use `kubectl get svc -n <namespace>` and look for the `PORT(S)` column.
+
+### Using port-forward
+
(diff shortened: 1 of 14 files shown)


pinheadmz:
I think mechanically this is the right choice to get remote access to each pod individually. It seems more correct and kubernetes-y than https://github.com/bitcoin-dev-project/warnet/pull/686 which addresses the same goal. 

Questions:
- How does the local user get the IP:port of the pod they want to connect to?
- How does this work in a large scale cluster with multiple kubernetes nodes, multiple pods on each node?

Jhoyola:
> * How does the local user get the IP:port of the pod they want to connect to?

User would add the below to the node on their network configuration. Then it's accessible at `localhost:30443`. I added documentation about this. Also added a small section about `kubectl port-forward` similar to https://github.com/bitcoin-dev-project/warnet/pull/686 but without a warnet command.

```
    service:
      type: NodePort
      rpcNodePort: 30443
```

> * How does this work in a large scale cluster with multiple kubernetes nodes, multiple pods on each node?

Haven't tested it, but according to Kubernetes docs it works just as: `<kubernetes-node-ip>:<exposed-port>`

Jhoyola:
Rebased

pinheadmz:
Sorry about the long delay. Rebased on main mainly to refresh CI. WIll review today.

pinheadmz:
fixed a nit in the doc and added a test

pinheadmz:
adding new command `warnet host` which can be used to get the cluster node IP or hostname (including minikube) for use with NodePort

pinheadmz:
Tested locally with macos using Docker Desktop and Minikube backends. Ran a remote cluster on digital ocean and tested that setup from debian. Along with the CI which runs ubuntu and Minikube, I think this is working as well as expected. 

The tricky part was `warnet host` -- determining where to connect to those NodePorts on all platforms. Especially on macos, where kubernetes is run inside a VM with IP addresses normally inaccessible to the host.

Unfortunately there is no good way to test the feature on Macos using minkube backend, so I programmed the test to just skip if that is the condition

pinheadmz:
@Jhoyola I think this is ready to merge, do you wanna take one last look and make sure I didn't break anything you need?

Jhoyola:
Looks good to me. Thanks!

PR #791, opened 2026-02-23
fix: allow onion addresses and external peers in messages command

### LINKED ISSUE #761: captured messages retrieval should allow onion addresses or peers that are not known tanks
"`warnet bitcoin messages tank_a tank_b`" expects both `tank_a` and `tank_b` to be names of tanks, warnet figures out their IP addresses and then finds the directory in `messge_captrue` that matches. However, when using Tor there may be messages captured from onion peers, and warnet does not know how to find them. As a one-off I wrote a brutal patch just so I could execute this:

`warnet bitcoin messages tank-0001 x76xhrwp6urgosotdfwx37v2qeeklyf32gllgzevj6kevpzfift4lnid.onion_18444`

```diff
diff --git a/src/warnet/bitcoin.py b/src/warnet/bitcoin.py
index 5b392e22..c58a4949 100644
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -178,20 +178,20 @@ def messages(tank_a: str, tank_b: str, chain: str):
         sys.exit(1)
 
     tank_a, namespace_a = parse_name_and_namespace(tank_a)
-    tank_b, namespace_b = parse_name_and_namespace(tank_b)
+    # tank_b, namespace_b = parse_name_and_namespace(tank_b)
 
     try:
         namespace_a = get_default_namespace_or(namespace_a)
-        namespace_b = get_default_namespace_or(namespace_b)
+        # namespace_b = get_default_namespace_or(namespace_b)
 
         # Get the messages
         messages = get_messages(
-            tank_a, tank_b, chain, namespace_a=namespace_a, namespace_b=namespace_b
+            tank_a, tank_b, chain, namespace_a=namespace_a, namespace_b="default"
         )
 
         if not messages:
             print(
-                f"No messages found between {tank_a} ({namespace_a}) and {tank_b} ({namespace_b})"
+                f"No messages found between {tank_a} ({namespace_a}) and {tank_b} ({'default'})"
             )
             return
 
@@ -225,14 +225,14 @@ def get_messages(tank_a: str, tank_b: str, chain: str, namespace_a: str, namespa
     base_dir = f"/root/.bitcoin/{subdir}message_capture"
 
     # Get the IP of node_b
-    cmd = f"kubectl get pod {tank_b} -o jsonpath='{{.status.podIP}}' --namespace {namespace_b}"
-    tank_b_ip = run_command(cmd).strip()
+    # cmd = f"kubectl get pod {tank_b} -o jsonpath='{{.status.podIP}}' --namespace {namespace_b}"
+    tank_b_ip = tank_b
 
     # Get the service IP of node_b
-    cmd = (
-        f"kubectl get service {tank_b} -o jsonpath='{{.spec.clusterIP}}' --namespace {namespace_b}"
-    )
-    tank_b_service_ip = run_command(cmd).strip()
+    # cmd = (
+    #     f"kubectl get service {tank_b} -o jsonpath='{{.spec.clusterIP}}' --namespace {namespace_b}"
+    # )
+    tank_b_service_ip = tank_b
 
     # List directories in the message capture folder
     cmd = f"kubectl exec {tank_a} --namespace {namespace_a} -- ls {base_dir}"

```

nkatha23:
- Add is_external_peer() helper to detect onion addresses and non-tank peers
- Skip namespace parsing and kubectl IP lookups for external peers
- Match message capture directories directly by address for external peers
- Fix dot-split validation that would reject onion addresses containing dots
- Add check_messages() test to onion_test.py to verify the fix

Fixes #761

### DIFF
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -159,6 +159,7 @@ def messages(tank_a: str, tank_b: str, chain: str):
     """
     Fetch messages sent between <tank_a pod name> and <tank_b pod name> in [chain]
 
+    tank_b can be a tank pod name, an onion address, or any external peer identifier.
     Optionally, include a namespace like so: tank-name.namespace
     """
 
@@ -174,7 +175,7 @@ def messages(tank_a: str, tank_b: str, chain: str):
     tank_b_split = tank_b.split(".")
     if len(tank_a_split) > 2 or len(tank_b_split) > 2:
         click.secho("Accepted formats: tank-name OR tank-name.namespace")
-        click.secho(f"Foramts found: {tank_a} {tank_b}")
+        click.secho(f"Formats found: {tank_a} {tank_b}")
         sys.exit(1)
 
     tank_a, namespace_a = parse_name_and_namespace(tank_a)
@@ -219,20 +220,30 @@ def messages(tank_a: str, tank_b: str, chain: str):
 
 def get_messages(tank_a: str, tank_b: str, chain: str, namespace_a: str, namespace_b: str):
     """
-    Fetch messages from the message capture files
+    Fetch messages from the message capture files.
+
+    Resolves tank_b to an IP via kubectl for known pods. If that fails (e.g. for
+    onion addresses or external peers), falls back to matching tank_b directly
+    against message capture directory names.
     """
     subdir = "" if chain == "main" else f"{chain}/"
     base_dir = f"/root/.bitcoin/{subdir}message_capture"
 
-    # Get the IP of node_b
-    cmd = f"kubectl get pod {tank_b} -o jsonpath='{{.status.podIP}}' --namespace {namespace_b}"
-    tank_b_ip = run_command(cmd).strip()
+    # Try to resolve tank_b to IPs via kubectl; fall back to tank_b directly on failure.
+    try:
+        cmd = f"kubectl get pod {tank_b} -o jsonpath='{{.status.podIP}}' --namespace {namespace_b}"
+        tank_b_ip = run_command(cmd).strip()
+    except Exception:
+        tank_b_ip = ""
 
-    # Get the service IP of node_b
-    cmd = (
-        f"kubectl get service {tank_b} -o jsonpath='{{.spec.clusterIP}}' --namespace {namespace_b}"
-    )
-    tank_b_service_ip = run_command(cmd).strip()
+    try:
+        cmd = f"kubectl get service {tank_b} -o jsonpath='{{.spec.clusterIP}}' --namespace {namespace_b}"
+        tank_b_service_ip = run_command(cmd).strip()
+    except Exception:
+        tank_b_service_ip = ""
+
+    # Use resolved IPs for matching, or tank_b itself if kubectl lookups came up empty.
+    identifiers = [ip for ip in [tank_b_ip, tank_b_service_ip] if ip] or [tank_b]
 
     # List directories in the message capture folder
     cmd = f"kubectl exec {tank_a} --namespace {namespace_a} -- ls {base_dir}"
@@ -242,7 +253,7 @@ def get_messages(tank_a: str, tank_b: str, chain: str, namespace_a: str, namespa
     messages = []
 
     for dir_name in dirs:
-        if tank_b_ip in dir_name or tank_b_service_ip in dir_name:
+        if any(ident in dir_name for ident in identifiers):
             for file, outbound in [["msgs_recv.dat", False], ["msgs_sent.dat", True]]:
                 file_path = f"{base_dir}/{dir_name}/{file}"
                 # Fetch the file contents from the container
--- a/test/onion_test.py
+++ b/test/onion_test.py
@@ -76,6 +76,18 @@ class OnionTest(TestBase):
 
         self.wait_for_predicate(onion_connect, timeout=20 * 60)
 
+        self.log.info("Checking captured messages from onion peer...")
+
+        # tank-0001's onion address as seen by tank-0000
+        onion_b = onions["tank-0001"]
+
+        result = self.warnet(f"bitcoin messages tank-0000 {onion_b}")
+        self.log.info(f"Messages result: {result}")
+        assert result and len(result.strip()) > 0, (
+            f"Expected messages between tank-0000 and {onion_b} but got none"
+        )
+        self.log.info("Successfully retrieved messages from onion peer!")
+
 
 if __name__ == "__main__":
     test = OnionTest()


nkatha23:
@pinheadmz  I investigated the rpc_test failure and I believe it's a pre-existing issue unrelated to my changes.
The test/data/12_node_ring/node-defaults.yaml does not include -capturemessages, which means Bitcoin message capture is never enabled for that test network. Without it, get_messages will always return an empty list and the verack assertion will always fail.
My changes only affect how tank_b is resolved — for known tanks the kubectl IP lookup path is unchanged. The rpc_test failure would occur on main as well.
Happy to also fix the node-defaults.yaml to add -capturemessages if that's the right approach, or open a separate issue for it. Let me know how you'd like to proceed.

nkatha23:
14/16 checks passing. The only remaining failure is plugin_test which is a pre-existing issue in test_base.py, while investigating it I discovered that inspect.getsource() can't handle functools.partial objects, causing a TypeError that masks the real timeout error. I've opened a separate PR #795 to fix that. This PR is ready for review.

pinheadmz:
rebased on main after merging #803

src/warnet/bitcoin.py
```diff
@@ -151,6 +151,15 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
     return matching_logs
 
 
+@bitcoin.command()
+@click.argument("tank_a", type=str, required=True)
+@click.argument("tank_b", type=str, required=True)
+@click.option("--chain", default="regtest", show_default=True)
```
pinheadmz: f4566e6a642bea24127d6f5851950a55bc193676

I see you fixed this in a subsequent commit but itd be easier to review if you squashed the changes

src/warnet/bitcoin.py
```diff
@@ -151,6 +151,15 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
...
+def is_external_peer(name: str) -> bool:
+    """Check if tank_b is an external/non-tank peer (e.g. onion address or raw IP)"""
+    return ".onion" in name or not name.startswith("tank-")
```
pinheadmz: f4566e6a642bea24127d6f5851950a55bc193676

there's actually nothing requiring tanks' names to start with `tank-` (Its common for one tank to just be called `miner` for example)

You can change one of the tank names in `test/data/onion/network.yaml` to make sure your patch still works.

test/onion_test.py
```diff
@@ -76,6 +76,26 @@ def onion_connect():
 
         self.wait_for_predicate(onion_connect, timeout=20 * 60)
 
+        self.check_messages(onions)
```
pinheadmz: f4566e6a642bea24127d6f5851950a55bc193676

Normally this would go inside the try/catch in `run_test` up top. However it wouldn't pass as a separate independent test since it relies on everything in `check_tor` to have already happened. Therefore I think you can just add your test to the existing block:

```diff
diff --git a/test/onion_test.py b/test/onion_test.py
index b50cda48..cfb93975 100755
--- a/test/onion_test.py
+++ b/test/onion_test.py
@@ -76,9 +76,6 @@ class OnionTest(TestBase):
 
         self.wait_for_predicate(onion_connect, timeout=20 * 60)
 
-        self.check_messages(onions)
-
-    def check_messages(self, onions: dict):
         self.log.info("Checking captured messages from onion peer...")
 
         # tank-0001's onion address as seen by tank-0000
```

src/warnet/bitcoin.py
```diff
@@ -177,22 +187,35 @@ def parse_name_and_namespace(tank: str) -> tuple[str, Optional[str]]:
         click.secho(f"Foramts found: {tank_a} {tank_b}")
         sys.exit(1)
 
+        # Only validate tank_b format if it's not an external peer
```
pinheadmz: f4566e6a642bea24127d6f5851950a55bc193676

nit: comment is extra-indented

pinheadmz (commented):
code review  41892322 

Thanks for taking this on! I left a few comments below but in general, please squash any fixes you make within the branch. To me it looks like this whole PR could be one single commit.

And then, I think the approach could be improved. I think `messages()` could be entirely left alone, and only in `get_messages()` we fall back to whatever the exact value of `tank_b: str` is if we fail to map that tank name to an IP address with a directory in `message_capture`.

I don't think `namespace_b is None` should be a trigger for the alternate behavior. 

Also I wonder if we can add the port internally? Instead of the user needing to suffix `._18444` or whatever?

nkatha23:
Hello @pinheadmz, I've addressed all review feedback:

Removed is_external_peer() helper
Restored messages() to original shape, fixed typo
Updated get_messages() with try/except kubectl fallback,  if IP lookup fails, falls back to matching tank_b directly against directory names
Inlined onion message check into check_tor per your diff
Squashed all commits into one

pinheadmz (approved):
ACK 71ef99bac0534753863403ffb7b04419b9150341

Great work THANK YOU!!!

```
Hash: SHA256

ACK 71ef99bac0534753863403ffb7b04419b9150341
-----BEGIN PGP SIGNATURE-----

iQJPBAEBCAA5FiEE5hdzzW4BBA4vG9eM5+KYS2KJyToFAmnmNTkbFIAAAAAABAAO
bWFudTIsMi41KzEuMTIsMCwzAAoJEOfimEtiick6rxUQAK/343MatPEwzfxchrN5
Rj3YxEHcCSpA80bLBPBMIMw2yg73qTiYuUcyk0e/TEGu/dDfzcVr/yy9bWNKEFjP
FBTzhhngq+wiGsICbd90FS7j1A7YjVIXLlIvPgp3jS32NExlzeY97py5fpi3T7oJ
L1rt2ml7O2ZVp3MHZujQ2feTQ31D6g/y/XICZ0sTsIgArcLc/svReJr4xcBc9Vc5
vfLq6YQFfxMLkQG/d5++ol2qCtV5+kPkOoE2/SQbYMewiAcKknw1qq/0nQmD0uwJ
JGXeRz2mX5vkus9Iv4i6MMs7B4M41v18vTbUSQ5KNwaGEha1uH5gwKw+mG4eppKG
VKqAAuUKva9MULmeCxYcEWbnfLF3sBGrO39AtWyAsRjZy+mnOoxmkPsn8gW/RigE
JW87esW4//2nctomfkG+V3KHGbGNT18xH3IEW43HyMW003lkpBuV7jB/OvCRasgm
JgarrUFFANg0PAb2ml9TJ2+5/b9WOHvt2JIKfX2kMNp/wjE1k7a2Quski8s38uWz
RmGmkC9rxYRBYKYhLBaCqnt91BM/+1m7wmLndoyIpgpnbcsqa+PuBMolDr1q4Bep
8l3JeEEUX1i3JyN1JXQUJah2oC6Fv296szS+9QKNlsg0VH761DGWTLWEGxADvmI7
8KMT05Z6EjeJRRHDk197uuvV
=JUKa
-----END PGP SIGNATURE-----
```

pinheadmz's public key is [on openpgp.org](https://keys.openpgp.org/vks/v1/by-fingerprint/E61773CD6E01040E2F1BD78CE7E2984B6289C93A)
</details>

PR #795, opened 2026-02-26
fix: handle partial objects in wait_for_predicate timeout message

nkatha23:
Found this while investigating a CI failure on PR #791. plugin_test was timing out, but instead of seeing the real timeout error, CI was throwing a confusing TypeError because inspect.getsource() doesn't support functools.partial objects. This fix makes the error reporting robust by falling back to repr() when getsource() fails, so the actual timeout reason is visible instead of a misleading TypeError.

### DIFF
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -97,9 +97,11 @@ class TestBase:
             timeout -= interval
         import inspect
 
-        raise Exception(
-            f"Timed out waiting for Truth from predicate: {inspect.getsource(predicate).strip()}"
-        )
+        try:
+            source = inspect.getsource(predicate).strip()
+        except (TypeError, OSError):
+            source = repr(predicate)
+        raise Exception(f"Timed out waiting for Truth from predicate: {source}")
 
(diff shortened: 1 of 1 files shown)


pinheadmz:
Hi there and sorry for the long delay. I'm looking in to this but what I think makes more sense is just removing `partial` from the plugin test altogether, it seems unnecessarily abstract. I'll try to simplify that with a PR against main branch and then rebase your onion address work on top so we can see the errors in the CI log more clearly. Thanks!

nkatha23:
> Hi there and sorry for the long delay. I'm looking in to this but what I think makes more sense is just removing `partial` from the plugin test altogether, it seems unnecessarily abstract. I'll try to simplify that with a PR against main branch and then rebase your onion address work on top so we can see the errors in the CI log more clearly. Thanks!

Thanks for looking into it. That makes total sense.  removing the partial entirely is cleaner than patching the error handling around it. I'll close this PR since #803 addresses the root cause. Looking forward to seeing the onion address work rebased on top

PR #798, opened 2026-03-10
Guard missing peer addr in connect_dag assertions

ashishkrshaw:
I noticed a small edge case in connect_dag.py: some peer entries can have a missing or null addr, and the current assertions can crash while splitting or substring checks. This change keeps behavior the same for valid data, but adds a safe fallback using d.get('addr') or '' for both DNS and IP assertions. Also fixed a typo in the error message from conectee to connectee.


pinheadmz:
Have you observed this error running the test locally or on our CI?
Why would the `addr` be null?

PR #799, opened 2026-03-11
lnd: only init wallet with root key if supported

ekzyis:
~This makes it possible to override `startupProbe` of an LND node.~

~This is required for https://github.com/bitcoin-dev-project/wrath-of-nalo/pull/8 because lnd v0.15.5-beta does not support `macaroon_root_key` in `initwallet`. (Support was added in v0.16.0, see https://github.com/lightningnetwork/lnd/pull/6457.~

~I couldn't simply use `{{- toYaml .Values.startupProbe | nindent 8 }}` and put the current default value into values.yaml similar to `livenessProbe` etc. because `{{ .Values.global.chain }}` wouldn't get interpolated.~

~This was tested by deploying the battlefield in https://github.com/bitcoin-dev-project/wrath-of-nalo/pull/8 and inspecting the pod running lnd v0.15.5-beta.~

Instead of the above, it now parses the version, see https://github.com/bitcoin-dev-project/warnet/pull/799#issuecomment-4399219753.

### DIFF
--- a/resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
+++ b/resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
@@ -85,3 +85,15 @@ Create a hex-encoded RGB color derived from the namespace
 {{- printf "#%s" (substr 0 6 $hash) -}}
 {{- end -}}
 
+{{/*
+Data to init wallet with root key if lnd >= v0.16.0-beta
+*/}}
+{{- define "lnd.initwalletData" -}}
+{{- $tag := .Values.image.tag -}}
+{{- $supportsRootKey := semverCompare ">=0.16.0-beta" $tag -}}
+{{- if $supportsRootKey -}}
+"{\"macaroon_root_key\":\"{{ .Values.macaroonRootKey }}\", \"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"
+{{- else -}}
+"{\"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"
+{{- end -}}
+{{- end -}}
--- a/resources/scenarios/commander.py
+++ b/resources/scenarios/commander.py
@@ -16,7 +16,7 @@ from time import sleep
 
 from kubernetes import client, config
 from kubernetes.stream import stream
-from ln_framework.ln import CLN, LND, LNNode
+from ln_framework.ln import CLN, LND, LNNode, get_admin_macaroon
 from test_framework.authproxy import AuthServiceProxy
 from test_framework.blocktools import get_witness_script, script_BIP34_coinbase_height
 from test_framework.messages import (
@@ -117,7 +117,7 @@ for pod in pods.items:
                 pod.metadata.name,
                 pod.metadata.namespace,
                 pod_ip,
-                pod.metadata.annotations["adminMacaroon"],
+                get_admin_macaroon(sclient, pod),
             )
         if "cln" in pod.metadata.labels["app.kubernetes.io/name"]:
             lnnode = CLN(pod.metadata.name, pod.metadata.namespace, pod.status.pod_ip)
--- a/test/ln_basic_test.py
+++ b/test/ln_basic_test.py
@@ -27,6 +27,7 @@ class LNBasicTest(TestBase):
             "tank-0003-ln",
             "tank-0004-ln",
             "tank-0005-ln",
+            "tank-0006-ln",
         ]
 
         self.cb_port = 9235
@@ -101,7 +102,22 @@ class LNBasicTest(TestBase):
         except Exception as e:
             assert "verification failed: signature mismatch after caveat verification" in str(e)
         try:
+            self.warnet("ln rpc tank-0001-ln --rpcserver=tank-0006-ln.default:10009 getinfo")
+            raise AssertionError("That should not have worked!")
+        except Exception as e:
+            assert "verification failed: signature mismatch after caveat verification" in str(e)
+        try:
             self.warnet("ln rpc tank-0003-ln --rpcserver=tank-0004-ln.default:10009 getinfo")
+            raise AssertionError("That should not have worked!")
+        except Exception as e:
+            assert "verification failed: signature mismatch after caveat verification" in str(e)
+        try:
+            self.warnet("ln rpc tank-0003-ln --rpcserver=tank-0006-ln.default:10009 getinfo")
+            raise AssertionError("That should not have worked!")
+        except Exception as e:
+            assert "verification failed: signature mismatch after caveat verification" in str(e)
+        try:
+            self.warnet("ln rpc tank-0006-ln --rpcserver=tank-0004-ln.default:10009 getinfo")
             raise AssertionError("That should not have worked!")
         except Exception as e:
             assert "verification failed: signature mismatch after caveat verification" in str(e)
--- a/resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
+++ b/resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
@@ -44,6 +44,9 @@ spec:
       readinessProbe:
         {{- toYaml .Values.readinessProbe | nindent 8 }}  
       startupProbe:
+        {{- if .Values.startupProbe }}
+        {{- toYaml .Values.startupProbe | nindent 8 }}
+        {{- else }}
         failureThreshold: 10
         periodSeconds: 5
         successThreshold: 1
@@ -72,6 +75,7 @@ spec:
               until curl --fail --insecure https://localhost:8080/v1/initwallet --data "{\"macaroon_root_key\":\"{{ .Values.macaroonRootKey }}\", \"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"; do
                 sleep 5
               done
+        {{- end }}
       resources:
         {{- toYaml .Values.resources | nindent 8 }}
       volumeMounts:
--- a/resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
+++ b/resources/charts/bitcoincore/charts/lnd/templates/pod.yaml
@@ -69,7 +69,7 @@ spec:
 
               PHRASE=$(cat /tmp/genseed.json | grep -o '\[[^]]*\]')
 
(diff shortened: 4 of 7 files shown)


pinheadmz:
Instead of replacing the entire startup probe, can we add a flag (or even better, actual version parsing like [this](https://github.com/bitcoin-dev-project/warnet/blob/ff632b36a6d6183e037ae9f4a6edd2b27a7b7c8b/resources/charts/bitcoincore/templates/_helpers.tpl#L59-L70)) and just leave out the macaroon root key option? 

What happens if the option is still left in, does lnd 15 refuse to start?

I think the macaroon root key logic is mainly for friendly experimentation, and isn't used in battle games where each node has to have security from players, so I think this is generally an ok thing to do.

ekzyis:
> Instead of replacing the entire startup probe, can we add a flag (or even better, actual version parsing like [this](https://github.com/bitcoin-dev-project/warnet/blob/ff632b36a6d6183e037ae9f4a6edd2b27a7b7c8b/resources/charts/bitcoincore/templates/_helpers.tpl#L59-L70)) and just leave out the macaroon root key option?

Sounds good, I'll look into this, thanks!

> What happens if the option is still left in, does lnd 15 refuse to start?

The LND node starts, but it will wait forever for the wallet:

```
$ kubectl get pods
NAME           READY   STATUS    RESTARTS      AGE
aries-lnd      1/1     Running   0             21m
aries-lnd-ln   0/1     Running   0             21m
miner          1/1     Running   0             21m
miner-ln       3/3     Running   2 (19m ago)   21m
$ kubectl logs aries-lnd-ln | tail -1
2026-05-07 12:09:28.534 [INF] LTND: Waiting for wallet encryption password. Use `lncli create` to create a wallet, `lncli unlock` to unlock an existing wallet, or `lncli changepassword` to change the password of an existing wallet and unlock it.
```

because startupProbe times out:

```
$ kubectl describe pod aries-lnd-ln
...
Events:
  Type     Reason     Age                  From               Message
  ----     ------     ----                 ----               -------
  Normal   Scheduled  14m                  default-scheduler  Successfully assigned default/aries-lnd-ln to minikube
  Normal   Pulling    14m                  kubelet            Pulling image "lightninglabs/lnd:v0.15.5-beta"
  Normal   Pulled     14m                  kubelet            Successfully pulled image "lightninglabs/lnd:v0.15.5-beta" in 4.3s (15.665s including waiting). Image size: 101225027 bytes.
  Normal   Created    14m                  kubelet            Created container: lnd
  Normal   Started    14m                  kubelet            Started container lnd
  Warning  Unhealthy  4m9s (x2 over 9m9s)  kubelet            Startup probe failed: command timed out: "/bin/sh -c MACAROON_PATH=\"/root/.lnd/data/chain/bitcoin/regtest/admin.macaroon\"\n\n# If wallet is already initialized, unlock it and
exit\nif [ -f \"$MACAROON_PATH\" ]; then\n  until [ \"$(curl --silent --insecure https://localhost:8080/v1/unlockwallet --data \"{\\\"wallet_password\\\":\\\"AAAAAAAAAAA=\\\"}\")\" == \"{}\" ]; do\n    sleep 5\n  done\n  exit 0\nfi\n\nuntil curl --silent --insecure https://localhost:8080/v1/genseed > /tmp/genseed.json; do\n  sleep 5\ndone\n\nPHRASE=$(cat /tmp/genseed.json | grep -o '\\[[^]]*\\]')\n\nuntil curl --fail --insecure https://localhost:8080/v1/initwallet --data \"{\\\"macaroon_root_key\\\":\\\"IFwE0Bhza6cC+ZsnZsinpcdExoRO3D1PZH6adjXPlBM=\\\", \\\"wallet_password\\\":\\\"AAAAAAAAAAA=\\\", \\\"cipher_seed_mnemonic\\\": $PHRASE}\"; do\n  sleep 5\ndone\n" timed out after 5m0s
```

because this loop runs forever:

```bash
until curl --fail --insecure https://localhost:8080/v1/initwallet --data "{\"macaroon_root_key\":\"{{ .Values.macaroonRootKey }}\", \"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"; do
  sleep 5
done
```

(At least I'm pretty sure that's what's happening. k8s doesn't seem to expose the output of startupProbe.)

---

Edit: Could be that I misunderstood what you meant with "What happens if the option is still left in". Maybe you didn't mean what _currently_ happens (`macaroon_root_key` as part of `/initwallet`), but what would happen _after_ your suggestion was implemented, and `macaroonRootKey` is still part of the pod YAML config. Will check!

ekzyis:
> (or even better, actual version parsing like [this](https://github.com/bitcoin-dev-project/warnet/blob/ff632b36a6d6183e037ae9f4a6edd2b27a7b7c8b/resources/charts/bitcoincore/templates/_helpers.tpl#L59-L70))

5efdcc6 now does this, and it's much better!

> What happens if the option is still left in, does lnd 15 refuse to start?

It now doesn't matter if `lnd.macaroonRootKey` is set or not for lnd<0.16.0-beta, it won't be used.

I have tested this like this:

```
$ pip install -e .
$ warnet setup
...
$ warnet deploy battlefields/regtest_fake_channel_dos/
$ kubectl get pods
NAME           READY   STATUS    RESTARTS        AGE
aries-lnd      1/1     Running   0               2m40s
aries-lnd-ln   1/1     Running   0               2m40s
miner          1/1     Running   0               2m40s
miner-ln       3/3     Running   4 (2m11s ago)   2m40s
$ kubectl describe pods | grep initwallet
until curl --fail --insecure https://localhost:8080/v1/initwallet --data "{\"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"; do
until curl --fail --insecure https://localhost:8080/v1/initwallet --data "{\"macaroon_root_key\":\"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\", \"wallet_password\":\"AAAAAAAAAAA=\", \"cipher_seed_mnemonic\": $PHRASE}"; do
$ kubectl logs aries-lnd-ln | grep wallet
2026-05-07 16:46:49.160 [INF] LTND: Waiting for wallet encryption password. Use `lncli create` to create a wallet, `lncli unlock` to unlock an existing wallet, or `lncli changepassword` to change the password of an existing wallet and unlock it.
2026-05-07 16:46:55.217 [INF] LNWL: Opened wallet
2026-05-07 16:46:55.655 [INF] LNWL: The wallet has been unlocked without a time limit
2026-05-07 16:46:55.682 [INF] WLKT: Baking macaroons for WalletKit RPC Server at: /root/.lnd/data/chain/bitcoin/regtest/walletkit.macaroon
```

(The `miner-ln` pod is running v0.20.1-beta.)

resources/charts/bitcoincore/charts/lnd/templates/_helpers.tpl
```diff
@@ -85,3 +85,15 @@ Create a hex-encoded RGB color derived from the namespace
 {{- printf "#%s" (substr 0 6 $hash) -}}
 {{- end -}}
 
+{{/*
+Data to init wallet with root key if lnd >= v0.16.0-beta
+*/}}
+{{- define "lnd.initwalletData" -}}
+{{- $tag := .Values.image.tag -}}
+{{- $supportsRootKey := semverCompare ">=0.16.0-beta" $tag -}}
```
pinheadmz, line 93: It'd be great if you could add a 0.15.5 node in one of the tests so we can see this work in CI

ekzyis:
> It'd be great if you could add a 0.15.5 node in one of the tests so we can see this work in CI

Done in 56334d2. Included the node in `test_admin_macaroons`.

During network setup, wallet addresses are fetched to fund the node. I had to read admin.macaroon from the pod to pass RPC authentication.

Since reading admin.macaroon requires permission to execute code on the pod, any scenario that includes a pre-16 lnd node now needs to be run as admin, which is why I added `--admin` to the pyln_connect scenario.

I don't like this, but short of some hack to make pre-16 lnd support custom macaroon root keys, like writing macaroons.db with the root key material manually, or somehow wiring the return value of `initwallet` back into the pod config (both ideas sound crazy), I don't see how we could pass RPC auth without reading admin.macaroon.

resources/scenarios/commander.py
```diff
@@ -85,6 +86,45 @@
...
+    return RuntimeError("no lnd container found")
+
+
+def supports_macaroon_root_key(pod):
+    semver = get_lnd_semver(pod)
+    minor = int(semver.split(".")[1])
+    return minor >= 16
+
+
+def read_admin_macaroon(pod):
```
pinheadmz: Do these methods need to live in commander? Or could they be part of ln_framework/ln.py ?

test/ln_basic_test.py
```diff
@@ -72,7 +73,9 @@ def test_pyln_scenario(self):
         self.log.info("Running pyln_connect scenario")
         scenario_file = self.scen_dir / "test_scenarios" / "pyln_connect.py"
         self.log.info(f"Running scenario from: {scenario_file}")
-        stream_command(f"warnet run {scenario_file} --source_dir={self.scen_dir} --debug")
+        # --admin is required for networks with lnd versions below v0.16.0-beta
+        # to read admin.macaroon from the pod
+        stream_command(f"warnet run {scenario_file} --source_dir={self.scen_dir} --admin --debug")
```
pinheadmz: Could maybe try adding the exec privilege to non-admin role:

I think this is still safe for wargames because the namespaces continue to isolate players from targets...

```diff
diff --git a/resources/charts/commander/templates/rbac.yaml b/resources/charts/commander/templates/rbac.yaml
index da4d653b..69590077 100644
--- a/resources/charts/commander/templates/rbac.yaml
+++ b/resources/charts/commander/templates/rbac.yaml
@@ -16,7 +16,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["pods", "configmaps"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list", "watch", "exec"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
```

resources/scenarios/commander.py
ekzyis: moved the functions to ln_framework/ln.py in 32db780 + some refactor to have less lnd-specific code in commander

test/ln_basic_test.py
ekzyis: I like this more than conditionally requiring `--admin`.

I used the rule from `ClusterRole` in 32db780 so no `create` permission is unnecessarily granted for the resources `pods` and `configmaps`.

ekzyis:
Accidentally used `return RuntimeError` instead of `raise RuntimeError` in 32db780. Fixed in cd63f00.

(Interesting that the linter didn't catch that)

pinheadmz:
Tested this some more today and I like how its all working. i added v15 nodes to wrath of nalo and it seemed to deploy OK. If there's anything else we need to integrate these old nodes in warnet it can be added in a follow-up

PR #806, opened 2026-04-25
claude generated docs

pinheadmz:


mercie-ux:
I ran `apidocs.py` locally and went through the doc changes. The new reference files are a nice addition.

docs/plugins.md
```diff
@@ -1,72 +1,259 @@
...
+| `preDeploy` | Before anything else is deployed |
+| `postDeploy` | After all nodes and the network are deployed |
+| `preNode` | Before each individual node is deployed *(once per node)* |
+| `postNode` | After each individual node is deployed *(once per node)* |
+| `preNetwork` | After logging infrastructure, before nodes are launched |
+| `postNetwork` | After all node deploy threads have completed |
+
+### Per-node hooks and `node_name`
+
+For `preNode` and `postNode`, Warnet passes the current node's name in the context under the key `node_name`. Plugins can read this to act on a specific node. The pod name Warnet produces for a per-node plugin follows the pattern:
```
mercie-ux: This says `node_name` but `deploy.py` passes it inside `annex` should be `annex.node_name` to match the table below and avoid broken plugin code.

pinheadmz: Thanks for reviewing! I think this entire section can come out actually. I think the LLM generated a general rule based on what it saw in the example "hello" plugin, but this isn't how the `run_plugin()` code works.

pinheadmz:
I reviewed and cleaned up the claude output, i think this is good to go.

PR #808, opened 2026-05-27
Quote bitcoin kubectl command arguments

Asylian21:
## Summary
- build bitcoin RPC and log kubectl commands with `shlex.join` so dynamic arguments are shell-quoted consistently
- avoid `shell=True` when reading captured message files
- add a regression check that shell metacharacters remain single command arguments

## Validation
- `PYTHONPATH=.:src:resources/scenarios:test uv run --frozen python test/bitcoin_rpc_args_test.py`
- `uvx ruff check src/warnet/bitcoin.py test/bitcoin_rpc_args_test.py`
- `git diff --check`

### DIFF
--- a/src/warnet/bitcoin.py
+++ b/src/warnet/bitcoin.py
@@ -2,6 +2,7 @@ import json
 import os
 import re
 import shlex
+import subprocess
 import sys
 from datetime import datetime
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


pinheadmz:
Nice LLM output. Did you review before submitting? What's the motivation for this change?

PR #810, opened 2026-06-11
Feat/btcd nodes

lucasdbr05:
## Add btcd node support to Warnet

This PR integrates [btcd](https://github.com/btcsuite/btcd) as a first-class node implementation in Warnet

### What's Changed

- To enable btcd support, we added a new Helm chart under `resources/charts/btcd` along with a 6-node simnet example network under `examples/networks/6_node_btcd`.

 - Configuration and Helpers: updated constants.py and added btcd.py to manage implementation constants, detect pod chains/ports, and format credential flags for the `btcctl` CLI.

- RPC and Log Routing: modified `bitcoin.py` to dynamically route RPC calls (using bitcoin-cli or btcctl accordingly) and stream logs from the correct active container.

- Deployment: updated `deploy.py` to spin up the appropriate Helm chart based on the node's configured implementation and collect port definitions for the fork observer.

### Testing

- Run the 6-node simnet example:
  ```
  warnet deploy examples/networks/6_node_btcd
  ```
- Execute the RPC test scenario:
  ```
  warnet run resources/scenarios/test_scenarios/btcd_rpc_test.py
  ```
- Execute the btcd miner scenario:
   ```
    warnet run resources/scenarios/btcd_miner.py
    ```

### DIFF
--- a/examples/networks/6_node_btcd/node-defaults.yaml
+++ b/examples/networks/6_node_btcd/node-defaults.yaml
@@ -1,0 +1,13 @@
+chain: regtest
+
+implementation: btcd
+
+collectLogs: true
+metricsExport: false
+
+resources: {}
+
+image:
+  repository: lucasdbr05/btcd
+  pullPolicy: IfNotPresent
+  tag: "latest"
--- a/examples/networks/6_node_btcd/node-defaults.yaml
+++ b/examples/networks/6_node_btcd/node-defaults.yaml
@@ -1,6 +1,7 @@
 chain: regtest
 
 implementation: btcd
+restartPolicy: OnFailure
 
 collectLogs: true
 metricsExport: false
--- a/examples/networks/6_node_btcd/node-defaults.yaml
+++ b/examples/networks/6_node_btcd/node-defaults.yaml
@@ -3,6 +3,10 @@ chain: regtest
 implementation: btcd
 restartPolicy: OnFailure
 
+# TODO: find a better approach to can mine blocks (this address is just for tests)
+config: |
+  miningaddr=bcrt1qthmht0k2qnh3wy7336z05lu2km7emzfpm3wg46
+
 collectLogs: true
 metricsExport: false
 
--- a/resources/charts/btcd/templates/_helpers.tpl
+++ b/resources/charts/btcd/templates/_helpers.tpl
@@ -1,0 +1,59 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "btcd.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+*/}}
+{{- define "btcd.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "btcd.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "btcd.labels" -}}
+helm.sh/chart: {{ include "btcd.chart" . }}
+{{ include "btcd.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "btcd.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "btcd.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Map chain name to btcd config file flag
+*/}}
+{{- define "btcd.chainFlag" -}}
+{{- if eq .Values.global.chain "regtest" -}}
+simnet=1
+{{- else if eq .Values.global.chain "signet" -}}
+signet=1
+{{- else if eq .Values.global.chain "testnet" -}}
+testnet=1
+{{- else -}}
+{{/* mainnet: no flag needed */}}
+{{- end -}}
+{{- end -}}
--- a/resources/charts/btcd/templates/_helpers.tpl
+++ b/resources/charts/btcd/templates/_helpers.tpl
@@ -1,0 +1,59 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "btcd.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+*/}}
+{{- define "btcd.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "btcd.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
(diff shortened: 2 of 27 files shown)


pinheadmz:
Added a lint commit and rebased on main to run CI

resources/charts/btcd/templates/_helpers.tpl
```diff
@@ -0,0 +1,59 @@
...
+app.kubernetes.io/name: {{ include "btcd.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Map chain name to btcd config file flag
+*/}}
+{{- define "btcd.chainFlag" -}}
+{{- if eq .Values.global.chain "regtest" -}}
+simnet=1
```
pinheadmz: b84ca37f03721ca469b91b51cd802cc4aec90332

btcd doesn't have a regtest? is simnet different in any other way?

resources/charts/btcd/values.yaml
```diff
@@ -0,0 +1,107 @@
+# Default values for btcd.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+namespace: warnet
+
+restartPolicy: Never
+
+image:
+  repository: lucasdbr05/btcd
```
pinheadmz, line 9: b84ca37f03721ca469b91b51cd802cc4aec90332

Anything special about this image? do the lightning labs / roasbeef / btcsuite repository images on docker hub work as well?

src/warnet/bitcoin.py
```diff
@@ -65,10 +80,16 @@ def _rpc(tank: str, method: str, params: list[str], namespace: Optional[str] = N
             # Quote each parameter individually to preserve them as separate arguments.
             param_str = " ".join(shlex.quote(p) for p in params)
 
-        cmd = f"kubectl -n {namespace} exec {tank} --container {BITCOINCORE_CONTAINER} -- bitcoin-cli {method} {param_str}"
+        if is_btcd:
+            cmd = f"kubectl -n {namespace} exec {tank} --container {container} -- /bin/linux_amd64/btcctl --rpcuser=user --rpcpass=gn0cchi --rpccert=/root/.btcd/rpc.cert --rpcserver=127.0.0.1:18334 --simnet {method} {param_str}"
```
pinheadmz: dddc5f0fb6d2ad0ddf2c836d17ac300502051a72

We can't hard code rpc credentials or network name. We don't have to for Bitcoin Core because bitcoin-cli can read all the configuration details from `bitcoin.conf`. If btcctl doesn't do something similar we need a different way to get the config from the environment

src/warnet/bitcoin.py
```diff
@@ -111,7 +132,12 @@ def grep_logs(pattern: str, show_k8s_timestamps: bool, no_sort: bool):
             longest_namespace_len = len(tank.metadata.namespace)
 
         pod_name = tank.metadata.name
-        logs = pod_log(pod_name, BITCOINCORE_CONTAINER)
+        container_names = [c.name for c in tank.spec.containers]
+        if BTCD_CONTAINER in container_names:
+            log_container = BTCD_CONTAINER
+        else:
+            log_container = BITCOINCORE_CONTAINER
+        logs = pod_log(pod_name, log_container)
```
pinheadmz, line 180: dddc5f0fb6d2ad0ddf2c836d17ac300502051a72

Since we're breaking the seal on alternative implementations here, lets prepare for more. Instead of this kind "if btcd/else bitcoin core" logic everywhere, I think a cleaner approach would be to add details directly to the pod metadata.

For example we already have this line:
>         pod_name = tank.metadata.name

why not also add metadata like
>        bitcoin_implementation_name = tank.metadata.bitcoin_implementation

where the value is either "btcd", "bitcoincore", or None ?

The same kind of technique could replace `_get_node_container()`

This way if we ever add more implementations, the labels in their charts will do most of the work

examples/networks/6_node_btcd/node-defaults.yaml
```diff
@@ -1,6 +1,7 @@
 chain: regtest
 
 implementation: btcd
+restartPolicy: OnFailure
```
pinheadmz, line 4: c47fbbbb1d3428442709837e73dd4f22e5e4a7f2

Is this required for btcd? For lnd for example we expect it to crash and restart a few times on deployment while it waits for the bitcoin core node to come up. However, for bitcoin core itself we usually don't want this, because if bitcoin core crashes we don't want it to silently restart, we want to know why it crashed.

resources/scenarios/btcd_framework/btcd.py
```diff
@@ -0,0 +1,251 @@
...
+        payload = self._build_payload(method, list(params))
+        headers = self._build_headers(payload)
+        status, raw = self._send_with_retry(method, payload, headers)
+        return self._parse_response(method, status, raw)
+
+
+
+    # standard Bitcoin methods 
+
+    def getblockcount(self) -> int:
```
pinheadmz: d1e1d8d14f44a0ff9f3621870ef9619bc668f28d

I don't think you should have to implement all the RPCs this way. Python has a trick where if it doesn't recognize the method you are calling, it can be passed as an argument instead, with packed arguments. This is how the bitcoin core test framework works as well.

See how `__call__()` is used in authproxy.py and test_node.py

resources/scenarios/commander.py
```diff
@@ -102,6 +102,7 @@
                 "tank": pod.metadata.name,
                 "namespace": pod.metadata.namespace,
                 "chain": pod.metadata.labels["chain"],
+                "implementation": pod.metadata.labels.get("implementation", "bitcoincore"),
```
pinheadmz: d1e1d8d14f44a0ff9f3621870ef9619bc668f28d

This relates to my comment about adding labels to the chart. We shouldn't need a default here, just add the "bitcoincore" to the bitcoin core chart.

src/warnet/control.py
```diff
@@ -380,6 +380,7 @@ def filter(path):
                 "commander.py",
                 "test_framework",
                 "ln_framework",
+                "btcd_framework",
```
pinheadmz, line 383: d1e1d8d14f44a0ff9f3621870ef9619bc668f28d

This is ok for now, obviously we do the same thing for LN. But if we keep adding dependencies to the commander it'll get bloated. No need to change anything now but it might be time to start thinking about how to filter out unneeded stuff.

examples/networks/6_node_btcd/node-defaults.yaml
```diff
@@ -3,6 +3,10 @@ chain: regtest
 implementation: btcd
 restartPolicy: OnFailure
 
+# TODO: find a better approach to can mine blocks (this address is just for tests)
+config: |
```
pinheadmz, line 7: f50d525e7fef68c74c30d21d0253bdb93d1e5e87

We have `ensure_miner()` in the bitcoin test framework. Then in the mining scenario we call that and then `getnewaddress`

resources/scenarios/test_scenarios/btcd_rpc_test.py
```diff
@@ -0,0 +1,241 @@
+import time
+
+from commander import Commander, WARNET
+from btcd_framework import BtcdRPC, BtcdRPCError
+
+
+class BtcdRpcTest(Commander):
+    def set_test_params(self):
+        self.num_nodes = 1 
+        self.mining_addr = "Sh6VJ4TabWtfBm9kvLWPzj8WGNsMmtyaGF"
```
pinheadmz: f50d525e7fef68c74c30d21d0253bdb93d1e5e87

I assume these coins go nowhere? Or is the btcd wallet deterministic?

resources/scenarios/btcd_miner.py
```diff
@@ -0,0 +1,188 @@
...
+                    "peers":   peers,
+                    "mempool": mempool,
+                }
+            except Exception as exc:
+                status[node._tank_name] = {"error": str(exc)}
+        return status
+
+    def _log_status(self, round_num: int, status: dict):
+        self.log.info(
+            f"┌─────────────────────────────────────────────────────────────┐"
```
pinheadmz: Something just a little bit wrong with the tables in my output

Image

pinheadmz (commented):
This is really great work, thank you! I'm so glad you're using Warnet for the fountain codes research. The PR is coming along great, I have some feedback. The biggest blocker being the hard coded configuration values in `_rpc()`. I think everything else should work but could be cleaner. I also left some notes about abstracting the node implementation even further to make it easier for the **next** one we add ;-)

Also, the scenario and network you added are great but I'd like to see an extra test added to the CI that deploys the network, runs the two scenarios, then sends some commands to assert the expected values (like block height, wallet balance, etc). See the `test/` directory for examples, and add your new test to `.github/workflows/test.yml`


p.s. seeing a ton of these log messages from all the btcd nodes, all with unique port numbers:

> 2026/06/29 17:59:46 http: TLS handshake error from 10.1.0.1:60012: EOF

pinheadmz:
rebased on main to include #812 CI fix

pinheadmz:
For testing, you can move `/examples/networks/6_node_btcd/` to `test/data` (actually we should probably just delete `examples/` entirely) and add a test similar to the others there that starts the network and runs commands and scenarios. Add the test to `workflows/test.yml` and make sure the network includes some bitcoin core nodes as well! We'll have a nice integration test that way.

PR #814, opened 2026-07-06
Allow post-deployment connections straight from graph files

sr-gi:
Allows creating connections from a graph file other than manual. This connections are created after deployment using the addconnection (test-only) RPC.

Connections are stored under a new tag in the graph files: addconnection. If any of this tags is found, the scenario is run right after deployment to create those connections. Otherwise, this is a noop.

Graph file entry example:

```
nodes:
- addnode:
    - tank-0015
    - tank-0013
    - tank-0005
    - tank-0006
    - tank-0019
    - tank-0014
    - tank-0004
    - tank-0021 
- addconnection:
    - tank-0010
    - tank-0020
```

Disclaimer: addconnection.py has been highly vibecoded. Happy to trim it down or re-work it.

### DIFF
--- a/resources/scenarios/addconnection.py
+++ b/resources/scenarios/addconnection.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python3
+"""Open post-deploy peer connections via the addconnection RPC.
+
+Warnet wires the per-node ``addnode`` topology into bitcoin.conf at startup.
+This connections are flagged as manual, and don't allow testing some of the
+connection-specific functionality (e.g. block-only).
+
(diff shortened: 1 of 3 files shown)


pinheadmz:
This is cool, it's a feature I want, but I'm starting my own branch here: https://github.com/bitcoin-dev-project/warnet/pull/815

The syntax in network.yaml might end up being slightly different

sr-gi:
SGTM, happy to adapt to the format you decide on downstream when merged.

I'm currently running my patch downstream, but I'm also happy to ditch that one and run yours so I can give some practical review feedback

PR #818, opened 2026-08-25
docs: fix hello plugin example in plugins.md

mercie-ux:
The `hello` plugin example in `plugins.md` doesn't work. It tells you to run `warnet init` then `cat plugins/hello/plugin.py`, but that file doesn't exist. `hello` got moved to `test/data/plugins/hello/` in c9dcc86 , and the docs was never updated to reflect this.

I have fixed the example in the docs to actually work: make a project dir, run `warnet init`, then copy `hello` in from the source repo.

Tested the updated steps and `plugin.py` prints out now.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -68,10 +68,12 @@ for key, value in plugin_config.items():
 run_command(command)
 ```
 
-Start from the `hello` plugin included in every initialised project:
+The `hello` plugin is a complete example that exercises every hook. It isn't copied into new projects automatically — it lives at `test/data/plugins/hello/`. Replace `<path-to-warnet-repo>` below with wherever you cloned warnet:
 
 ```sh
+mkdir my_project && cd my_project
 warnet init
+cp -r <path-to-warnet-repo>/test/data/plugins/hello plugins/
 cat plugins/hello/plugin.py
 ```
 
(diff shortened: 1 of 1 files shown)


pinheadmz:
Thanks for catching this. I think the hello plugin is just used for testing the plugin hooks and that's why we moved it. Probably makes more sense to remove its mention from the doc entirely. Is there another plugin that gets installed by init we could explain instead? I'd prefer new users learn about plugins without having to copy a test-only module to their project.

mercie-ux:
> Thanks for catching this. I think the hello plugin is just used for testing the plugin hooks and that's why we moved it. Probably makes more sense to remove its mention from the doc entirely. Is there another plugin that gets installed by init we could explain instead? I'd prefer new users learn about plugins without having to copy a test-only module to their project.

Yes, there are two plugins, `simln`and `tor`, although documentation for these two already exists. I think removing the `hello `example is the simplest fix, so I've removed it.

pinheadmz:
What do you think about the "full hook example"?

mercie-ux:
> What do you think about the "full hook example"?

I had missed the “full hook example” when removing the `hello` plugin documentation. I’ve removed it as well, since it referenced the unavailable `hello` plugin.

pinheadmz:
ACK

thanks!

